Issue #185, opened 2015-01-07
PixHawk timestamp instead of ROS time


thedinuka:
I was looking at imu_pub.cpp and realized that imu/data_raw  topic is published with the header timestamp set to ROS::Time::now() 
(https://github.com/mavlink/mavros/blob/master/mavros/src/plugins/imu_pub.cpp#L250)

The MAVLINK_MSG_ID_HIGHRES_IMU message from PixHawk does contain a time_usec filed which actually represent the time at which the imu data was captured. For certain applications like for synchronizing a camera and IMU, it maybe useful to have that as the timestamp instead of the ros time. 

I'm wondering whether this would qualify as a new feature request? I could do a PR for this, but the change in the code is really simple anyway.

mhkabir:
Hi, 
We have timesync working. I will be updating the plugins soon to take advantage of this. We synchronise Pixhawk timestamps to ROS time and vice-versa.
What you propose would not be ideal, as ROS wall time is calculated from Unix Epoch and Pixhawk time is calculated from boot.

thedinuka:
@mhkabir timesync would be ideal, I agree. What sort of accuracies can you achieve with it?

mhkabir:
We get microsecond precision.

thedinuka:
ahh, excellent. would you be able to comment on when this will be available?

mhkabir:
Pretty soon hopefully. Synchronisation works already (check system_time.cpp ) The plugins just need to be updated.

thedinuka:
@mhkabir I'm still reading through system_time.cpp (actually  it's sys_time.cpp right?) but as a preliminary question, is the idea to maintain the difference between ros time and PixHawk time and then use that offset in other plugins to set the correct timestamp?

mhkabir:
Yes

vooon:
I testing PR #186.

Setup:
- Wandboard Quad with Ubuntu 14.04 and latest ROS for UbuntuARM
- Pixhawk with latest master

Results:
First run without custom `-m onboard` mode.
Got timsync, but strange timestamps:

*/mavros/imu/data*
```yaml
---
header: 
  seq: 4316
  stamp: 
    secs: 4294967295
    nsecs: 4294967295
  frame_id: fcu
orientation: 
  x: -0.0430093572484
  y: 0.00982332692738
  z: -0.307822874007
  w: 0.950420315269
[...]

PR #186, opened 2015-01-11
plugin: add synchronisation to most plugins

mhkabir:
This allows timesync between a PX4 FCU and mavros to be used in most plugins : #185 
This is untested, so don't merge yet.

I don't have time for this ATM, so testers welcome. @thedinuka, interested?

### DIFF
--- a/mavros/include/mavros/mavros_uas.h
+++ b/mavros/include/mavros/mavros_uas.h
@@ -324,14 +324,23 @@ public:
 	bool cmode_from_str(std::string cmode_str, uint32_t &custom_mode);
 
 	/**
-	 * @brief Compute FCU message time from time_boot_ms field
+	 * @brief Compute FCU message time from time_boot_ms or time_usec field
 	 *
 	 * Uses time_offset for calculation
 	 *
-	 * @todo
-	 * @return FCU time if if is known else current wall time.
+	 * @return FCU time if it is known else current wall time.
 	 */
-	//ros::Time fcu_time(uint32_t time_boot_ms);
+	inline ros::Time synchronise_stamp(uint32_t time_boot_ms) {
+		if(time_offset > 0)		
+			return ros::Time(time_boot_ms + time_offset/1000000);
+		else return ros::Time::now();
+	}
+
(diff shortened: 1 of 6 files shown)


thedinuka:
I'll definitely run some tests.  Will get back to you  after.

mhkabir:
You can simply merge the branch, and add two printf/Throttled print to one
of the plugins. Print the synchronised timestamp and ros::Time::now(). As
long as they are similar, it's fine.

vooon:
Looks good, but testing required.

thedinuka:
Hi,
I'm trying to understand this a bit more clearly before I run tests with it. Hope you'd be able to answer a few questions.
From what I understood, time offset is the instantaneous time difference between ROS wall time and the pixhawk time. Then what the `synchronise_stamp` function does is to add that offset to the time stamp attached by the pixhawk to each sensor data packet. So effectively, what 
```
header.stamp = uas->synchronise_stamp(imu_hr.time_usec);
```
does is to set the timestamp on the message published by ros to the ros time at which the IMU (or other sensor) was captured as opposed to ros::Time::now(). Given that there is a finite delay in transmitting the data from PixHawk to host computer, I'd assume the header.stamp to always lag ros::Time::now(). 

This is what confused me a bit, because you said:
>Print the synchronised timestamp and ros::Time::now(). As long as they are similar, it's fine

Hope you could shed some light on this. Thanks

mhkabir:
The lag would be low in any case. The current wall time would be very close to the synchronised stamp anyway. (lagging by estimator time + link latency) 

As I said, they should be close to each other, not the same.

mhkabir:
It's fine with me if you can find a better way to test it :)

mhkabir:
@dinuka You can plot the IMU in rqt_plot and move the system around.
Visualisation should be normal if synchronisation works properly :)

thedinuka:
@mhkabir I will do those testing as soon as I get some time.

mhkabir:
@tonybaltovski

tonybaltovski:
@mhkabir will do ASAP.

Issue #49, opened 2014-07-12
Check frame conversions (NED->ENU ROS)


vooon:
Current implementation use this definitions [av8n.com](http://www.av8n.com/physics/coords.htm) of NED and ENU. Conversion is just invert Y & Z axis (all right-handed).

But according to [CoordinateFramConventions](http://wiki.ros.org/geometry/CoordinateFrameConventions) it is left-handed?

ziyangli:
@vooon 
I have the same question ---- Why magnetometer use different conversion?

tonybaltovski:
The mag is absolute and the others are relative.

supermice:
Hi, everyone
     Does that means the attitude and position use the different base_coordinates in ROS? I tried to display the 'tf' data in ROS rviz use mavros . I find that the attitude's x axis is in red and when I move the aircraft along the attitude's x axis the 'tf' shows move to left(-y direction).  Is there any problem  with the local_origin tf or I use the old mavros version?

PR #208, opened 2015-02-04
Changes some frames from world to body conversion for NED to ENU.

tonybaltovski:
For a desired position that is (1,5, 7) in ROS ENU, the must go forward 1 m which is X for both the ROS coordinate system and the PX4.  For y-axis, it is -5 m on the PX4 and the same for the z-axis.  I believe @mhkabir can verify this.  The frame issue #49 with two different frame conversions is due to keeping X as forward on the physical vehicle in ROS's coordinate system.  If we made Y forward, the IMU frames would be one conversion for body and inertial from NED to ENU thus being simpler to use.  However, this does break the ROS standards for coordinates.  @vooon Any insight?

### DIFF
--- a/mavros/src/plugins/local_position.cpp
+++ b/mavros/src/plugins/local_position.cpp
@@ -80,7 +80,7 @@ private:
 		 * orientation in ENU, body-fixed
 		 */
 		tf::Transform transform;
-		transform.setOrigin(tf::Vector3(pos_ned.y, pos_ned.x, -pos_ned.z));
+		transform.setOrigin(tf::Vector3(pos_ned.x, -pos_ned.y, -pos_ned.z));
 		transform.setRotation(uas->get_attitude_orientation());
 
 		auto pose = boost::make_shared<geometry_msgs::PoseStamped>();
--- a/mavros/src/plugins/setpoint_accel.cpp
+++ b/mavros/src/plugins/setpoint_accel.cpp
@@ -81,7 +81,7 @@ private:
 				ignore_all_except_a_xyz,
 				0.0, 0.0, 0.0,
 				0.0, 0.0, 0.0,
-				afy, afx, -afz,
+				afx, -afy, -afz,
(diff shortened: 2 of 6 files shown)


TSC21:
@vooon ping.
@mhkabir are you able to test this?

TSC21:
Related: https://github.com/mavlink/mavros/issues/225#issuecomment-108244805. Is this supposed to work?

mhkabir:
This PR is actually not the way to go forward. This breaks ROS standards. 

Please refer to #216 and add your comments

tonybaltovski:
This does not break any ROS standards.

Closed #208 <https://github.com/mavlink/mavros/pull/208>.

—
Reply to this email directly or view it on GitHub
<https://github.com/mavlink/mavros/pull/208#event-329060188>.

mhkabir:
@tonybaltovski Ok, in that case, we can get this back up after verification. It's gone stale anyway. 
Thanks for your contributions. 
Can you comment on the discussion please?

vooon:
@mhkabir reopened just not to lose pr.

tonybaltovski:
I believe that @congleetea confirmed this in #216 for the position frames.

vooon:
Ok, so i do merge.

TSC21:
@vooon I don't think this was supposed to merge yet. For all I know, @tonybaltovski was testing it today...

vooon:
Ooops. I may revert if needed.

TSC21:
Better you do it then :)

vooon:
That generate new merge, so before revert i want to hear "thats bug, revert it with fire!" :)

TSC21:
Ahahah sure! Let's see what @tonybaltovski tests say.

TSC21:
@tonybaltovski `vision_speed_estimate` plugin also require this changes (if they are to be accepted). Would you mind add fix to those to? Thanks!

PR #213, opened 2015-02-07
Corrects vim mode lines.

jgoppert:


jgoppert:
On Ubuntu 14.04 some of the mode lines were missing the closing colon which gave an annoying error when using vim. So I used modeliner to fix those cases and also standardized all mode lines using vim modeliner plugin.

vooon:
Ok, i will fix vim modelines, but i don't like to set ts and other settings. Its user choice.
I prefer `ts=8 sw=8 noet` for C and `ts=4 sw=4 et` for python. And this mostly defined in vimrc.

PR #221, opened 2015-02-25
Fix parameters in example commands

claymation:
The software doesn't seem to recognize parameter names with a leading underscore.

### DIFF
--- a/mavros/README.md
+++ b/mavros/README.md
@@ -62,7 +62,7 @@ Main node. Allow disable GCS proxy by setting empty URL.
 
 Run example:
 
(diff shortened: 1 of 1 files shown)


vooon:
No, for nodes `_` is required. It is represents that parameter uses node namespace `~`.
I.e. in wiki: '~fcu_url` => `_fcu_url` or `/mavros/fcu_url` in rosrun.

But launch files declares their arguments (`<arg />`) so for roslaunch you should use name from arg tag.
Look at [node.launch](https://github.com/mavlink/mavros/blob/master/mavros/launch/node.launch) and [px4 config](https://github.com/mavlink/mavros/blob/master/mavros/launch/px4_config.yaml).

claymation:
Cool, thanks. I'm new to ROS and really have no clue about namespaces yet :)

PR #223, opened 2015-02-26
plugin: setpoint_velocity: Fix vx setpoint

claymation:
vz should have been vx.


mhkabir:
Nice catch, thanks!

vooon:
Thanks!

PR #224, opened 2015-02-26
launch: enable setpoint plugins for APM

claymation:
As of ArduCopter 3.2, APM supports position and velocity setpoints via SET_POSITION_TARGET_LOCAL_NED.

### DIFF
--- a/mavros/launch/apm2_blacklist.yaml
+++ b/mavros/launch/apm2_blacklist.yaml
@@ -2,7 +2,6 @@ plugin_blacklist:
 - '3dr_radio'
 - '*_position'
 - 'safety_area'
-- 'setpoint_*'
 - 'px4flow'
 - 'image_pub'
 - 'vision_*'


mhkabir:
How do they handle the local frame? As far as I knew , all the controllers
worked in global...?

Do they reproject from a local to global using a reference? Or do it all in
local and reproject the global into local frame? (Like PX4?)

vooon:
goodnewseveryone

But i need to check how do they do it :).

claymation:
It looks to me like APM computes in local coordinates and converts global lat/long to local coordinates when necessary. Compare the handling of `SET_POSITION_TARGET_LOCAL_NED` to that of `SET_POSITION_TARGET_GLOBAL_INT` below:

https://github.com/diydrones/ardupilot/blob/master/ArduCopter/GCS_Mavlink.pde#L1307

That call to `pv_location_to_vector()` for the global target does the conversion:

https://github.com/diydrones/ardupilot/blob/77d8f88e45331ce2a54195c14b0ea1f76a624e36/ArduCopter/position_vector.pde#L11

Also note that z in inverted, suggesting a NEU frame is used internally:

https://github.com/diydrones/ardupilot/blob/master/ArduCopter/GCS_Mavlink.pde#L1330

But internals aside, it does work—I've tested in the SITL and found the following to be true:

+vx moves east
-vx moves west
+vy moves north
-vy moves south
+vz moves up
-vz moves down

vooon:
Not checked axes (if they not NED -- that apm bug!) only handler, now they support position, velocity, acceleration setpoints. APM chould be in GUIDED mode.

PR #227, opened 2015-02-27
1st try to make cmd_vel work as intended

mstuettgen:
Can u look at the bitmask for the mavlink message as I don't know how to set it correctly? This is the way I think cmd_vel should work as I want it

### DIFF
--- a/mavros/include/mavros/setpoint_mixin.h
+++ b/mavros/include/mavros/setpoint_mixin.h
@@ -55,7 +55,7 @@ public:
 				coordinate_frame,
 				type_mask,
 				x, y, z,
-				vz, vy, vz,
+				vx, vy, vz,
 				afx, afy, afz,
 				yaw, yaw_rate);
 		UAS_FCU(_uas)->send_message(&msg);


mhkabir:
@mstuettgen I simply don't get what you're trying to do here. Can you list the issues which you are trying to fix.

mhkabir:
For your purpose, use the setpoint_velocity plugin. It is what you need. Blacklist the rest to avoid interference (esp. setpoint_attitude). From the docs : 
* The Twist message linear velocity is sent to controllers ( You can use it for throttle, or otherwise mix XY in your controller)
* Angular velocity on Z (yaw rate) is sent

vooon:
@mhkabir for some reasons (bug?) px4 don't handle velocity correct (look gitter history).
But yeah, i think better find bug in PX4 side, because @mstuettgen anyway must add rover support to FW.
Also changing attitude by rover sounds confusing (controllable suspension? :)).

But need to look what frame used in typical ros commander packages (`cmd_vel`), i think that usually it is body frame, so linear X thrust, angular Z steering.
But in our setpoints used local, so thrust decomposes into linear XY.

mhkabir:
Either way, it would make sense to use the setpoint velocity plugin since it is already doing what he wants to do. 
Just check the thrust to XY thing, and you're good.

mstuettgen:
I dont think setpoint velocity can be used for what I want to do....

First of all, you do not use the "set_attitute" method of mavlink. Having some information passed to set_position_target_local_NED does not help me at all...

I need the steering and thrust data to be sent to https://pixhawk.ethz.ch/mavlink/#SET_ATTITUDE_TARGET, because I can only access this data in my rover steering app.

I hope now it is a little bit more clear why and what I changed

LorenzMeier:
@vooon Looking at the code you might want to consider adding a style guide and adding something like this: https://github.com/PX4/Firmware/blob/master/Tools/fix_code_style.sh

Its getting hard to read and maintain if braces are all over the place ;-).

mstuettgen:
And what is really confusing too is that setpoint_attitude and setpoint_velocity both take a cmd_vel msg and handle them totally different.

TSC21:
@mstuettgen if you check out the code of both files, you'll see that `setpoint_attitude` handles only the angular velocity component of `cmd_vel` message, while `setpoint_velelocity` handles the linear component. That is due to the type of mavlink messages that are handled in each plugin: `SET_ATTITUDE_TARGET` is handled in the attitude plugin, and in that message you control angular velocities, while `SET_POSITION_TARGET_LOCAL_NED` is handled in the velocity plugin.
If you want for example, control the linear velocities of a quadrotor, you send cmd_vel linear velocities, and it will affect the FCU in the linear components, while if you want to change the yaw rate, you send a angular velocity in the cmd_vel, so to change the yaw rate. This is what can be done to quadrotors, so if yow want to do that to rovers you need to do something similar in your control side, not changing the actual plugin code to your needs, since this is a generic implementation.

mstuettgen:
I get your point, but since your design is completely for copters / planes I guess, I need to find a way to test it and to find a good working solution. The problem is, I do not now how I can access the set_position_target_local_ned in the firmware yet, but I will find out soon.

For now, i will test with my attitude-hack and try to understand how it works.

Did you look at my attitude_roll_pitch_yaw_throttle() method? 

my intention is that the roll pitch yaw and throttle values I send via mavlink get into the attitute_setpoint_targets values 
att_sp->roll_body
att_sp->pitch_body
att_sp->yaw_body
att_sp->throttle

Even if that does not make sense, should it work in theory? Just for my understanding... Thx!

mstuettgen:
<code c++> #include<iostream></code>

mhkabir:
Apologies, but this isn't making any sense at all. 
You just need yaw rate (or plain yaw) and throttle for the rover. Or am I wrong with this basic assumption?
You can easily use the setpoint_velocity plugin for that. 
Otherwise, if you want to send position targets, you can use the setpoint_position plugin and take the data into your controller from the corresponding uORB struct in PX4.

mstuettgen:
Apologies, but as I said before, the question is not if that makes sense, I want to understand how it works. 
And to answer your question: yes, I need plain yaw and plain thrust for my rover, nothing else. The rover_steering app code is a code template I got from Lorenz Meier and we are trying to find out the best way to control a rover. Thanks for your help!

mhkabir:
^edited

mstuettgen:
And since my template code only give me access to setpoint_attitute_target (att_sp), I am hacking my way into all this stuff.... Since setpoint_velocity sends its data to local_position_ned, and I have no access to it in my app, the only way this can work somehow is using setpoint_attitude. Do you understand what i mean ?

mhkabir:
OK, I went through your steering controller code. Basically, you can use the current attitude plugin. Send the yaw in the quaternion. Send the throttle separately. Done. No code changes required in the plugin or PX4.
As the your plugin edit stands now, you are sending yaw velocities, NOT the plain yaw. Use the quaternion for that, instead of the velocity fields.

mstuettgen:
Thanks a lot ! But this would require sending a TF for the yaw on /mavros/setpoint/attitude and a float on /mavor/att_thrust, correct?

mstuettgen:
And.... I am wondering what the quaternion for max steering to the left e.g. would have to look like

mstuettgen:
i can use tf::createQuaterionFromYaw for that, but what yaw values should i feed in?

mhkabir:
Correct.
A TF would work, or just a PoseStamped or PoseWithCovarianceStamped with only yaw data encoded into the quaternion. (You can use TF functions for the Euler -> Quaternion conversion)

mhkabir:
Erm, that's up to you. Seemingly you can use anything from 0 - 360 degrees, I guess... ;)

LorenzMeier:
@mstuettgen The low-level controller closes the loop. You just provide the desired heading. If you want to directly control actuators you should take a completely different route and send the newly added actuator control messages, which are directly passed down to the outputs.

mstuettgen:
I will try with quaternion and give then feedback. If that works I think thats fine for now. Thanks!

mstuettgen:
So the max steering angle is about 45° in my car, i will try that as a starting point

LorenzMeier:
The attitude control is world frame, not body frame. So you would put the desired heading in there and the low-level controller will try to reach it. If you do it in body frame you need to take care of latency, since you can get oscillations due to the world -> body transformation at different time steps. You don't have that problem in the global frame.

mstuettgen:
I read something about mavros taking care about the ROS <-> pixhwak frame conversion ? Or are you talking about something completely different?

mhkabir:
Marcel, the yaw you'll be sending from ROS will be with respect to the
world frame. It would be zero degrees pointing North and increase
anticlockwise if I've got the ROS standard correct.

It is irrespective of the maximum turning angle of your car. The controller
will keep trying to turn  until the desired world-frame heading is
achieved. Servo actuation would be through a PID loop.

For rovers, this would imply that the controller would get saturated
without forward motion, since unlike a copter or plane, yawing isn't
possible without moving forward for a rover.

mstuettgen:
so the steering via attitude is kind of "absolute" if you know what i mean?

mhkabir:
Yes exactly

mstuettgen:
ok i will try that... I need to figure something out to make it a virtual "relative" steering then. If that doesnt work i would be really interested in accessing the actuators directly

mhkabir:
If you want direct servo/ESC control, you should use the new actuator control messages which are passed to the outputs, as Lorenz said.
There is no implementation of it in mavros at the moment , so you'd have to implement it yourself.

mstuettgen:
The more I think about that, thats probably the best way to go for it. I can try to start the implementation if you give me a starting point

mstuettgen:
I am gonna test the yaw quaternion anyway, can u maybe check my px4_config.yaml ?

It should now listen to attitude TF and the thrust via /setpoint/att_thrust

https://github.com/mstuettgen/maskor_rover/blob/master/rover_bringup/launch/px4_config.yaml

mhkabir:
Looks okay.
I've opened #229 for tracking the actuator control plugin.

mhkabir:
And just in case, use a clean Mavros master build for testing since no changes should be required. Also reduces the chance of spurious code mess ups.

mstuettgen:
thomasgubler just said that the mavlink receiver app already supports this command, so the implementation only needs to be done on mavros side.

Is there any way to access http://mavlink.org/messages/common#MANUAL_CONTROL via mavros? thomasgubler just suggested that

thomasgubler:
@mstuettgen This is what I meant https://github.com/PX4/Firmware/pull/1836

mstuettgen:
So we then would need an implementation for mavros to use the MANUAL_CONTROL message ? Or is done already?

PR #256, opened 2015-03-23
node: ignore messages not sent by the target uas

claymation:
This change is intended to prevent heartbeat messages sent
by a GCS from interfering with those sent by the target UAS.

### DIFF
--- a/mavros/include/mavros/mavros_uas.h
+++ b/mavros/include/mavros/mavros_uas.h
@@ -152,6 +152,12 @@ public:
 		target_component = comp;
 	}
 
+	/**
+	 * @brief Return true if this message was sent by this UAS
+	 */
+	inline bool is_sender(uint8_t sys, uint8_t comp) {
+		return (target_system == sys && target_component == comp);
+	}
 
 	/* -*- IMU data -*- */
 	void update_attitude_imu(tf::Quaternion &q, tf::Vector3 &av, tf::Vector3 &lacc);
--- a/mavros/src/lib/mavros.cpp
+++ b/mavros/src/lib/mavros.cpp
@@ -88,6 +88,11 @@ MavRos::MavRos() :
 			.unreliable()
 			.maxDatagramSize(1024));
 
+	mav_uas.set_tgt(tgt_system_id, tgt_component_id);
+	UAS_FCU(&mav_uas) = fcu_link;
+	mav_uas.sig_connection_changed.connect(boost::bind(&MavlinkDiag::set_connection_status, &fcu_link_diag, _1));
+	mav_uas.sig_connection_changed.connect(boost::bind(&MavRos::log_connect_change, this, _1));
+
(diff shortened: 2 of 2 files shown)


claymation:
The reason for this PR (and #254) is that my ROS application inspects the "mode" field of the /mavros/state message to determine what it should do: if the FCU (APM) is in GUIDED mode, my code will do one thing, and if the FCU is in any other mode, my code will wait pending further instructions. This is useful for being able to toggle functionality using a flight mode switch on my RC transmitter.

The problem is that my FCU also has attached to it a 3DR radio, over which it receives MAVLink packets from APM Planner. The APM code, for better or worse, forwards all heartbeat messages across all links, causing mavros to receive the heartbeats from APM Planner in addition to those from APM. After #254, the mode strings will be correct, but they will still confuse my code, as it oscillates between `GUIDED` (from FCU) and `CMODE(0)` (from GCS).

This change prevents that by filtering out foreign messages upon receipt.

vooon:
This change breaks `3dr_radio` plugin. Also seems that you breaks UART with multiple connections.
UART (RS232) is point-to-point.

Better connect your modem to OBC and then use `gcs_url:=serial://device`.

claymation:
Yeah, I'm sure this "fix" has many problems, but I don't think that putting the telemetry modem behind mavros is a fix, either. The telemetry modem provides out-of-band monitoring and control should anything go wrong with my code, mavros, or the OBC. Besides, the Pixhawk has multiple serial ports and APM & PX4 support multiple serial links for a reason.

Really what I'm not understanding is, if MAVLink can support up to 255 devices, why do most GCS's seem to assume there will only be one FCU on the link? mavproxy.py has the same problem that mavros has, namely that it assumes that every heartbeat message it hears is coming from the target system, without actually verifying that.

So I don't know where exactly the fix belongs, but I don't think it belongs in hardware.

vooon:
Hm, you mean that you used TELEM1/2 for OBC and modem? But why OBS listens modem messages? Only if APM route it internally, maybe disable this feature?

Initially mavros designed for onboard computer, where two FCU on same link is very unusual.
As i know only QGC, and based on it APM Planner 2 will support multiple UAS connections.

claymation:
Correct, TELEM1 is connected to the radio and the USB UART is connected to mavros. APM automatically forwards heartbeats across all links—there is currently no way to disable this:

https://github.com/diydrones/ardupilot/blob/master/libraries/GCS_MAVLink/MAVLink_routing.cpp#L191

I am happy to submit a pull request to them to make the forwarding optional. It serves no purpose to me, and seems like a waste of radio bandwidth, but I'm sure they have a reason for it.

vooon:
Please try my latest changes.

vooon:
Any update?

claymation:
Yes, this works great, thanks! Sorry for the tardy reply.

samacumen:
How can one change System_Id and Component_Id on Mavros?

PR #292, opened 2015-05-11
DISTANCE_SENSOR support

TSC21:
Code is done! Tested the readings with a `rosserial_python` node which was retrieving data from an Arduino into a Range type of message, which is read by the plugin usage. So now this only lacks implementation on the FCU side, which includes receive and send and publish/advertise it a DISTANCE_SENSOR uOrb topic.


mavros_extras/src/plugins/distance_sensor.cpp
```diff
@@ -0,0 +1,261 @@
...
+		// TODO: 
+		//		- variable topic advertising depending on the rangefinder type;
+		//		this means incrementing number of the publishing topics and change their names according to
+		//		the the type of sensor and its ID. p.e.: topics "laser_distance_1", "laser_distance_2"...
+
+		if(rangefinder_type == 0)
+			dist_laser_in_pub = dist_nh.advertise<mavros_extras::LaserDistanceSensor>("laser_distance", 10);
+		else if (rangefinder_type == 1)
+			dist_sonar_in_pub = dist_nh.advertise<mavros_extras::SonarDistanceSensor>("sonar_distance", 10);
+		else ROS_ERROR_NAMED("rangefinder", "Invalid rangefinder type! Valid values are 0 (laser) and 1 (ultrasound)!");
```
vooon: Fix debug name.

mavros_extras/src/plugins/distance_sensor.cpp
```diff
@@ -0,0 +1,261 @@
...
+				covariance);
+		UAS_FCU(uas)->send_message(&msg);
+	}
+
+	/* -*- mid-level helpers -*- */
+
+	/**
+	 * Receive distance sensor data from FCU.
+	 */
+	void handle_dist_sensor(const mavlink_message_t *msg, uint8_t sysid, uint8_t compid) {
```
vooon: Fix style: `handle_distance_sensor()`

mavros_extras/src/plugins/distance_sensor.cpp
```diff
@@ -0,0 +1,261 @@
...
+			dist_laser_msg->laser.angle_min = 0.0;
+			dist_laser_msg->laser.angle_max = 0.0;
+			dist_laser_msg->laser.angle_increment = 0.0;
+			dist_laser_msg->laser.time_increment = 0.0;
+			dist_laser_msg->laser.scan_time = 0.0;
+			dist_laser_msg->laser.range_min = laser_min_range/1E2; 
+			dist_laser_msg->laser.range_max = laser_max_range/1E2;
+			dist_laser_msg->laser.ranges[0] = dist_sen.current_distance/1E2;
+			
+			for (int i = 1; i < dist_laser_msg->laser.ranges.size(); i++) {
```
vooon: `std::fill`.

vooon: Oh, no. It's `std::vector<float>`, fill not needed because size shall be 1.

mavros_extras/src/plugins/distance_sensor.cpp
```diff
@@ -0,0 +1,261 @@
...
+			dist_laser_msg->header = header;
+			dist_laser_msg->time_boot_ms = dist_sen.time_boot_ms;
+
+			dist_laser_msg->laser.header = header;
+			dist_laser_msg->laser.angle_min = 0.0;
+			dist_laser_msg->laser.angle_max = 0.0;
+			dist_laser_msg->laser.angle_increment = 0.0;
+			dist_laser_msg->laser.time_increment = 0.0;
+			dist_laser_msg->laser.scan_time = 0.0;
+			dist_laser_msg->laser.range_min = laser_min_range/1E2;
```
vooon: Why not just store float?

mavros_extras/src/plugins/distance_sensor.cpp
```diff
@@ -0,0 +1,261 @@
...
+
+	/**
+	 * Receive distance sensor data from FCU.
+	 */
+	void handle_dist_sensor(const mavlink_message_t *msg, uint8_t sysid, uint8_t compid) {
+		mavlink_distance_sensor_t dist_sen;
+		mavlink_msg_distance_sensor_decode(msg, &dist_sen);
+
+		std_msgs::Header header;
+		header.stamp = ros::Time::now();
```
vooon: I think here should be `synchronize_stamp()`

mavros_extras/src/plugins/distance_sensor.cpp
TSC21: @vooon what do you mean?

mavros_extras/src/plugins/distance_sensor.cpp
```diff
@@ -0,0 +1,261 @@
...
+	}
+
+	// XXX TODO: Calculate laser/sonar covariances (both in FCU and on the ROS app side)
+
+	// XXX TODO: Determine FOV for Sonar
+
+	/**
+	 * Send laser rangefinder distance sensor data to FCU.
+	 */
+	void send_laser_dist_sensor_data(const ros::Time &stamp, uint16_t min_distance,
```
vooon: Those two wrapper functions not needed.

mavros_extras/src/plugins/distance_sensor.cpp
TSC21: `synchronize_stamp()` ??

mavros_extras/src/plugins/distance_sensor.cpp
TSC21: Got it!

mavros_extras/src/plugins/distance_sensor.cpp
TSC21: Got it!

vooon:
Did you used [uncrustify](https://github.com/mavlink/mavros/tree/master/mavros#contributing)? Also i don't understand why you need non-standard messages.

Also i think that LASER type (LidarLite) is more like INFRARED range finders, rather than LaserScan sources.
So laser branch not needed (and as i remember there already exists filter node for Range->LaserScan)

mavros_extras/src/plugins/distance_sensor.cpp
vooon: Look UAS class.

mavros_extras/src/plugins/distance_sensor.cpp
vooon: For what purpose do divide 1e2?

TSC21: Yeah you are right, since the value from the FCU has to come in centimeters already.

TSC21:
1. No I didn't. Will do it ASAP! The non-standard messages is just to keep the message definition, plus th id of the sensor, orientation and covariance.

2. Hmm I started with that, but then I thought that in the future someone could add a rangefinder scanner and then they could use this definition instead. But, should I keep it or move to Range instead?

3. filter node for Range->LaserScan? Where?

vooon:
1.
2. Better is move to standard Range.msg. Instead of abusing id you may dynamically add/delete publishers and subscribers (connected change). Or predefine map in params (but it little bit harder, see diagnostics aggregator)
3. I think that i read somewhere about similar filter, maybe i wrong.

vooon:
Hmm, also i think we may ask for adding LASER range finder to Range enum.

TSC21:
Ok I will move to Range.msg then. Oredefine map in params is what I want to do. Can you give an hand here?

About the Range enum, how can we ask that?

mavros_extras/src/plugins/distance_sensor.cpp
TSC21: Yeah the wrappers are not needed if we use the same msg.

mavros_extras/src/plugins/distance_sensor.cpp
TSC21: Wait, what's the problem with the log name? :D

mavros_extras/src/plugins/distance_sensor.cpp
```diff
@@ -0,0 +1,261 @@
...
+						laser_orientation,
+						sonar_covariance);	
+	}
+
+	/**
+	 * Send laser rangefinder distance sensor data to FCU.
+	 */
+	void send_sonar_dist_sensor_data(const ros::Time &stamp, uint16_t min_distance,
+								uint16_t max_distance, uint16_t current_distance) {
+		distance_sensor(stamp.toNSec() / 1000000,
```
TSC21: Should `uas->synchronise_stamp()` also be used here and how?

vooon: No, synchronize needed for data comes from FCU. To FCU send current time (in us or ms).

vooon:
Asked at ros-users: http://lists.ros.org/pipermail/ros-users/2015-May/069403.html

TSC21:
@vooon changes done. Using INFRARED type for Laser for now. We'll have to wait for ROS community to implement new type for Range.msg enum.

mavros_extras/msg/DistanceSensor.msg
```diff
@@ -0,0 +1,14 @@
...
+
+sensor_msgs/Range range
+
+uint8 type			# see MAV_DISTANCE_SENSOR enum
+uint8 MAV_DISTANCE_SENSOR_LASER = 0
+uint8 MAV_DISTANCE_SENSOR_ULTRASOUND = 1
+
+uint8 id 			# XXX FIXME enum ??
+
+uint8 orientation 	# rad
```
vooon: Rad? I don't think so. Uint8 is more likely orientation enum used in params for ahrs.

mavros_extras/msg/DistanceSensor.msg
```diff
@@ -0,0 +1,14 @@
...
+uint8 covariance	# 1E-2 meters; 0 for unknown/invalid readings
```
vooon: `variance`, and better use float32 + convert to Si.

mavros_extras/msg/DistanceSensor.msg
```diff
@@ -0,0 +1,14 @@
+# DISTANCE_SENSOR message data
+
+Header header
```
vooon: Already in Range.

mavros_extras/src/plugins/distance_sensor.cpp
```diff
@@ -0,0 +1,244 @@
...
+	void initialize(UAS &uas_)
+	{
+		uas = &uas_;
+
+		dist_nh.param<std::string>("frame_id", frame_id, "distance_sensor");
+
+		dist_nh.param("input/rangefinder_type", rangefinder_type, 0);	// default = laser (0)
+
+		//Default sonar rangefinder is Maxbotix HRLV-EZ4
+		dist_nh.param("input/sonar_fov", sonar_fov, 0.0);	// TODO
```
vooon: I think that parameters should be reworked. Will explain in regular comment.

vooon:
I still think that new message not needed. Only one really needed data is variance.
While id, type and orientation constant (per id).

May be better do following param structure:
```yaml
distance_sensor:
  sensor_topic_name:
    id: 0 # sensor id
    frame_id: some_frame
  other_sensor_topic:
    id: 1
    frame_id: other_frame
  subscribed_topic:
    subscribe: true
    id: 42
    type: 0 # we may try to guess by Range.radiation_type 
    variance: ???
    orientation: 0
```

Pro:
- regular Range.msg, so you can use rviz plugins (not sure that it can viz it now, but i remember some demo few years ago).
- different frame_id's, so you can connect your urdf model with data.
- no need to filter sensors by id in client nodes

mavros_extras/msg/DistanceSensor.msg
TSC21: OK

mavros_extras/msg/DistanceSensor.msg
TSC21: Probably. Will change it then

mavros_extras/msg/DistanceSensor.msg
TSC21: https://github.com/mavlink/c_library/blob/master/common/mavlink_msg_distance_sensor.h#L14 I'm just using the definition. But I agree with `variance`, since we are talking about one unique dimension, which is `z`. Unless we consider the derivative and also consider speed measurements, which considers also time and `vz`. So, should I change to `variance` then and warn PX4 dev team to change message definition?

TSC21:
@vooon,
1. Besides the `variance`, we also have the `orientation`. So to keep just a default Range.msg, and to use the full data of the Mavlink message I purpose this:
* [ ] - send the `variance` in a `Float32` message format, so to use in another node
* [ ] - use the received `orientation` so to publish the transform between the rangefinder frame and the FCU frame
What do you think?

2. I was considering dynamically add/delete publishers, depending on the rangefinder type and its id of the data coming from the FCU. But I'm not really sure how to do it. Can you give a hand here?

TSC21:
BTW, here's the RViz representation you were talking about: http://wiki.ros.org/rviz/DisplayTypes/Range

vooon:
@TSC21 You don't need to use orientation, because it already coded in frame_id + transform (i think from URDF).
But maybe add check FCU orientation == plugin config is not bad. Because orientation **FIXME enum** :) may differ from definition.

Maybe for now we may drop variance?  Or show in diagnostics. At least until someone not need it.


I think instead of dynamic add/del better specify topics in config file (px4_config.yaml) and provide mapping to id (plus warning messages about unmapped id's).
I think that it should be easier to work with, but harder in rosparam walk.
See example https://github.com/mavlink/mavros/blob/master/mavros/src/plugins/param.cpp#L796

mavros_extras/msg/DistanceSensor.msg
```diff
@@ -0,0 +1,14 @@
...
+
+sensor_msgs/Range range
+
+uint8 type			# see MAV_DISTANCE_SENSOR enum
```
tfoote: This is already in the Range message

mavros_extras/msg/DistanceSensor.msg
```diff
@@ -0,0 +1,14 @@
...
+uint8 MAV_DISTANCE_SENSOR_LASER = 0
+uint8 MAV_DISTANCE_SENSOR_ULTRASOUND = 1
+
+uint8 id 			# XXX FIXME enum ??
```
tfoote: You should use the frame_id from the header to know what sensor it came from.

mavros_extras/msg/DistanceSensor.msg
tfoote: This also should be captured in the frame_id using tf information.

tfoote:
I agree with @vooon to not create a new message datatype. The variance is not currently captured in the Range message, but that should not be a problem to parameterize properly when processing. If we need it we can consider adding it to the Range message, however from experience it tends to be application specific more than device specific. Similarly the [LaserScan](http://docs.ros.org/api/sensor_msgs/html/msg/LaserScan.html) does not have variance/covariance.

TSC21:
@tfoote thanks for your feedback. Yes, I will for sure just stick to the `Range.msg`. That me and @vooon was asking to be added to the msg definition is the `radiation_type` `LASER`, as `uint8 LASER=2`.
Adding the covariance to the msg wouldn't be a bad add, but as you said it is more specific to the application that to the device. Still though, it can be used or not, so having it there could be good.

mavros_extras/msg/DistanceSensor.msg
TSC21: In this case, we are translating the message from the MAVLink frame to the ROS msg, so it should be the other way (i think): knowing the type of data received in the MAVLink frame, it should define what's the sensor frame, and as a consequence, the `radiation_type`. Correct me if I'm wrong.

mavros_extras/msg/DistanceSensor.msg
TSC21: Yes you are right. But since it won't be used, I will delete this custom msg. Anyway, it's useful info for future implementation, so thanks :)

TSC21:
> @TSC21 You don't need to use orientation, because it already coded in frame_id + transform (i think from URDF).

What do you mean? What I was thinking is dependent of the orientation (which is a data topic that is sent and configured in the FCU), then a conditional statement could be implemented, where this orientation gives a static transform between the FCU and the rangefinder.

> But maybe add check FCU orientation == plugin config is not bad.
How? (code sample, pls (:  )

> Maybe for now we may drop variance? Or show in diagnostics. At least until someone not need it.
We can stick to the `Float32.msg` for now, unless @tfoote adds it to the `Range.msg` def.

> I think instead of dynamic add/del better specify topics in config file (px4_config.yaml) and provide mapping to id (plus warning messages about unmapped id's).
-| Help ? ? |-

tfoote:
@TSC21 The laser sensors use infrared light beams, we should use the INFRARED enum. 

I would suggest we use the message as is. All the messages in sensor_msgs have been through a review and are stable. If through use we determine that the variance is important to keep with the data we can propose a change to the message, but until we have a strongly demonstrated need we don't want to speculatively change the message.

TSC21:
@tfoote Lidar uses ultraviolet, visible, or near infrared light beams. Mostly near-infrared. So it's not mainly infrared. That's why we consider it as a different source.

OK, I agree then to keep the message as is. Will publish covariance as Float32 then.

TSC21:
I reworked the PR:
* Using just `Range.msg` + `Float32.msg` for covariance
* Now `frame_id` name is dynamically changed according to the received `id` and `type` of the sensor: https://github.com/mavlink/mavros/pull/292/files#diff-2adf7423447a0e5f4880b56161efdda9R155

vooon:
I think i can write param parser part. So i merging this PR now, but without my changes it not complete.

TSC21:
Thanks @vooon!
What about the orientation part? We are kinda dependent on the enum regrading the orientation, which is part of the MAVLink message definition.
Since there is no enum implemented, we can imagine that it can be something related to common angle deviation from the euclidean axis: like +45 degrees from Z, etc. Is there anything similar, besides the Mag orientation, on the PX4 firmware, that can we base in, and that you are aware of?

vooon:
@TSC21 i'm done. But it little incomplete, but not that not big deal.

Example test config:
```yaml
mavros:
  plugin_whitelist: ["sys_*", distance_sensor]
  distance_sensor:
    topic_pub_1:
      id: 0
      frame_id: sensor_1
    topic_pub_2:
      id: 1
      frame_id: sensor_2
      orientation: 2
    topic_sub_1:
      subscriber: true
      id: 2
      orientation: 2
```

But i don't tested recv/send (need FCU or emulator).

TSC21:
@vooon why have you changed all the code structure?

vooon:
@TSC21 because i done param to map, which requires many instances of pub/sub
i moved part of code to item class. Then i cleaned up, so from your code now left only skeleton :)

TSC21:
OK... but then I think the colaborative part somewhat disappeared on this. I want to give some contribution but like this is like doing nothing. You can add it as your own PR then.

vooon:
Sorry, but i done what i mentioned yesterday (topic-id-frame_id mapping).
Also I really don't have enough time to test, and i don't do some params (fov etc, see XXX comments).
Code now in master, you can further fix it by next PR.

TSC21:
Ok. Sent you a PM in GHangout regarding this. Anyway, I'll see what I can do about other stuff. Also I don't know why you deleted covariance publishing. We didn't agreed to drop it down.

TSC21:
@vooon, regarding this new config:
I have the following in `px4_config.yaml`:
```
distance_sensor:
  sonar_1_pub:
    id: 0
    frame_id: sonar_1
    orientation: 1
  laser_1_pub:
    id: 1
    frame_id: laser_2
    orientation: 1
  sonar_1_sub:
    subscriber: true
    id: 2
    orientation: 1
```
If in the `.ino` file I create a publisher like this: `ros::Publisher pub_range("/sonar_1_sub", &range_msg);` with `frame_id = 2`, is it suppose to subscribe? Cause mavros is not subscribing to this topic.
In the other way, if I use `ros::Publisher pub_range("/mavros/distance_sensor/sonar_1_sub", &range_msg);` then the topic subscribes, but then if I set `frame_id = 1` in the Arduino code, it still subscribes, but with `frame_id=1`, but it shouldn't cause `frame_id` != `id` in the config right?

vooon:
Subscriber not used frame_id, just use right topic (`/mavros/distance_sensor/sonar_1_sub`, look at `rosnode info mavros`).
And it should send `DISTANCE_SENSOR.id=2` for any incoming message on that topic.

TSC21:
OK, understood. Adding some enhancements to code now (variance calculation, etc.). Then I'll do PR.

TSC21:
Trying to troubleshoot an error, which is not allowing me to load the `mavros_extras` plugins. This is what I added to `px4_config.yaml`:
```
#distance_sensor
distance_sensor:
  hrlv_ez4_pub:
    id: 0
    frame_id: "hrlv_ez4_sonar"
    orientation: 1      # Default for Maxbotix HRLV-EZ4; in meters
    min_range: 0.3
    max_range: 5.0
    field_of_view: 0.0  # XXX TODO
  lidarlite_pub:
    id: 1
    frame_id: "lidarlite_laser"
    orientation: 1      # Default for PulsedLight LIDARLite; in meters
    min_range: 0.0
    max_range: 40.0
    field_of_view: 0.0  # XXX TODO
  sonar_1_sub:
    subscriber: true
    id: 2
    orientation: 1
  laser_1_sub:
    subscriber: true
    id: 3
    orientation: 1
```
(I put the min/max range and fov as needed param for pub).

I get this when launching the node:
```
[ERROR] [1431615369.006175623]: Plugin distance_sensor load exception: Failed to load library /home/nuno/AIMAV_Project/devel/lib//libmavros_extras.so. Make sure that you are calling the PLUGINLIB_EXPORT_CLASS macro in the library code, and that names are consistent between this macro and your XML. Error string: Could not load library (Poco exception = /home/nuno/AIMAV_Project/devel/lib//libmavros_extras.so: undefined symbol: _ZN9mavplugin20DistanceSensorPlugin8readingsE)
[ERROR] [1431615369.113900603]: Plugin mocap_pose_estimate load exception: MultiLibraryClassLoader: Could not create object of class type mavplugin::MocapPoseEstimatePlugin as no factory exists for it. Make sure that the library exists and was explicitly loaded through MultiLibraryClassLoader::loadLibrary()
[ERROR] [1431615369.166365302]: Plugin px4flow load exception: MultiLibraryClassLoader: Could not create object of class type mavplugin::PX4FlowPlugin as no factory exists for it. Make sure that the library exists and was explicitly loaded through MultiLibraryClassLoader::loadLibrary()
[ERROR] [1431615369.305468753]: Plugin vision_pose_estimate load exception: MultiLibraryClassLoader: Could not create object of class type mavplugin::VisionPoseEstimatePlugin as no factory exists for it. Make sure that the library exists and was explicitly loaded through MultiLibraryClassLoader::loadLibrary()
[ERROR] [1431615369.359173584]: Plugin vision_speed_estimate load exception: MultiLibraryClassLoader: Could not create object of class type mavplugin::VisionSpeedEstimatePlugin as no factory exists for it. Make sure that the library exists and was explicitly loaded through MultiLibraryClassLoader::loadLibrary()
[ERROR] [1431615369.411541570]: Plugin visualization load exception: MultiLibraryClassLoader: Could not create object of class type mavplugin::VisualizationPlugin as no factory exists for it. Make sure that the library exists and was explicitly loaded through MultiLibraryClassLoader::loadLibrary()
```
Solutions?

mhkabir:
Small note : The PX4Flow sonar doesn't publish to the PX4 ```distance_sensor``` uORB topic (Hence no mavlink message sent) . There will be no output for it.
The LL040S, SF0x, etc. do. The PX4Flow plugin already handles the sonar publishing. ;) 

Remove the HRLV-EZ4 please, or follow up with a PR to Firmware which publishes the sonar to the distance sensor topic and also the required modifications to all the estimators to use the topic.

TSC21:
> Small note : The PX4Flow sonar doesn't publish to the PX4 distance_sensor uORB topic (Hence no mavlink message sent) . There will be no output for it.
The LL040S, SF0x, etc. do. The PX4Flow plugin already handles the sonar publishing. ;)

I'm aware of it thanks ;)

> follow up with a PR to Firmware which publishes the sonar to the distance sensor topic and also the required modifications to all the estimators to use the topic.

It's my next step. Also I'm planning to create a .h with various sensor definitions to be used in the config.

Anyway, @mhkabir do you have any idea of what can be firing the errors above?

vooon:
@TSC21 try `catkin clean --all && catkin build`

TSC21:
`catkin clean --all && catkin build` is not recognized. But doing `catkin_make clean` and then `catkin_make` didn't help.

TSC21:
OK I installed `catkin build` tools so to use the above commands. But after compiling everything I still get the same errors!

TSC21:
Well, it is solved. Had to clone repo again and add changes. Will do PR now.

vooon:
Commit changes to some branch.

TSC21:
@vooon here it is: https://github.com/mavlink/mavros/commit/f5c411340abefcc414e662b6ff10e6a64589a68a.

vooon:
Commented in diff. Main problem source is static class members you added. Why you ever need static?

PR #311, opened 2015-06-16
LANDING_TARGET coder and decoder

TSC21:
This allows to receive and send data through `LANDING_TARGET` message. I convert the angles and distance (from which I compute the magnitude of XYZ vector) into `PoseStamped` messages or publish transform, and vice-versa. Please check the math so everything is OK.


LorenzMeier:
Is this tested?

TSC21:
No. That' why support on firmware is added above in the todo list

TSC21:
@vooon can you verify the consistency of the plugin code / applied math, please?

mavros_extras/src/plugins/landing_target.cpp
```diff
@@ -0,0 +1,198 @@
...
+		land_target_pub = sp_nh.advertise<geometry_msgs::PoseStamped>("landing_target", 10);
+
+		if (listen_tf) {
+			ROS_INFO_STREAM_NAMED("landing_target", "Listen to landing_target transform " << frame_id
+											<< " -> " << child_frame_id);
+			tf_start("LandingTargetTF", &LandingTargetPlugin::send_landing_target);
+		}
+
+		land_target_sub = sp_nh.subscribe("target", 10, &LandingTargetPlugin::land_target_cb, this);
+		local_position_sub = sp_nh.subscribe("/mavros/position/local", 10, &LandingTargetPlugin::local_pos_cb, this);
```
vooon: I don't like that, at lest it broke node renaming.
I don't decide best way yet, but at least use `~position/local` now.

mavros_extras/src/plugins/landing_target.cpp
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	const message_map get_rx_handlers() {
+		return {
+			       MESSAGE_HANDLER(MAVLINK_MSG_ID_LANDING_TARGET, &LandingTargetPlugin::handle_landing_target)
+		};
+	}
+
+private:
+	friend class TFListenerMixin;
```
vooon: Also you sure that you actually need TF here? Maybe for initial debugging use topics only?

mavros_extras/src/plugins/landing_target.cpp
TSC21: Ok will change it then

mavros_extras/src/plugins/landing_target.cpp
TSC21: Actually i use tf as a good debugging tool, mainly on Rviz :) but users have the param option to use it or not

mavros_extras/src/plugins/landing_target.cpp
```diff
@@ -0,0 +1,198 @@
...
+				frame,
+				angle_x,
+				angle_y,
+				distance);
+		UAS_FCU(uas)->send_message(&msg);
+	}
+
+	/* -*- mid-level helpers -*- */
+
+	/*	AUXILIAR MATH:
```
TSC21: @LorenzMeier do you confirm this math is acceptable?

mavros_extras/src/plugins/landing_target.cpp
```diff
@@ -0,0 +1,198 @@
...
+		mavlink_msg_landing_target_decode(msg, &land_target);
+
+		ROS_DEBUG_THROTTLE_NAMED(10, "land_target", "Landing target: "
+				"frame: %d angle offset:(X: %1.3frad, Y: %1.3frad) distance: %1.3fm",
+				land_target.frame, land_target.angle_x, land_target.angle_y, land_target.distance);
+
+		tf::Vector3 target;
+		tf::Vector3 local = transform.getOrigin();
+
+		target.setX(sqrt((pow(land_target.distance,2) - pow(local.z(),2)) * pow(cos(land_target.angle_x),2)));
```
TSC21: @LorenzMeier here I use local_position.z to calculate the other variables. Is it acceptable or should ground_distance be used instead?

mavros_extras/src/plugins/landing_target.cpp
TSC21: @rmackay9 ping!

mavros_extras/src/plugins/landing_target.cpp
LorenzMeier: You should use the target distance, not the Z value. It doesn't have to be at the ground. And don't use pow(x,2) - that's extremely inefficient. Use x*x for x^2.

TSC21: @LorenzMeier please look at https://github.com/mavlink/mavros/pull/311/files#diff-ab77f3e2037c4a99b3fd054f54e852c4R112. I have there the auxliar math to calculate target "distance". The Z value is used to compute the X and Y components, together with the angles. I can't see another way of determining this without decomposing the distance vector, which magnitude is the represented by the float value `distance`.

TSC21: > And don't use pow(x,2) - that's extremely inefficient. Use x*x for x^2.

Ok thanks for the tip. Will change it then.

mavros_extras/src/plugins/landing_target.cpp
```diff
@@ -0,0 +1,198 @@
...
+	 * Send landing target transform to FCU
+	 */
+	void send_landing_target(const tf::Transform &transf, const ros::Time &stamp) {
+		// origin position in ROS ENU frame
+		tf::Vector3 pos = transf.getOrigin();
+
+		float ang_x = atan(pos.y()/pos.x());
+		float ang_y = boost::math::constants::pi<double>()/2 - atan(pos.y()/pos.x());
+		float distance = sqrt(pow(pos.x(),2) + pow(pos.y(),2) + pow(pos.z(),2));
+
```
TSC21: @LorenzMeier the above represent the calculations of X,Y offset angles and distance to target based on the target position. This was the way I see we can get this values, based on simple trigonometry.

mavros_extras/src/plugins/landing_target.cpp
TSC21: OK I think it's better to present the calculations this way:
```c++
distance = sqrt(target_local_position.x^2 + target_local_position.y^2 + target_local_position.z^2)
target_local_position.x = sqrt((distance^2 - target_local_position.z^2) * cos^2(angle_x))
target_local_position.y = target_local_position.x * tan(angle_x)
target_local_position.z = distance^2 - target_local_position.x^2 - target_local_position.y^2;
```
If we assume that `distance` is the computed magnitude of a X,Y,Z vector, the above is correct. What we know are the angle offsets of X and Y and this "distance". But we don't have nothing that gives us a Z related measurement (no angle or Z distance component) besides the Z distance of the vehicle frame. So how can we compute the target X,Y,Z location if we don't have all the minimum necessary values?

LorenzMeier: You have the direction and distance. That's really all you need. Your target distance calculation seems wrong - this is the distance the target tracker is providing.

TSC21: > You have the direction and distance.

I don't have the complete direction because I don't have the Z offset. How can I compute a tranform this way?

LorenzMeier: You have two angles and a distance. That's a 3D position in space in the body frame.
The position of the landing pad is where you want to land. It doesn't matter where it is in absolute 3D space and it will move. You can transform it using the current vehicle position in each time frame.

TSC21: > You have two angles and a distance. That's a 3D position in space in the body frame.

What about world frame? We still lack Z offset.

> It doesn't matter where it is in absolute 3D space and it will move.

If one wants to represent a transform in world frame between the vehicle position and the target location, it indeed is necessary. The same thing if we want to land. If don't know the height of the landing pad, how can one compute the transform if the vehicle height is different from the landing pad height. There's some misconception here I'm trying to make clear. There's no way in the current implementation we can get global 3D position of the landing pad, because we don't have the 3 direction components. That's why I was asking if there's the possibility of adding `angle_z` to IRLock driver and by consequence, to `LANDING_TARGET` msg.

TSC21: OK think of the angles: Is `angle_x` and `angle_y` complementary? Meaning that `angle_y = PI - angle_x`. Or is this `angle_x` what I'm referring as the `angle_z`?? Cause if yes, then we have all the necessary data.

TSC21: If I get this right, this representation is suppose to be a reference to angular offsets in spherical coordinates, which means that `angle_x` should be `theta` and `angle_y` should be `phi` and that way we have `position.x = r*sin(phi)*cos(theta)`, `position.y = r*sin(phi)*sin(theta)` and `position.z = r*cos(theta)`, where `theta` gets values from 0 to 2PI and theta from 0 to PI and `r` is the distance.
The current angle names led me to think they were complementary and that we were lacking an vertical offset. If this was present as spherical coordinates it would make much more sense.

rmackay9:
@TSC21, just fyi - i'm really not familiar with mavros and not a reviewer for mavlink so Lorenz is your man!

TSC21:
@rmackay9 i wasn't asking for mavros or mavlink review but for a revision of concepts/ representation, since you were responsible to bring up this msg and irlock plugin into the code. This is more like trying to setup a correct scientific concept that correcting something more techinal of the code.

mavros_extras/src/plugins/landing_target.cpp
LorenzMeier: I don't think the names can be changed since its already deployed, nor does it seem more obvious with the notation you're using. The comment in the message could probably be improved to explicitly state that these are spherical coordinates.

TSC21: What do you mean with deployed? Is there any platform firmware using Mavlink that already is using this message? If not, i think it's better to change the name notation than just improve the comments. Isn't just my opinion, but I think the current notation seems messing spherical coordinates with rectangular ones, as there also a specific convertion between this two. So I don't understand why do you say my notation is not more obvious when it is the correct one.

Anyway, I'll be changing the mavros plugin code so to correct the maths since now I understand those are spherical coordinates. I'll leave the mavlink message implementation/notation as is. Will let you decide if it is to be changed or not. My opinion is that it is not only not obvious, it is also incorrect.

TSC21:
@vooon considering that we can have more than one landing target identified (that's why they have ID's), probably have to implement something like we done on `distance_sensor` plugin, refarding dynamic subs/pubs and theire params, right?

LorenzMeier:
Actually I was implicitly considering the normalised camera coordinate frame, but realise now that I expressed this very ambiguous - its not "the" polar system and uses a different notation. This is what people would normally use to represent a general point in a camera:
https://en.wikipedia.org/wiki/Camera_matrix#Normalized_camera_matrix_and_normalized_image_coordinates

If we want to represent the target at an arbitrary angle the spherical system might be less confusing (although the normalised camera coordinates are also able to represent points behind the camera.

By now I was also able to verify that the message is not yet deployed and only in developer use, and that we still have the opportunity to "do the right thing". I'd like to get the buy-in of the IR-lock developers though. @rmackay9 It would be good to sync them to this discussion, so that we're not operating up in the air.

What do you think?

TSC21:
@LorenzMeier, I'm expecting to receive an IRLock soon, but more than using this with an IRLock, ir could be interesting to use with markers, which position is computed in a OBC.

TSC21:
@vooon need a new mavlink release, if you may. Thanks!

mavros/launch/apm_config.yaml
```diff
@@ -142,4 +142,29 @@ vision_speed:
 vibration:
   frame_id: "vibration"
 
+# landing_target
+landing_target:
+  send_tf: true
```
vooon: Please move tf params to `tf` subsection.
e.g.
```
landing_target:
  tf:
    send: true
    listen: false
    rate_limit: 10.0
    frame_id: a
    child_frame_id: b
```

mavros/src/lib/uas_stringify.cpp
```diff
@@ -30,21 +30,21 @@ typedef std::unordered_map<uint32_t, const std::string> cmode_map;
...
-	{ 6, "FBWB" },
-	{ 7, "CRUISE" },
-	{ 8, "AUTOTUNE" },
-	{ 10, "AUTO" },
-	{ 11, "RTL" },
-	{ 12, "LOITER" },
-	{ 14, "LAND" },		// not in list
-	{ 15, "GUIDED" },
-	{ 16, "INITIALISING" }
+{ 0, "MANUAL" },
```
vooon: Bad change.

mavros/src/lib/uas_stringify.cpp
```diff
@@ -316,3 +316,39 @@ std::string UAS::str_system_status(enum MAV_STATE st)
...
+std::string UAS::str_frame(enum MAV_FRAME fr)
+{
+	size_t idx = size_t(fr);
+	if (idx >= frame_strings.size())
+		return std::to_string(idx);
+
+	return frame_strings[idx];
+}
+
+uint8_t UAS::idx_frame(std::string mav_frame)
```
vooon: Better name it `frame_from_str()`.

mavros/src/lib/uas_stringify.cpp
```diff
@@ -316,3 +316,39 @@ std::string UAS::str_system_status(enum MAV_STATE st)
...
+{
+	uint8_t index, idx = 0;
+	for (const auto &frame : frame_strings) {
+		if (frame.compare(mav_frame) == 0)
```
vooon: Why you every time do `str1.compare(str2)` instead of simple `str1 == str2`?

mavros_extras/src/plugins/landing_target.cpp
```diff
@@ -0,0 +1,242 @@
...
+	/**
+	 * Send landing target transform to FCU
+	 */
+	void send_landing_target(const ros::Time &stamp, const Eigen::Affine3d &transf) {
+		// origin position in ROS ENU frame
+		auto pos = UAS::transform_frame_enu_ned(Eigen::Vector3d(transf.translation()));
+
+		float distance = Eigen::internal::psqrt(pos[0] * pos[0] + pos[1] * pos[1] + pos[2] * pos[2]);
+		float phi = atan(Eigen::internal::psqrt(pos[0] * pos[0] + pos[1] * pos[1]) / pos[2]);		// = angle_x
```
vooon: I don't think that calling Eigen::**internal** is a good idea.

vooon: Also use .x(), .y(), .z() or (0), (1), (2).

mavros_extras/src/plugins/landing_target.cpp
```diff
@@ -0,0 +1,242 @@
...
+		float theta = atan(pos[1] / pos[0]);		// = angle_y
+
+		float size_x_rad = target_size_x * phi;		// assuming this is the arc length of the circle in X-axis
+		float size_y_rad = target_size_y * theta;	// assuming this is the arc length of the circle in Y-axis
+
+		uint8_t frame = UAS::idx_frame(mav_frame);	// MAV_FRAME index based on given frame name
+		if (frame == -1) {
```
vooon: **u**int8 can't be -1, also do finding once in int. mav_frame don't change over time, so why eat cpu cycles with useless calls?

mavros_extras/src/plugins/landing_target.cpp
```diff
@@ -0,0 +1,242 @@
...
+		auto orientation = Eigen::Quaterniond(1,0,0,0);	// TODO : Set pose depending on MAV_FRAME enum
+
+		ROS_DEBUG_THROTTLE_NAMED(10, "land_target", "Landing target: "
+				"frame: %s angle offset:(X: %1.3frad, Y: %1.3frad) "
+				"distance: %1.3fm position:(%1.3f, %1.3f, %1.3f)",
+				mavros::UAS::str_frame(static_cast<enum MAV_FRAME>(land_target.frame)).c_str(),
+				phi, theta, distance, position[0], position[1], position[2]);
+
+		auto pose = boost::make_shared<geometry_msgs::PoseStamped>();
+		pose->header = uas->synchronized_header(tf_frame_id, land_target.time_usec);
```
vooon: Use different frame_id param here.

mavros_extras/src/plugins/landing_target.cpp
```diff
@@ -0,0 +1,242 @@
...
+			transform.transform.rotation = pose->pose.orientation;
+			tf::vectorEigenToMsg(position, transform.transform.translation);
+
+			uas->tf2_broadcaster.sendTransform(transform);
+		}
+
+		auto tg_size = Eigen::Vector3d(land_target.size_x / phi, land_target.size_x / (M_PI - phi), land_target.size_y / theta);
+		auto tg_size_msg = boost::make_shared<geometry_msgs::Vector3Stamped>();
+
+		tg_size_msg->header = uas->synchronized_header(tf_frame_id, land_target.time_usec);
```
vooon: Copy `pose->header`.

mavros/src/lib/uas_stringify.cpp
TSC21: Uncrustify did this :p

mavros/launch/apm_config.yaml
TSC21: Ok

mavros/src/lib/uas_stringify.cpp
TSC21: Well this was implemented some time ago, and I didn't check it this time. I got used to compare somewhat :)

mavros/src/lib/uas_stringify.cpp
TSC21: why `frame_from_str`? this one gives the index of the frame.

mavros_extras/src/plugins/landing_target.cpp
TSC21: > uint8 can't be -1

Well my bad... forgot that detail.

> also do finding once in int. mav_frame don't change over time

Good point.

mavros_extras/src/plugins/landing_target.cpp
TSC21: Why different if it is the same frame?

mavros_extras/src/plugins/landing_target.cpp
TSC21: Yep seems simpler. thanks

TSC21: Ok I understood the purpose.

mavros_extras/src/plugins/landing_target.cpp
TSC21: Ok I understood the purpose.

vooon:
`idx_frame()` is bad because you should return corresponding enum value, not index.
`frame_from_str()` - to be same as `cmode_from_str()`.

TSC21:
> idx_frame() is bad because you should return corresponding enum value, not index.
frame_from_str() - to be same as cmode_from_str().

Solved with https://github.com/TSC21/mavros/commit/7aa2cdc0a303f0b06bba12cbabb98f56331faea7.

mavros_extras/CMakeLists.txt
```diff
@@ -135,10 +136,10 @@ target_link_libraries(gcs_image_bridge
   ${catkin_LIBRARIES}
 )
 
-add_executable(copter_visualization
-  src/copter_visualization.cpp
+add_executable(visualization
```
vooon: Why rename it? It still copter visualizer.

TSC21: But not just copter visualizer. Now it has also landing targets. So it's more general.

mhkabir:
@TSC21 I have some testing time now, so let me know what you would like to be test-flown. Already have copter setup to track aprilgrids.
Firmware integration is what we need at the moment :)

TSC21:
Hi Kabir. For now I think it would be interesting to check the code I added to the visualization file, so to check if the marker is ok. Then we need to focus on firmware integration as you said. Best thing I can think of is check and implement something similar as done with APM firmware.

mhkabir:
@TSC21 I actually wrote a small offboard script to test landing marker landing for now. Using full 3D pose estimate of the marker.

TSC21:
Well that maybe be interesting and probably the recommended but you have to consider that IRLock, for example, can only give position data. Not full pose.

rmackay9:
I'm planning on getting back to the ardupilot implementation (for Copter) in the near-ish future.. within the next month, it's just the AC3.3 release has been a real headache but hopefully it's almost done now.

TSC21:
Thanks for your feedback Randy. Can you give some pointers of the APM code that may be useful to implement a similiar thing on PX4 side?

rmackay9:
@TSC21, This is what's planned for ardupilot Copter:
https://github.com/rmackay9/rmackay9-ardupilot/commits/precland8

It's the last 8 commits or so.  It's actually not the best algorithm because it assumes a "flat earth" and uses a position controller rather than a velocity controller.  Anyway, despite that, this will be the first implementation in ardupilot and we will improve it from there.

TSC21:
> @TSC21, This is what's planned for ardupilot Copter:
https://github.com/rmackay9/rmackay9-ardupilot/commits/precland8

It's the last 8 commits or so. It's actually not the best algorithm because it assumes a "flat earth" and uses a position controller rather than a velocity controller. Anyway, despite that, this will be the first implementation in ardupilot and we will improve it from there.

Thanks. I will check those and see what can be adapted to PX4 use case.

mavros_extras/src/plugins/landing_target.cpp
```diff
@@ -0,0 +1,248 @@
...
+
+		float distance = land_target.distance;
+		float phi = land_target.angle_x;
+		float theta = land_target.angle_y;
+
+		auto position = UAS::transform_frame_ned_enu(Eigen::Vector3d(distance * sin(phi) * sin(theta),	// right now in NED
+					distance * sin(phi) * sin(theta),
+					distance * cos(phi)));
+		// From the data coming from Firmware, it is not possible to determine the orientation, so it's kept as constant:
+		auto orientation = Eigen::Quaterniond(1,0,0,0);
```
vooon: Use `Eigen::Quaterniond::Identity()` here.

TSC21:
Well code updated to latest mavros master and code structure. Will try to check what we can do in the Firmware side. https://github.com/PX4/Firmware/pull/2446 will be a good starting point to continue what I left, put I'll need some inputs so this can be feasible.
@LorenzMeier @mhkabir any thoughts on where to start? For example, should a landing target be identified as the landing target on a Mission?

TSC21:
@mhkabir do you have suggestions here so we can proceed on the FW side?

mhkabir:
Hey, yes I have given this some thought, but I need some time to crystallize it. Will ping you tomorrow with a flow diagram of the architecture on the firmware side.

TSC21:
@mhkabir that is perfect thank you! I'm also syncing with @dagar on this so we can push something to Firmware side that may work with both IRLock or vision-based markers/target detection. Probably picking up on my legacy PR  PX4/Firmware#2446 would be a good start (with proper adaptations to the current structure of course).

TSC21:
@mhkabir another thing I remember you proposing was adding the full pose of the target to `LANDING_TARGET` msg, which I think should be a must mainly for vision-based targets.

TSC21:
> Hey, yes I have given this some thought, but I need some time to crystallize it. Will ping you tomorrow with a flow diagram of the architecture on the firmware side.

@mhkabir do you have the diagram already that you can share? That would help on getting this into the Firmware.

mavros/src/lib/enum_to_string.cpp
```diff
@@ -245,5 +246,55 @@ timesync_mode timesync_mode_from_str(const std::string &mode)
...
+std::string to_string(MAV_FRAME e)
+{
+	size_t idx = enum_value(e);
+	if (idx >= frame_strings.size()) {
+		ROS_ERROR_NAMED("uas", "FRAME: wrong frame index: %zu", idx);
+		return std::to_string(idx);
+	}
+
+	return frame_strings[idx];
+}
```
vooon: Array and that function should be generated from pymavlink!

mavros_extras/mavros_plugins.xml
```diff
@@ -20,7 +20,10 @@
 	<class name="vibration" type="mavros::extra_plugins::VibrationPlugin" base_class_type="mavros::plugin::PluginBase">
 		<description>Publish VIBRATION message data from FCU.</description>
 	</class>
-        <class name="cam_imu_sync" type="mavros::extra_plugins::CamIMUSyncPlugin" base_class_type="mavros::plugin::PluginBase">
+	<class name="landing_target" type="mavros::extra_plugins::LandingTargetPlugin" base_class_type="mavros::plugin::PluginBase">
+		<description>Publish/subscribe to data stream with location of a landing area captured from a downward facing camera.</description>
+	</class>
+  <class name="cam_imu_sync" type="mavros::extra_plugins::CamIMUSyncPlugin" base_class_type="mavros::plugin::PluginBase">
```
vooon: Indent!

mavros_extras/CMakeLists.txt
```diff
@@ -109,7 +110,7 @@ install(PROGRAMS
 )
 
 ## Mark executables and/or libraries for installation
-install(TARGETS mavros_extras copter_visualization servo_state_publisher
+install(TARGETS mavros_extras visualization servo_state_publisher
```
vooon: Do not rename node. No reason. That still models simple quadcopter, not plane or anything else.

TSC21: > anything else.

Not true. It models waypoints also and now it can model landing targets. The model is a quadcopter, but in the end this is a visualization node for several features. more than just a multicopter. It means we can generalize it also to add other models. So I agree with this change.

mavros/src/lib/enum_to_string.cpp
TSC21: the other arrays and functions were also generated from pymavlink? where can I change that so it can also generate them?

TSC21: oh I saw now cog. will redo this then thanks

TSC21:
> Hey, yes I have given this some thought, but I need some time to crystallize it. Will ping you tomorrow with a flow diagram of the architecture on the firmware side.

@mhkabir do you have the diagram already that you can share? That would help on getting this into the Firmware.

TSC21:
@vooon something is very wrong with Travis atm

TSC21:
I think we should review our CI system ASAP

amilcarlucas:
@TSC21 this needs a rebase

TSC21:
@amilcarlucas I know ;) but there more it than rebasing.

TSC21:
I will finish the PR as soon as I have some time. Thanks for the attention on it.

TSC21:
@amilcarlucas rebased. I am going to give some tries with SITL and Gazebo and evaluate if this is working properly with PX4 at least.

TSC21:
@vooon we need to understand why https://travis-ci.org/mavlink/mavros/jobs/487275236#L1581 happens.

TSC21:
> Ardupilot fully supports the LANDING_TARGET mavlink message. So it is probably a better candidate for testing.

So does PX4 ;)

But nothing stops you to test it on Ardupilot. Feel free to do it so.

amilcarlucas:
> So does PX4 ;)

Then it makes sense to remove the "Waits FCU support" label.

amilcarlucas:
It probably makes sense to squash some of the commits

TSC21:
> It probably makes sense to squash some of the commits

Don't see a reason why. I will make sure it works then I will merge it.

TSC21:
I just tested this PR with upstream PX4 and it works as expected. Finally merging this after a long time! :confetti_ball:

PR #312, opened 2015-06-16
Camera IMU synchronisation support added

mhkabir:
@vooon Please review. Don't merge yet. 

Also, I'm unable to figure out how the python mavcmd script argument parsing works, so I couldn't edit it to add the command sending. Can you please add that or tell me how? Or maybe @TSC21 is interested?

PX4 pull is here : https://github.com/PX4/Firmware/pull/2207


mavros/src/plugins/command.cpp
```diff
@@ -316,6 +319,14 @@ class CommandPlugin : public MavRosPlugin {
 				0, 0, 0, 0, 0, 0,
 				res.success, res.result);
 	}
+   
+        bool trigger_control_cb(mavros::CommandTriggerControl::Request &req,
+			mavros::CommandTriggerControl::Response &res) {
+		return send_command_long_and_wait(MAV_CMD_DO_TRIGGER_CONTROL, (req.trigger_enable)? 1.0 : 0.0,
```
vooon: You passed `param1` to `confirmation` argument!

mavros_extras/msg/CamIMUStamp.msg
```diff
@@ -0,0 +1,4 @@
+# IMU-Camera synchronisation data
```
vooon: Maybe also provide IMU data? It stored in UAS, so it might be relatively easy.

vooon:
About python parser, read documentation: https://docs.python.org/2.7/library/argparse.html

Service proxy moved here: [mavros.command](https://github.com/mavlink/mavros/blob/master/mavros/src/mavros/command.py)

mavros_extras/msg/CamIMUStamp.msg
mhkabir: No need here. This message will be parsed by whatever camera driver the user is using, sequence ID matched and the image timestamp set to timestamp from this message :)

mavros/src/plugins/command.cpp
mhkabir: @vooon Whoops. Fixed.

mhkabir:
@vooon Some help needed : 
```
kabir@Kabir-Envy:~$ rosrun mavros mavcmd trigger_control -e -i 300
Request failed. Check mavros logs. ACK: 4
```
What's going on here? I can't find the referred logs to track down what goes wrong.

vooon:
Hmm, usually i don't see ACK from PX4 at all.
But 4 should be MAV_CMD_ACK_ERR_NOT_SUPPORTED.

LorenzMeier:
@mhkabir Does your issue with the rejected command persist? I saw a commit on the Firmware side which should resolve this.

mhkabir:
@LorenzMeier Yes, this persists. I had applied the "commander ignore" commit when testing this. No error / reject message comes in via mavlink log though.

vooon:
@mhkabir status now?

mhkabir:
@vooon No luck. Can you have a look? Latest PX4 master has the trigger app, but mavros still fails.

vooon:
Now i'm going to reassembly my plane, so can't test master right now.

vooon:
For me this PR looks good. Reject seems related to Firmware side, so i merge as is.

PR #315, opened 2015-06-23
Setpoint control plugin added

stevsics:


mhkabir:
@stevsics We already support the generic mavlink messages for this.

TSC21:
@stevsics, as Kabir said, we already have support `SETPOINT_POSITION_LOCAL_NED`. No need to add it for a new msg which, for all I can tell, doesn't seem to make sense. Related: https://github.com/mavlink/mavlink/pull/405

Issue #148, opened 2014-09-08
Conversion for covariances


vooon:
New mavlink messages have covariance information, but i don't know how convert frames for that.

Related #49, #94.

TSC21:
@tonybaltovski do you have some bandwidth to apply https://github.com/mavlink/mavros/pull/208 also here?

mhkabir:
http://robotics.stackexchange.com/questions/2556/how-to-rotate-covariance

TSC21:
#317 adds the solution to this.

PR #317, opened 2015-06-26
Frame conversions

### LINKED ISSUE #194: Generalize frame conversion routines as part of libmavros

TSC21:
Finally, after some tinkering with @blackcoder, we were able to get to the proper conversions to position and orientation between ROS ENU and PX4 NED frames.
So, as this is a common code for many plugins, I decided to create a specific lib file where all the functions for frame conversions are installed. This functions are called by the plugins to do the conversions online.
**ADD:** This PR also updates some comments and doxygen documentation on plugins.

Right now, I think I'm just missing:
- [x] Add covariance frame conversion to `uas_frame_conversions.cpp` (partially)

**Update:** Relative to covariance conversions, I'm still missing:
* Conversion to 7x7 full pose covariance matrix (attitude in quaternions instead of RPY)
* Conversion to quaternion covariance only (based on the previous)
- Note that `transform_frame_covariance_pose6x6()` can also be used to other covariances matrices, (p.e., lin_vel + RPY, ang_vel+RPY, and why not position+lin_vel since the conversion is the same - not very used)

**Update2:** I will leave the above covariances conversions to another PR.

This PR Fix #194, #148 (please continue list).

### DIFF
--- a/mavros/src/lib/uas_frame_conversions.cpp
+++ b/mavros/src/lib/uas_frame_conversions.cpp
@@ -21,49 +21,28 @@ const double PI = boost::math::constants::pi<double>();
 
 using namespace mavros;
 
-static tf::Vector3 convert_position(float _x, float _y, float _z){
+static tf::Vector3 transform_frame_general_xyz(float _x, float _y, float _z){
 	float x =  _x;
 	float y = -_y;
 	float z = -_z;
 	return tf::Vector3(x, y, z);
 };
 
-static tf::Vector3 convert_velocity(float _vx, float _vy, float _vz){
-	float vx =  _vx;
-	float vy = -_vy;
-	float vz = -_vz;
-	return tf::Vector3(vx, vy, vz);
-};
-
-static tf::Vector3 convert_accel(float _ax, float _ay, float _az){
-	float ax =  _ax;
-	float ay = -_ay;
-	float az = -_az;
-	return tf::Vector3(ax, ay, az);
-};
-
-static tf::Vector3 convert_general_xyz(float _x, float _y, float _z){
-	float x =  _x;
-	float y = -_y;
-	float z = -_z;
-	return tf::Vector3(x, y, z);
-};
-
-static tf::Quaternion convert_attitude_q(tf::Quaternion qo){
+static tf::Quaternion transform_frame_attitude_q(tf::Quaternion qo){
 	double roll = PI, pitch = 0.0 , yaw = 0.0f; 
 	tf::Quaternion qr = tf::createQuaternionFromRPY(roll, pitch, yaw);
 	tf::Quaternion qt = qo * qr; 
 	return qt;
 };
 
-static tf::Vector3 convert_attitude_rpy(float _roll, float _pitch, float _yaw){
+static tf::Vector3 transform_frame_attitude_rpy(float _roll, float _pitch, float _yaw){
 	float roll = _roll + PI;
 	float pitch = _pitch;
 	float yaw = _yaw;
 	return tf::Vector3(roll, pitch, yaw);
 };
 
-static std::array<float, 36> convert_covariance_pose6x6(std::array<float, 36> _covariance){
+static std::array<float, 36> transform_frame_covariance_pose6x6(std::array<float, 36> _covariance){
 
 	std::array<float, 36> rotation = {1 ,0,0,0,0,0,
     					  0,-1,0,0,0,0,
@@ -74,6 +53,23 @@ static std::array<float, 36> convert_covariance_pose6x6(std::array<float, 36> _c
 
    	std::array<float, 36> covariance;							
 	std::array<float, 36> temp;		// temporary matrix = R * C
+
+	// The rotation matrix in this case is a diagonal matrix so R = R^T
+	
+	std::transform(rotation.begin()+1, rotation.end(), _covariance.begin()+1, temp.begin(), std::multiplies<float>());
+	std::transform(temp.begin()+1, temp.end(), rotation.begin()+1, covariance.begin(), std::multiplies<float>());
+
+	return covariance;
+};
+
+static std::array<float, 9> transform_frame_covariance_general3x3(std::array<float, 9> _covariance){
+
+	std::array<float, 9> rotation = {1 ,0,0,
+    					 0,-1,0,
+    					 0,0,-1};
+
+   	std::array<float, 9> covariance;							
+	std::array<float, 9> temp;		// temporary matrix = R * C
 
 	// The rotation matrix in this case is a diagonal matrix so R = R^T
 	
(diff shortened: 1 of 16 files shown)


mavros/src/lib/uas_frame_conversions.cpp
```diff
@@ -0,0 +1,64 @@
...
+using namespace mavros;
+
+static tf::Vector3 convert_position(float _x, float _y, float _z){
+	float x =  _x;
+	float y = -_y;
+	float z = -_z;
+	return tf::Vector3(x, y, z);
+};
+
+static tf::Vector3 convert_velocity(float _vx, float _vy, float _vz){
```
tonybaltovski: Why have different functions that perform the same conversion?

TSC21: Just naming types

mhkabir:
I suggest that the naming should be changed to "transform_frame" and no need to make 5x copies of the same function, since we have the exact same conversion for all data from NED to ENU and back.
@TSC21 @vooon @tonybaltovski

TSC21:
> I suggest that the naming should be changed to "transform_frame"

Ok why not

> 5x copies of the same function

This was just a naming convention, as we also have functions to convert attitude (rpy and q). Like this we can name different functions for different physical variables (position, velocity, acceleration, ...). If not, how to call it? I created a `general_xyz` which is being used for any case that doesn't fit into the previous one.

vooon:
I don't see why you really need different function for each type. And if you use same conversion do one general func and call it in inlined pos, att etc. funcs.

Also is i see, there no more difference between body and local?
Conversion is same, but specialized ned-enu, enu-ned suffixes more readable, and describe what we want to do.

TSC21:
Ok i'll launch a general conversion function then. But as you say that there's no need to multiple functions to different types, i don't see why we have to say the direction since the conversion is exactly the same. The direction is marked on comments as you can see.

From flight tests done by @blackcoder, it seems that there's no need to differentiate local from body frames. Mocap and vision plugins already filght tested. Just have to be sure that imu plugin data correspondts to what is expect (which it should since the rotation matrix values correspond to the previous transformations applied to pitch and yaw values). As to ENU to NED conversions used on plugins, it's all good to go, since it was tested and confirmed using a mocap system.

vooon:
Best documentation is the code itself. Commenting each entry worse, than two inlined ned_enu, enu_ned variants.
Anyway compiller optimize to direct call underlying func.

vooon:
And i see, that there difference in general and rpy. Why?

vooon:
Also a note, that those functions are best target for unit tests. Now test exist only for libmavconn, because most plugins require real FCU (at least SITL), write test for them little complex.

TSC21:
> Best documentation is the code itself. Commenting each entry worse, than two inlined ned_enu, enu_ned variants.

I still can't find a use to tell the direction on the function name, as the conversion is exactly the same.

TSC21:
> And i see, that there difference in general and rpy. Why?

No there's no difference. The difference you see is just applying the rotation matrix or not. If we do `roll+PI`, it's the same that applying a rotation matrix where we do `-pitch` and `-yaw`. This is just two separate ways of seeing the same thing, even though it's easier to understand the applied rotation when we add 180º to Roll. So the rotation is exactly the same for both position and rotation. Check the covariance transform functions and you'll see what I mean.

TSC21:
> Also a note, that those functions are best target for unit tests. Now test exist only for libmavconn, because most plugins require real FCU (at least SITL), write test for them little complex.

I'm currently out of bandwidth to write unit tests on these. As I said, ROS to FCU conversion is flight tested. As FCU to ROS has the same conversion applied, we can consider it's also tested cause it was previously working (but now is implemented on a specific function on the library).

TSC21:
Does anyone as a suggestion in how to apply the frame transformation to a 7x7 full pose matrix, i.e. position+quaternion covariance matrix? I've been trying to understand how to apply the rotation in the case of the quaternion but it's not trivial. Thanks!

vooon:
Where did you find 7x7 covariance? [geometry_msgs/PoseWithCovariance](http://docs.ros.org/jade/api/geometry_msgs/html/msg/PoseWithCovariance.html) uses 6x6.

TSC21:
> Where did you find 7x7 covariance? [geometry_msgs/PoseWithCovariance](http://docs.ros.org/jade/api/geometry_msgs/html/msg/PoseWithCovariance.html) uses 6x6.

Well 7x7 pose matrices exists and can be used. Check MRPT: http://www.mrpt.org/tutorials/programming/maths-and-geometry/2d_3d_geometry/#4_Conversions_between_classes.
Quaternion covariance also exists. Check page 7 of [this paper](http://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/20030093641.pdf).
So I assume a 7x7 covariance matrix can be built, even though ROS doesn't support it yet, but many applications may depend on a covariance matrix like this. The problem is that the propagation of covariance in quaternion representation is no so trivial.

mavros/src/plugins/imu_pub.cpp
```diff
@@ -207,13 +206,12 @@ class IMUPubPlugin : public MavRosPlugin {
 
 		auto imu_msg = boost::make_shared<sensor_msgs::Imu>();
 
-		// PX4 NED (w x y z) -> ROS ENU (x -y -z w) (body-fixed)
-		tf::Quaternion orientation(att_q.q2, -att_q.q3, -att_q.q4, att_q.q1);
+		tf::Quaternion orientation(att_q.q1, att_q.q2, att_q.q3, att_q.q4);
```
vooon: NO! You missed that FCU and ROS use different field convention.
It still should be `orientation(q2, q3, q4, q1)`.

TSC21: FCU: w x y z
ROS: x y z w
right?

mavros/src/plugins/imu_pub.cpp
```diff
@@ -251,12 +252,13 @@ class IMUPubPlugin : public MavRosPlugin {
 		if (imu_hr.fields_updated & (7 << 6)) {
 			auto magn_msg = boost::make_shared<sensor_msgs::MagneticField>();
 
-			// Convert from local NED plane to ENU
-			magn_msg->magnetic_field.x = imu_hr.ymag * GAUSS_TO_TESLA;
-			magn_msg->magnetic_field.y = imu_hr.xmag * GAUSS_TO_TESLA;
-			magn_msg->magnetic_field.z = -imu_hr.zmag * GAUSS_TO_TESLA;
+			auto mag_field = UAS::transform_frame_ned_enu_xyz(imu_hr.xmag, imu_hr.ymag, imu_hr.zmag);
```
vooon: There bee GAUSS units, not milliTESLA.

mavros/src/plugins/imu_pub.cpp
vooon: Yes.

mavros/src/plugins/imu_pub.cpp
TSC21: Got it

Issue #71, opened 2014-07-30
Replace TF with TF2


vooon:
Most of plugins use 'deprecated' tf package, while Hydro recommend tf2.

TSC21:
@vooon want to try to get this up and running?

vooon:
Maybe, but it also require using eigen.
18.05.2015 23:26 пользователь "TSC21" <notifications@github.com> написал:

dreamkeep:
can some one give a sample to takeoff use ./mavcmd  ?

vooon:
@dreamkeep wrong issue. Also https://github.com/mavlink/mavros#support-forums-and-chats

```
rosrun mavros mavsafety arm
rosrun mavros mavcmd takeoffcur 0 0 20
```

vooon:
Hmm, i'm readed the tf2_ros source code and thinking to move tf to UAS, so use single buffer instead of multiple ones.

PR #337, opened 2015-07-10
SITL tests architecture

TSC21:
This is **WIP**, but already usable.

* Defined test structure: 1. main node `sitl_test_node`, 2. main testing class `TestType`, 3. test classes (as `OffboardPosCtrlSquare`);
* First working test routine `offboard_posctl_square`; Usage: `roslaunch test_mavros iris_empty_world_offboard_posctl_square.launch` 

**NOTE:** Current code only applied to PX4 SITL; not tested with APM SITL.

**UPDATE:** added a launch file which integrates all the needed nodes (for OFFB POSCTL square shaped path)

**UPDATE2:** Testing other test routines: Change `shape` parameter on launch file.

**UPDATE3:** Added velocity setpoint control.

**UPDATE4:** Use Eigen for math. Generalize launch file for Offboard Control.


test_mavros/CMakeLists.txt
```diff
@@ -7,11 +7,18 @@ project(test_mavros)
...
+  std_msgs
+  geometry_msgs
 )
 
 ## System dependencies are found with CMake's conventions
 # find_package(Boost REQUIRED COMPONENTS system)
+find_package(Boost REQUIRED COMPONENTS system)
 
+include(EnableCXX11)
+include(MavrosMavlink)
```
vooon: Why mavlink? You really need encoder and decoder functions?

test_mavros/CMakeLists.txt
```diff
@@ -7,11 +7,18 @@ project(test_mavros)
 find_package(catkin REQUIRED COMPONENTS
   mavros
   mavros_extras
+  roscpp
+  tf
```
vooon: Please use tf2_ros. Today i'm done with #71, don't want tf1 code.

test_mavros/CMakeLists.txt
```diff
@@ -79,10 +86,10 @@ find_package(catkin REQUIRED COMPONENTS
 ## CATKIN_DEPENDS: catkin_packages dependent projects also need
 ## DEPENDS: system dependencies of this project that dependent projects also need
 catkin_package(
-#  INCLUDE_DIRS include
-#  LIBRARIES test_mavros
-#  CATKIN_DEPENDS mavros mavros_extras
-#  DEPENDS system_lib
+  INCLUDE_DIRS include
```
vooon: For test usually there nothing. It is not normal package, so no reason to export depends.

test_mavros/CMakeLists.txt
```diff
@@ -93,25 +100,44 @@ catkin_package(
...
+  include/tests/offboard_posctl_square.h
+  include/sitl_test/sitl_test.h
+  include/sitl_test/test_type.h
+)
 
 ## Declare a cpp executable
-# add_executable(test_mavros_node src/test_mavros_node.cpp)
+add_executable(
+  sitl_test_node
+  src/sitl_test/px4/sitl_test_node.cpp
```
vooon: Why `src`? Usually that code located in `test`.

vooon: Example: https://github.com/ros/geometry_experimental/tree/jade-devel/test_tf2

test_mavros/CMakeLists.txt
TSC21: Well thinking of it, probably not.

test_mavros/CMakeLists.txt
TSC21: Ok sure.

test_mavros/CMakeLists.txt
TSC21: Well as it is implemented, we probably need this as is.

test_mavros/CMakeLists.txt
TSC21: Well this was just to separate src code from the rest. But it's just a folder, so we can re-adapt no problem.

test_mavros/src/sitl_test/px4/sitl_test_node.cpp
```diff
@@ -0,0 +1,28 @@
...
+#include <mavros/mavros.h>
+#include <sitl_test/sitl_test.h>
+
+using namespace sitltest;
+using namespace mavros;
+
+int main(int argc, char *argv[])
+{
+	ROS_INFO("SITL Test node started...");
+	ROS_INFO(argv[1]);
```
TSC21: This line was just a debug one; will remove it asap.

test_mavros/include/tests/offboard_control.h
```diff
@@ -0,0 +1,471 @@
...
+        return sp;
+    }
+
+    /**
+     * @brief Square path motion routine
+     */
+    void square_path_motion(ros::Rate loop_rate){
+        uint8_t pos_target = 0;
+
+        local_pos_sp_pub = nh_sp.advertise<geometry_msgs::PoseStamped>("/mavros/setpoint_position/local", 10);
```
TSC21: I have to find a way of initializing this once without Travis giving any kind of error. @vooon any suggestion?

test_mavros/include/tests/offboard_control.h
```diff
@@ -0,0 +1,470 @@
...
+		return sp;
+	}
+
+	/**
+	 * @brief Square path motion routine
+	 */
+	void square_path_motion(ros::Rate loop_rate){
+		uint8_t pos_target = 0;
+
+		local_pos_sp_pub = nh_sp.advertise<geometry_msgs::PoseStamped>("/mavros/setpoint_position/local", 10);
```
TSC21: I have to find a way of initializing this once without Travis giving any kind of error. @vooon any suggestion?

test_mavros/include/tests/offboard_control.h
```diff
@@ -0,0 +1,470 @@
...
+		 * - eight
+		 * - ellipse (in 3D space)
+		 */
+		nh_sp.param<std::string>("shape", shape, "square");
+
+		ros::Rate loop_rate(10);
+
+		ROS_INFO("SITL Test: Offboard control test running!");
+
+		if (mode.compare("position") == 0) {
```
TSC21: Probably I can use an OR ( `x || y` ) here...

LorenzMeier:
@AndreasAntener do you have feedback on the integration here?

test_mavros/include/sitl_test/sitl_test.h
```diff
@@ -0,0 +1,37 @@
+/**
+ * @brief SitlTest node implementation class
+ * @file sitl_test.h
+ * @author Nuno Marques <n.marques21@hotmail.com>
+ *
+ * @addtogroup nodelib
```
vooon: Change group, let call this "tests".

test_mavros/include/tests/offboard_control.h
vooon: You can do `mode == "position"`. Also there nothing to do if mode incorrect,
so do first bunch of checks about mode, then shape.

test_mavros/include/tests/offboard_control.h
```diff
@@ -0,0 +1,470 @@
...
+	geometry_msgs::TwistStamped vs;
+
+	/* -*- alias -*- */
+    #define X  ps.pose.position.x
+    #define Y  ps.pose.position.y
+    #define Z  ps.pose.position.z
+    #define VX  vs.twist.linear.x
+    #define VY  vs.twist.linear.y
+    #define VZ  vs.twist.linear.z
+    #define sp_X  pos_setpoint().x
```
vooon: That don't work. Each call to pos_setpoint() return different Point object.

vooon: Also maybe better write explicit `ps.pose.position.x` rather that this preprocessor magic.

vooon: Or you can use Eigen for math, and then convert it to msg (`eigen_conversions.h`).

test_mavros/include/sitl_test/sitl_test.h
TSC21: OK

test_mavros/include/tests/offboard_control.h
```diff
@@ -0,0 +1,470 @@
...
+    #define current_x localpos.pose.position.x
+    #define current_y localpos.pose.position.y
+    #define current_z localpos.pose.position.z
+
+	/* -*- helper functions -*- */
+
+	/**
+	 * @brief Defines single position setpoint
+	 */
+	geometry_msgs::Point pos_setpoint(){
```
vooon: One thing that uncrustify don't do is space between `()` and `{`.

test_mavros/include/tests/offboard_control.h
TSC21: OK sure

test_mavros/include/tests/offboard_control.h
vooon: What error? Don't see any.

test_mavros/include/tests/offboard_control.h
TSC21: Well if you check out the code bellow you will see 2 things:
1. This works, and yes I want a different Point to the object. If it didn't work I would check it already in the simulation.
2. I use this as an alias with a purpose: reduce the size of the code lines and make them more explicit. If I didn't use them, you can see that there are some code lines that become a mess.

test_mavros/include/tests/offboard_control.h
```diff
@@ -0,0 +1,470 @@
...
+		uint8_t pos_target = 0;
+
+		local_pos_sp_pub = nh_sp.advertise<geometry_msgs::PoseStamped>("/mavros/setpoint_position/local", 10);
+		vel_sp_pub = nh_sp.advertise<geometry_msgs::TwistStamped>("/mavros/setpoint_velocity/cmd_vel", 10);
+		local_pos_sub = nh_sp.subscribe("/mavros/local_position/local", 10, &OffboardControl::local_pos_cb, this);
+
+		while (ros::ok()) {
+			ROS_INFO("Testing...");
+
+			if (mode.compare("position") == 0) {
```
vooon: Bad idea, better define an enum and select mode at param checking.

test_mavros/include/tests/offboard_control.h
TSC21: Right now there's none cause I'm initializing this inside the functions. But I have to do it in each of them. I I initialize outside the functions, it will give errors in Travis.

test_mavros/include/tests/offboard_control.h
TSC21: Hmm ok. But I've run uncrustify and this was the resulting code.

test_mavros/include/tests/offboard_control.h
```diff
@@ -0,0 +1,470 @@
...
+	 */
+	void square_path_motion(ros::Rate loop_rate){
+		uint8_t pos_target = 0;
+
+		local_pos_sp_pub = nh_sp.advertise<geometry_msgs::PoseStamped>("/mavros/setpoint_position/local", 10);
+		vel_sp_pub = nh_sp.advertise<geometry_msgs::TwistStamped>("/mavros/setpoint_velocity/cmd_vel", 10);
+		local_pos_sub = nh_sp.subscribe("/mavros/local_position/local", 10, &OffboardControl::local_pos_cb, this);
+
+		while (ros::ok()) {
+			ROS_INFO("Testing...");
```
vooon: On each iteration? Also add test type to this print.

test_mavros/include/tests/offboard_control.h
TSC21: Well for that we have to write a function for each mode, which I think it's also a bad idea.

test_mavros/include/tests/offboard_control.h
```diff
@@ -0,0 +1,470 @@
...
+			}
+
+			wait_destination(ps);
+
+			// motion routine
+			for (int theta = -180; theta <= 180; theta++) {
+				if (mode.compare("position") == 0)
+					local_pos_sp_pub.publish(eight_shape(theta));
+				else if (mode.compare("velocity") == 0) {
+					VX = eight_shape(theta).pose.position.x - current_x;
```
vooon: Each call you do calculations, why not store result in temporary variable?

test_mavros/include/tests/offboard_control.h
TSC21: > On each iteration?

Well in case of position yes.

> Also add test type to this print.

Have you already tested the behavior of the code? :) https://github.com/mavlink/mavros/pull/337/files#diff-3e2be3af4f08d8676c481f5b7228ca3dR70

test_mavros/include/tests/offboard_control.h
```diff
@@ -0,0 +1,470 @@
...
+ */
+/*
+ * Copyright 2015 Nuno Marques, Andre Nguyen.
+ *
+ * This file is part of the mavros package and subject to the license terms
+ * in the top-level LICENSE file of the mavros repository.
+ * https://github.com/mavlink/mavros/tree/master/LICENSE.md
+ */
+
+#include <mavros/mavros.h>
```
vooon: Why? Test shouldn't depend on internal mavros data.

test_mavros/include/tests/offboard_control.h
TSC21: But it has to do the calculations at each iteration. It's dynamic. Why would it store the result in a temporary variable if it's not going to use that value in the next iteration?

test_mavros/include/tests/offboard_control.h
```diff
@@ -0,0 +1,470 @@
...
+		while (ros::ok() && !stop) {
+			double distance = sqrt(
+					(dest.x - current_x) * (dest.x - current_x) +
+					(dest.y - current_y) * (dest.y - current_y) +
+					(dest.z - current_z) * (dest.z - current_z));
+
+			if (distance <= 0.1f)	/** @todo Add gaussian threshold */
+				stop = true;
+
+			if (mode.compare("position") == 0) {
```
vooon: Why you do publish in wait? Possibly bad naming.

test_mavros/include/tests/offboard_control.h
TSC21: Well this one doesn't so I can remove. For other test that regard on timestamps, etc., probably they will.

test_mavros/sitl_test/launch/base_node.launch
```diff
@@ -0,0 +1,13 @@
...
+	<!-- vim: set ft=xml noet : -->
+	<!-- Base node launch file-->
+	<arg name="args" />
+	<arg name="mode" />
+	<arg name="shape" />
+
+	<node pkg="test_mavros" type="sitl_test_node" args="$(arg args)" name="sitl_test_offboard_ctl" required="true" clear_params="true" output="screen">
+		<param name="mode" value="$(arg mode)" />
+		<param name="shape" value="$(arg shape)" />
+    </node>
```
vooon: Tab? not 4 spaces.

test_mavros/sitl_test/lib/sitl_test.cpp
```diff
@@ -0,0 +1,35 @@
...
+using namespace testtype;
+
+SitlTest::SitlTest()
+{}
+
+void SitlTest::spin(int argc, char *argv[])
+{
+	if (strcmp(argv[1],"offboard_control") == 0)
+	{
+		ros::init(argc, argv, "offboard_control");
```
vooon: `ros::init()` usually is first line of `main()` body.

test_mavros/include/tests/offboard_control.h
TSC21: Well it waits for reaching the setpoint before moving to the next one. Do you have suggestions to other names?

test_mavros/sitl_test/lib/sitl_test.cpp
vooon: Also i don't understand why you ever need that class.

test_mavros/sitl_test/launch/base_node.launch
TSC21: Ok have to review that.

test_mavros/sitl_test/lib/sitl_test.cpp
TSC21: https://github.com/mavlink/mavros/pull/337/files#diff-52524e75eeb8feac5976b23f5f4be2c5R24. It's not possible in this case because the node init depends on the arguments.

test_mavros/include/tests/offboard_control.h
vooon: You can't do initialization of non const outside of method. You may do that in constructor.

test_mavros/sitl_test/lib/sitl_test.cpp
TSC21: > Also i don't understand why you ever need that class.

It's a class to receive arguments and launch the nodes. And to be separate from the node file.

test_mavros/include/tests/offboard_control.h
TSC21: Hmm ok. I'll see what I can do.

test_mavros/include/tests/offboard_control.h
vooon:
```
typedef enum {
 POSITION,
 VELOCITY,
 ACCELERATION
} operation_mode;

class A
{
operation_mode mode;

init() {
std::string mode_;
nh.param("mode", mode_, "position");
if (mode_ == "position")
 mode = POSITION;
... 
}

spin() {
 if (mode == POSITION)
   do_a();
 else if (mode == VELOCITY)
   do_b();
...
}
}
```

test_mavros/include/tests/offboard_control.h
vooon: I'm about:
```
eight_shape(theta);
eight_shape(theta);
eight_shape(theta);
```

test_mavros/include/tests/offboard_control.h
TSC21: Ok I understand what you mean now. And the `init()` function seems a good idea, cause then I can also use it init the Publishers and the Subscribers. Thanks for the tip.

test_mavros/include/tests/offboard_control.h
TSC21: Oh ok, that's more concrete. So yes, store `eight_shape(theta);` at the beginning of the iteration and then use it in each calculation. Sounds good ;)

test_mavros/include/tests/offboard_control.h
vooon: For ros types directly include `<ros/ros.h>`.

TSC21: Yes makes sense. I'll review the header files and see what I add in each one. Mostly I have repeated .h files which are just needed once.

test_mavros/include/tests/offboard_control.h
vooon: Instead of init you can do this in constructor. So init is not necessary.

TSC21: Ok

test_mavros/sitl_test/lib/sitl_test.cpp
vooon: Why? Do simple node for each test group. Combining all in one is not good.

TSC21: I don't think it makes any difference because it's just a node at a time that is launched. And that node is launched based on the args given on `rosrun`/`roslaunch`, so it's a shorter and efficient implementation and didn't give me any problems till now.

test_mavros/include/tests/offboard_control.h
```diff
@@ -0,0 +1,508 @@
...
+			}
+			else if (shape == EIGHT) {
+				ROS_INFO("Test option: eight-shaped path...");
+				eight_path_motion(loop_rate, mode);
+			}
+			else if (shape == ELLIPSE) {
+				ROS_INFO("Test option: ellipse-shaped path...");
+				ellipse_path_motion(loop_rate, mode);
+			}
+		}
```
vooon: You have identical code in if(mode), why not move it out?

test_mavros/include/tests/offboard_control.h
```diff
@@ -0,0 +1,508 @@
...
+
+		else if (mode == ACCELERATION) {
+			ROS_INFO("Aceleration control mode selected.");
+			/**
```
vooon: Place error here, just to be sure why node fail.

test_mavros/include/tests/offboard_control.h
TSC21: I tried that once but it was failing and I thought it was because I was doing it. Looks like it wasn't because of it, cause I applied your suggestion now and worked.

test_mavros/include/tests/offboard_control.h
TSC21: OK

test_mavros/include/sitl_test/test_type.h
```diff
@@ -0,0 +1,35 @@
...
+ * in the top-level LICENSE file of the mavros repository.
+ * https://github.com/mavlink/mavros/tree/master/LICENSE.md
+ */
+
+#pragma once
+
+#include <tests/offboard_control.h>
+
+namespace testtype {
+class TestType {
```
vooon: Don't see any usage of that class.

test_mavros/include/sitl_test/test_type.h
```diff
@@ -0,0 +1,35 @@
+/**
+ * @brief SITL test type
+ * @file test_type.h
+ * @author Nuno Marques <n.marques21@hotmail.com>
+ *
+ * @addtogroup sitl_test
```
vooon: Also we should use one group, sitl_test or tests.

test_mavros/include/tests/offboard_control.h
```diff
@@ -0,0 +1,492 @@
...
+				ros::spinOnce();
+			}
+		}
+	}
+
+	/**
+	 * @brief Defines the accepted threshold to the destination/target position
+	 * before moving to the next setpoint.
+	 */
+	void wait_and_move(geometry_msgs::PoseStamped target){
```
vooon: For me it looks like move & wait in that order. But ok.

test_mavros/sitl_test/launch/px4/iris_empty_world_offboard_posctl_square.launch
```diff
@@ -0,0 +1,59 @@
...
+        <arg name="ns" value="$(arg ns)"/>
+    </include>
+
+    <!-- MAVROS launcher -->
+    <arg name="fcu_url" default="udp://localhost:14560@localhost:14565" />
+    <arg name="gcs_url" default="" />
+    <arg name="tgt_system" default="1" />
+    <arg name="tgt_component" default="50" />
+
+    <include file="$(find mavros)/launch/node.launch">
```
vooon: Include px4.launch instead. All that code needed only if you want to pass different config file.

test_mavros/sitl_test/launch/px4/iris_empty_world_offboard_posctl_square.launch
```diff
@@ -0,0 +1,59 @@
...
+    
+    <include file="$(find px4)/launch/iris.launch">
+        <arg name="ns" value="$(arg ns)"/>
+    </include>
+
+    <!-- MAVROS launcher -->
+    <arg name="fcu_url" default="udp://localhost:14560@localhost:14565" />
+    <arg name="gcs_url" default="" />
+    <arg name="tgt_system" default="1" />
+    <arg name="tgt_component" default="50" />
```
vooon: Now that 1. See #340.

test_mavros/sitl_test/launch/px4/iris_empty_world_offboard_posctl_square.launch
```diff
@@ -0,0 +1,59 @@
...
+        <arg name="headless" value="$(arg headless)"/>
+        <arg name="gui" value="$(arg gui)"/>
+        <arg name="enable_logging" value="$(arg enable_logging)" />
+        <arg name="enable_ground_truth" value="$(arg enable_ground_truth)" />
+        <arg name="log_file" value="$(arg log_file)"/>
+    </include>
+    
+    <include file="$(find px4)/launch/iris.launch">
+        <arg name="ns" value="$(arg ns)"/>
+    </include>
```
vooon: One bad thing that we can't change is absence rotors_simulator for jade. And PX4 too, so we can't add it to package.xml.
Only may try as `<run_depend>`.

test_mavros/include/sitl_test/test_type.h
TSC21: https://github.com/mavlink/mavros/pull/337/files#diff-a972179b5f960cd4f882fe71f48d887cR29. Haven't put one, but test cycle rates, number of test loops, setpoint precision before moving to the next one, etc.

test_mavros/include/sitl_test/test_type.h
TSC21: Well I think I will stick to `sitl_test` as it makes more sense. I will do the same on the rest.

test_mavros/include/tests/offboard_control.h
TSC21: Well it waits (https://github.com/mavlink/mavros/pull/337/files#diff-3e2be3af4f08d8676c481f5b7228ca3dR464) then it moves (https://github.com/mavlink/mavros/pull/337/files#diff-3e2be3af4f08d8676c481f5b7228ca3dR467)

test_mavros/sitl_test/launch/px4/iris_empty_world_offboard_posctl_square.launch
TSC21: Yep seems so. What do you suggest to add to `package.xml` then? (code sample)

test_mavros/sitl_test/launch/px4/iris_empty_world_offboard_posctl_square.launch
TSC21: Yep thanks!

test_mavros/sitl_test/launch/px4/iris_empty_world_offboard_posctl_square.launch
TSC21: But we are passing different config (`fcu_url` for example).

test_mavros/include/tests/offboard_control.h
andre-nguyen: It's because if you stop publishing, offboard control will time out. You have to keep publishing while you wait to get to your destination.

TSC21:
Regarding the velocity control: As to avoid extreme overshoots on the FCU internal controller, I'll be using `control_toolbox::Pid` to setup a proper velocity controller in MAVROS side. I'll probably implement it as a `lib` file in `mavros`, and add to `utils.h` on `include` folder. @vooon what do you think?

TSC21:
Next PR will have instructions on how to use this. Will add them to README.md.

AndreasAntener:
Sorry I didn't have time yet to provide any feedback on this regarding integration.

Is the idea of this to provide support classes for others writing integration tests? Or should this be more or less standalone as a default set of tests for any MAVROS SITL? (or both?)

I'd be glad to start using this for PX4 SITL tests, as we still miss a velocity control test.

TSC21:
Hi @AndreasAntener! Thanks, your feedback is very welcomed!
Well it turns out to be both,  eventhough the main idea is to use this as a standard test architecture for MAVROS<->SITL connectivity and also test modifications on both MAVROS and Firmware side. But this can also be used as standard classes which people can use to implement other kind of tests (which I also thank so this is not a one-man's job).
> I'd be glad to start using this for PX4 SITL tests, as we still miss a velocity control test.

`offboard_control.h` already implements tests for velocity control and you can check they work... but they don't work so well cause we lack a PID controller in MAVROS side to smooth the velocity fed into the controller side of the Firmware. I'm on my way right now to implement a controller class in `test_mavros` right now with an extension to `control_toolbox::Pid`.

AndreasAntener:
Velocity, acceleration and attitude control test are hard because they require a certain behaviour from the simulator you test against. Integrating a PID controller for velocity control sounds interesting, looking forward to see that in action.

I see you haven't added attitude control yet. Any particular reason or just didn't get to it yet? :)

TSC21:
> I see you haven't added attitude control yet. Any particular reason or just didn't get to it yet? :)

Well good point: in the case of attitude control, the code that has to be implemented has to be different from the code on the the other three, to which I can just define paths I want the vehicle to follow. In case of attitude I think the test bench must include RP tilting and Yaw testing. So probably I'll split offboard control into to different headers: one for direct RPY control and one for control of position, velocity and accel.

TSC21:
Also Andreas, if you have some ideas, I ask you to please leave the suggestions on https://github.com/mavlink/mavros/issues/338 so we can discuss them there :)

AndreasAntener:
Sure, no problem :)

PR #342, opened 2015-07-16
Orientation enum name

TSC21:
* Added names to orientation list;
* Added functions: one to retrieve the name when having the enum, other to retrieve the number when having the name.


mavros/include/mavros/mavros_uas.h
```diff
@@ -330,11 +330,21 @@ class UAS {
...
 
+	/**
+	 * @brief Retrieve alias of the orientation received by MAVLink msg.
+	 */
+	static std::string str_sensor_orientation(MAV_SENSOR_ORIENTATION orientation);
+
+	/**
+	 * @brief Retrieve sensor orientation number from alias name.
+	 */
+	static int orientation_from_str(std::string sensor_orientation);
```
vooon: Change `std::string sensor_orientation` to `const std::string &sensor_orientation`.

Also this function should be capable convert numeric value of enum (e.g. strtol() then check & return).
That needed for backward compatibility.

mavros/launch/px4_config.yaml
```diff
@@ -114,25 +114,67 @@ distance_sensor:
   hrlv_ez4_pub:
     id: 0
     frame_id: "hrlv_ez4_sonar"
-    orientation: 8 # RPY:{180.0, 0.0, 0.0}
+    orientation: "ROLL_180" # RPY:{180.0, 0.0, 0.0}
```
vooon: `"` not necessary, and perhaps better to delete.
Example:
```yaml
pub:
  orientation: ROLL_180
```

mavros/launch/px4_config.yaml
```diff
@@ -114,25 +114,67 @@ distance_sensor:
...
+#  "ROLL_270_PITCH_90"
+#  "ROLL_90_PITCH_180"
+#  "ROLL_270_PITCH_180"
+#  "ROLL_90_PITCH_270"
+#  "ROLL_180_PITCH_270"
+#  "ROLL_270_PITCH_270"
+#  "ROLL_90_PITCH_180_YAW_90"
+#  "ROLL_90_YAW_270"
+#  "ROLL_315_PITCH_315_YAW_315"
+##
```
vooon: Tooooo long. That should be in ros wiki (and link to that doc).

mavros/src/lib/uas_sensor_orientation.cpp
```diff
@@ -26,66 +25,87 @@ typedef std::pair<const std::string, const Eigen::Quaterniond> OrientationPair;
...
+/* 10 */ make_orientation("ROLL_180_YAW_90", 			180.0,   0.0,  90.0),
+/* 11 */ make_orientation("ROLL_180_YAW_135", 			180.0,   0.0, 135.0),
+/* 12 */ make_orientation("PITCH_180", 				  0.0, 180.0,   0.0),
+/* 13 */ make_orientation("ROLL_180_YAW_225", 			180.0,   0.0, 225.0),
+/* 14 */ make_orientation("ROLL_180_YAW_270", 			180.0,   0.0, 270.0),
+/* 15 */ make_orientation("ROLL_180_YAW_315", 			180.0,   0.0, 315.0),
+/* 16 */ make_orientation("ROLL_90", 			 	 90.0,   0.0,   0.0),
+/* 17 */ make_orientation("ROLL_90_YAW_45", 			 90.0,   0.0,  45.0),
+/* 18 */ make_orientation("ROLL_90_YAW_90", 			 90.0,   0.0,  90.0),
+/* 19 */ make_orientation("ROLL_90_YAW_135", 			 90.0,   0.0, 135.0),
```
vooon: Here better to use spaces, else formatting will depend on editor.
Same exception about matrix init code.

mavros/src/lib/uas_sensor_orientation.cpp
```diff
@@ -26,66 +25,87 @@ typedef std::pair<const std::string, const Eigen::Quaterniond> OrientationPair;
...
 		return Eigen::Quaterniond::Identity();
 	}
 
 	return sensor_orientations[idx].second;
 }
+
+int UAS::orientation_from_str(std::string sensor_orientation)
+{
+	int index, idx = 0;
+	for (const auto &orientation : sensor_orientations) {
```
vooon: Here you may use regular for, not foreach.
Also not this function do not work, almost always return -1 because it should look that (pseudocode):
```cpp
for (size_t idx = 0; idx < so.size(); idx++) {
 if (so[idx].first == key)
  return idx;
}
// fallback
v = strtol(key); // this should not fail, so use try {} catch

// not found at all
return -1;
```

vooon: https://github.com/mavlink/mavros/blob/master/mavros/src/lib/uas_stringify.cpp#L185

mavros_extras/src/plugins/distance_sensor.cpp
```diff
@@ -191,10 +191,12 @@ class DistanceSensorPlugin : public MavRosPlugin {
 			return;
 		}
 
-		if (sensor->orientation >= 0 && dist_sen.orientation != sensor->orientation) {
+		if (UAS::orientation_from_str(sensor->orientation) >= 0 && dist_sen.orientation != UAS::orientation_from_str(sensor->orientation)) {
```
vooon: Not again! Please don't do expensive calls where you may avoid that.
String comparison costs much more than int. Also that broke behavior with undefined orientation.

TSC21: ...
This still is a int comparison... The only thing I can do is store the value previously before do the comparison... Didn't broke nothing

vooon: And `UAS::orientation_from_str()` does not do string comparison?

TSC21: Well yes. But sensor->orientation is supposed to be a string, that's the purpose. What do you suggest?

TSC21: Ok I know what to do. Will let you know soon

mavros/src/lib/uas_sensor_orientation.cpp
TSC21: @vooon why do I need
```c++
// fallback
v = strtol(key); // this should not fail, so use try {} catch
```

?

TSC21:
```c++
for (size_t idx = 0; idx < so.size(); idx++) {
 if (so[idx].first == key)
  return idx;
}
```

if `key` is a string, the for loop does a comparison and returns the idx only if that string name exists in the array. If the loop finishes without returning the `idx`, then it
```c++
return -1;
```

So I don't see why we have to `try {} catch` a string to long conversion... The function or returns an index or -1...

TSC21: https://github.com/mavlink/mavros/blob/master/mavros/src/lib/uas_stringify.cpp#L185 only makes sense because you are using a `map`, which requires you to convert `mode.first` to `int`. In this case, I can even use `int` instead of `size_t` and I'll have my orientation number directly.

mavros/include/mavros/mavros_uas.h
TSC21: > Change `std::string sensor_orientation` to `const std::string &sensor_orientation`.

Why? And why not also `Eigen::Quaterniond sensor_orientation_matching`?

> Also this function should be capable convert numeric value of enum (e.g. strtol() then check & return).

Convert to what? `MAV_SENSOR_ORIENTATION orientation` is supposed to be a number, which is used as an index to return: orientation name in case of `UAS::str_sensor_orientation()`; and orientation values in case of `UAS::sensor_orientation_matching()`. And `orientation_from_str()` was created so we can return the orientation number/index if we give the orientation name. So I think we have all the functionalities we need.

mavros/launch/px4_config.yaml
TSC21: So add it :)

vooon: Why me? You also can edit. **WIKI**

mavros/include/mavros/mavros_uas.h
vooon: > Why?
Because this reduces object copying. What you want change in UAS::sensor_orientation_matching()?

> strtol()

Because there may be lots of configs with number representation. Why not just simply try to fall back to old behavior?

mavros_extras/src/plugins/distance_sensor.cpp
```diff
@@ -37,7 +37,7 @@ class DistanceSensorItem {
 		send_tf(false),
 		sensor_id(0),
 		field_of_view(0),
-		orientation(-1),
+		orientation("LOCAL_NED"),
```
vooon: This make it worse. Nope-nope-nope!
Leave `orientation` as integer.

mavros_extras/src/plugins/distance_sensor.cpp
```diff
@@ -191,10 +191,13 @@ class DistanceSensorPlugin : public MavRosPlugin {
 			return;
 		}
 
-		if (sensor->orientation >= 0 && dist_sen.orientation != sensor->orientation) {
```
vooon: Leave this if as is. Wrong place.

mavros_extras/src/plugins/distance_sensor.cpp
```diff
@@ -246,8 +249,11 @@ void DistanceSensorItem::range_cb(const sensor_msgs::Range::ConstPtr &msg)
 	uint8_t type = 0;
 	uint8_t covariance_ = 0;
 
+	int orient_value = UAS::orientation_from_str(orientation);
```
vooon: Instead of one call at object creation you repeat that many times.

mavros_extras/src/plugins/distance_sensor.cpp
```diff
@@ -263,13 +269,14 @@ void DistanceSensorItem::range_cb(const sensor_msgs::Range::ConstPtr &msg)
...
 			sensor_id,
-			orientation,
+			orient_value,
 			covariance_);
 }
 
 DistanceSensorItem::Ptr DistanceSensorItem::create_item(DistanceSensorPlugin *owner, std::string topic_name)
 {
 	auto p = boost::make_shared<DistanceSensorItem>();
+	auto orient_value = UAS::orientation_from_str(p->orientation);
```
vooon: Great! Use uninit variable (with obviously bad value LOCAL_NED).

mavros_extras/src/plugins/distance_sensor.cpp
```diff
@@ -302,7 +309,7 @@ DistanceSensorItem::Ptr DistanceSensorItem::create_item(DistanceSensorPlugin *ow
 		}
 
 		// orientation check
-		pnh.param("orientation", p->orientation, -1);
+		pnh.param("orientation", orient_value, -1);
```
vooon: And instead of string you pulled int, which unused...

vooon:
Only one thing need to change in distance_sensor.cpp: `DistanceSensorItem::create_item()`

You should pull `orientation` parameter to temporary std::string, then do lookup and save result in orientation field.

mavros/launch/px4_config.yaml
TSC21: Where do I add it? There's no place for enums there right now and this is not specific for `distance_sensor` only.

mavros_extras/src/plugins/distance_sensor.cpp
TSC21: But if I leave it as intenger, how can https://github.com/mavlink/mavros/pull/342/files#diff-57d2f68a3e0e5cfa2db14ac20d8845ddR117 be a string???

mavros_extras/src/plugins/distance_sensor.cpp
TSC21: I'm not understanding this. `orientation` param is supposed to be a string (like `"LOCAL_NED"`). Can you suggest a way of receiving as a string a treat it as an integer? Cause that's what I'm trying to do since ever!

mavros_extras/src/plugins/distance_sensor.cpp
TSC21: `p->orientation` has to be a string, but how do I treat it as an integer? Isn't `p->orientation` same as https://github.com/mavlink/mavros/pull/342/files#diff-2adf7423447a0e5f4880b56161efdda9R51?

mavros_extras/src/plugins/distance_sensor.cpp
TSC21: Instead of saying "Great!" you could make yourself clear...

mavros_extras/src/plugins/distance_sensor.cpp
TSC21: Also, this is supposed to be https://github.com/mavlink/mavros/pull/342/files#diff-2adf7423447a0e5f4880b56161efdda9R51, so how can I init as an integer, use it as a string (which it is)?!

mavros/launch/px4_config.yaml
vooon: Maybe: http://wiki.ros.org/mavros/OrientationEnum

TSC21:
> You should pull orientation parameter to temporary std::string, then do lookup and save result in orientation field.

I'm kinda lost so I restored to default (https://github.com/mavlink/mavros/commit/ad7a7b45f1b176027a57addb79af7c3a8e1a699e).
Also I don't know what you pretend (here)[https://github.com/mavlink/mavros/pull/342#discussion_r34915116].

TSC21:
> You should pull orientation parameter to temporary std::string, then do lookup and save result in orientation field.

Don't know what you mean...

mavros_extras/src/plugins/distance_sensor.cpp
```diff
@@ -43,13 +43,13 @@ class DistanceSensorItem {
...
 
 	// params
-	bool is_subscriber;		//!< this item is a subscriber, else is a publisher
-	bool send_tf;			//!< defines if a transform is sent or not
-	uint8_t sensor_id;		//!< id of the sensor
+	bool is_subscriber;	//!< this item is a subscriber, else is a publisher
+	bool send_tf;		//!< defines if a transform is sent or not
+	uint8_t sensor_id;	//!< id of the sensor
 	double field_of_view;	//!< FOV of the sensor
 	tf::Vector3 position;	//!< sensor position
```
vooon: That's one field that need to change type: Eigen::Vector3d.

mavros/launch/px4_config.yaml
TSC21: I think it should have a general Enumerations page like http://wiki.ros.org/mavros/Enumerations. And then we add them all there (there's also others like `MAV_FRAME`).

mavros_extras/src/plugins/distance_sensor.cpp
TSC21: OK. But please(!), instead, go to specific lines to change so to orientation handling work ok (that's the purpose of this PR).

mavros/launch/px4_config.yaml
vooon: Ok.

mavros/include/mavros/mavros_uas.h
TSC21: > Because this reduces object copying. What you want change in UAS::sensor_orientation_matching()?

This I undestand, but:
> Because there may be lots of configs with number representation. Why not just simply try to fall back to old behavior?

... I can't. Can you code an example please?

vooon:
Rebased and merged.

PR #348, opened 2015-07-23
Px4 posix ros sitl

devbharat:
This launch file is called from PX4 launch file to start the gazebo posix sitl.


vooon:
Rebased in master.

PR #351, opened 2015-07-28
Fixed bug in send_attitude_target()

anuragmakineni:
The transformed quaternion wasn't being passed to set_attitude_target(), resulting in an incorrect attitude setpoint. I've now fixed this issue.

### DIFF
--- a/mavros/src/plugins/setpoint_attitude.cpp
+++ b/mavros/src/plugins/setpoint_attitude.cpp
@@ -119,10 +119,17 @@ private:
 		// Eigen use same convention as mavlink: w x y z
 		Eigen::Map<Eigen::Quaternionf> q_out(q);
 		q_out = UAS::transform_frame_enu_ned(Eigen::Quaterniond(tr.rotation())).cast<float>();
+		
+		float q_transformed[4];
+		
+		q_transformed[0] = q_out.w();
+		q_transformed[1] = q_out.x();
+		q_transformed[2] = q_out.y();
+		q_transformed[3] = q_out.z();
 
 		set_attitude_target(stamp.toNSec() / 1000000,
 				ignore_all_except_q,
-				q,
+				q_transformed,
 				0.0, 0.0, 0.0,
 				0.0);
 	}


LorenzMeier:
Have you flight-tested this? The fix looks correct on the math level, but we need a system-level verification as well.

anuragmakineni:
Yes. Flew with this code yesterday. 

-
Anurag

vooon:
I don't see change. `Eigen::Map<>` should place values in q[4] in wxyz, so your code does nothing.

vooon:
Hmm, here is the test:
```cpp
/* Simple test for Eigen::Map
 * Compile:
 *     g++ -I/usr/include/eigen3 -I/usr/src/gtest -pthread /usr/src/gtest/src/gtest-all.cc /tmp/test.cpp 
 */

#include <gtest/gtest.h>
#include <Eigen/Core>
#include <Eigen/Geometry>

const Eigen::Quaterniond Q0(
		Eigen::AngleAxisd(1.0, Eigen::Vector3d::UnitZ()) *
		Eigen::AngleAxisd(2.0, Eigen::Vector3d::UnitY()) *
		Eigen::AngleAxisd(3.0, Eigen::Vector3d::UnitX()));

const double epsilon = 1e-9;

TEST(QMAP, hand_made_copy)
{
	Eigen::Quaternionf q1 = Q0.cast<float>();
	float q2[4];

	// hand made copy
	q2[0] = q1.w();
	q2[1] = q1.x();
	q2[2] = q1.y();
	q2[3] = q1.z();

	EXPECT_NEAR(q1.w(), q2[0], epsilon);
	EXPECT_NEAR(q1.x(), q2[1], epsilon);
	EXPECT_NEAR(q1.y(), q2[2], epsilon);
	EXPECT_NEAR(q1.z(), q2[3], epsilon);
}

TEST(QMAP, eigen_map_storage)
{
	float q2[4];
	Eigen::Map<Eigen::Quaternionf> q1(q2);

	// Map copy
	q1 = Q0.cast<float>();

	EXPECT_NEAR(q1.w(), q2[0], epsilon);
	EXPECT_NEAR(q1.x(), q2[1], epsilon);
	EXPECT_NEAR(q1.y(), q2[2], epsilon);
	EXPECT_NEAR(q1.z(), q2[3], epsilon);
}

int main(int argc, char *argv[])
{
	::testing::InitGoogleTest(&argc, argv);
	return RUN_ALL_TESTS();
}
```

And it fails. So bug confirmed. But there should be another way to fix Map.

PR #360, opened 2015-08-04
global_position: update pose and twist to odom msg

TSC21:


mavros/src/plugins/global_position.cpp
```diff
@@ -68,8 +68,7 @@ class GlobalPositionPlugin : public MavRosPlugin {
 
 		// fused global position
 		gp_fix_pub = gp_nh.advertise<sensor_msgs::NavSatFix>("global", 10);
-		gp_pos_pub = gp_nh.advertise<geometry_msgs::PoseWithCovarianceStamped>("local", 10);
-		gp_vel_pub = gp_nh.advertise<geometry_msgs::TwistStamped>("gp_vel", 10);
+		gp_odom_pub = gp_nh.advertise<nav_msgs::Odometry>("local", 10);
```
vooon: Maybe name it "odom"? Users will confused on topic type change.

TSC21: Well turns out to gives us local position. This is just to have congruence with `global`, which gives the fix topic. What do you think?

mavros/src/plugins/global_position.cpp
```diff
@@ -233,18 +237,15 @@ class GlobalPositionPlugin : public MavRosPlugin {
...
-		pose_cov->header = header;
-		pose_cov->pose.pose.position.x = easting;
-		pose_cov->pose.pose.position.y = northing;
-		pose_cov->pose.pose.position.z = relative_alt->data;
-
-		// XXX We really need attitude on UTM?
-		pose_cov->pose.pose.orientation = uas->get_attitude_orientation();
+		odom->pose.pose.position.x = easting;
+		odom->pose.pose.position.y = northing;
+		odom->pose.pose.position.z = relative_alt->data;
```
vooon: You moved assign of `relative_alt->data` below that line.

mavros/src/plugins/global_position.cpp
vooon: Ok, leave as is.

TSC21: Anyhow, from what I know, no one is using data from this topic yet (fused gps data), so maybe we can keep it (also so we can then calculate the threshold from `local_position/local`).

mavros/src/plugins/global_position.cpp
TSC21: ups! fixing that now... forgot I was using this.

TSC21:
@vooon does it lgtm?

PR #364, opened 2015-08-08
[distance_sensor] fix orientation empty error

andre-nguyen:
Basically the distance_sensor plugin could never initialize with the subscribers defined in the config.


vooon:
Thanks!

PR #365, opened 2015-08-09
opencv 3.0/2.4 header compatibility

v01d:
New stable release of opencv v3.0 includes some header changes. These changes, are retro-compatible with previous stable version 2.4, so they should be safe on both versions.


vooon:
Thanks.

PR #377, opened 2015-08-28
add MANUAL_CONTROL mavlink message publisher as a new plugin

v01d:
I've created a new plugin which simply publishes the MANUAL_CONTROL mavlink as a ROS message. This is a little easier to use than the raw RC input.

I would like to also receive switches states (such as AUX1, ...) but it appears the pixhawk does not publish these, AFAIK.

### DIFF
--- a/mavros/launch/px4_pluginlists.yaml
+++ b/mavros/launch/px4_pluginlists.yaml
@@ -5,6 +5,10 @@ plugin_blacklist:
 - image_pub
 - vibration
 - distance_sensor
+#- setpoint_position
+- setpoint_accel
+- setpoint_velocity
+- actuator_control
 
 plugin_whitelist: []
 #- 'sys_*'
--- a/mavros/mavros_plugins.xml
+++ b/mavros/mavros_plugins.xml
@@ -56,5 +56,8 @@
 	<class name="actuator_control" type="mavplugin::ActuatorControlPlugin" base_class_type="mavplugin::MavRosPlugin">
 		<description>Send direct controls values to the actuators</description>
 	</class>
+  <class name="manual_controls" type="mavplugin::ManualControlsPlugin" base_class_type="mavplugin::MavRosPlugin">
+    <description>Publish manual control values</description>
+  </class>
 </library>
 
--- a/mavros/scripts/mavparam
+++ b/mavros/scripts/mavparam
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 # vim:set ts=4 sw=4 et:
 #
 # Copyright 2014 Vladimir Ermakov.
--- a/mavros/src/plugins/manual_controls.cpp
+++ b/mavros/src/plugins/manual_controls.cpp
@@ -1,0 +1,70 @@
+/**
+ * @brief ManualControls plugin
+ * @file manual_controls.cpp
+ * @author Matias Nitsche <mnitsche@dc.uba.ar>
+ *
+ * @addtogroup plugin
+ * @{
+ */
+/*
+ * This file is part of the mavros package and subject to the license terms
+ * in the top-level LICENSE file of the mavros repository.
+ * https://github.com/mavlink/mavros/tree/master/LICENSE.md
+ */
+
+#include <mavros/mavros_plugin.h>
+#include <pluginlib/class_list_macros.h>
+
+#include <mavros_msgs/ManualControls.h>
+
+namespace mavplugin {
+/**
+ * @brief Manual Controls plugin
+ */
+class ManualControlsPlugin : public MavRosPlugin {
+public:
+  ManualControlsPlugin() :
+    rc_nh("~manual_controls"),
+    uas(nullptr)
+	{ };
+
+	void initialize(UAS &uas_)
+	{
+		uas = &uas_;
+
+    controls_pub = rc_nh.advertise<mavros_msgs::ManualControls>("controls", 10);
+    //uas->sig_connection_changed.connect(boost::bind(&RCIOPlugin::connection_cb, this, _1));
+	};
+
+	const message_map get_rx_handlers() {
+		return {
+             MESSAGE_HANDLER(MAVLINK_MSG_ID_MANUAL_CONTROL, &ManualControlsPlugin::handle_manual_control),
+		};
+	}
+
+private:
+	ros::NodeHandle rc_nh;
+	UAS *uas;
+
+  ros::Publisher controls_pub;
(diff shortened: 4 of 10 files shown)


mavros/src/plugins/manual_controls.cpp
```diff
@@ -0,0 +1,70 @@
...
+  void handle_manual_control(const mavlink_message_t *msg, uint8_t sysid, uint8_t compid) {
+    mavlink_manual_control_t manual_control;
+    mavlink_msg_manual_control_decode(msg, &manual_control);
+
+    mavros_msgs::ManualControls manual_controls_ros;
+    manual_controls_ros.header.stamp = ros::Time::now(); // correct?
+    manual_controls_ros.x = manual_control.x;
+    manual_controls_ros.y = manual_control.y;
+    manual_controls_ros.z = manual_control.z;
+    manual_controls_ros.r = manual_control.r;
```
vooon: Please add buttons field too.

mavros_msgs/CMakeLists.txt
```diff
@@ -17,6 +17,7 @@ add_message_files(
   OpticalFlowRad.msg
   OverrideRCIn.msg
   ParamValue.msg
+  ManualControls.msg
```
vooon: Why plural form? Please name it `ManualControl.msg` like corresponding mavlink message.

mavros/src/plugins/manual_controls.cpp
```diff
@@ -0,0 +1,70 @@
...
+	};
+
+	const message_map get_rx_handlers() {
+		return {
+             MESSAGE_HANDLER(MAVLINK_MSG_ID_MANUAL_CONTROL, &ManualControlsPlugin::handle_manual_control),
+		};
+	}
+
+private:
+	ros::NodeHandle rc_nh;
```
vooon: Please rename node handle.

mavros/src/plugins/manual_controls.cpp
```diff
@@ -0,0 +1,70 @@
...
+#include <mavros/mavros_plugin.h>
+#include <pluginlib/class_list_macros.h>
+
+#include <mavros_msgs/ManualControls.h>
+
+namespace mavplugin {
+/**
+ * @brief Manual Controls plugin
+ */
+class ManualControlsPlugin : public MavRosPlugin {
```
vooon: Why plural?

mavros_msgs/CMakeLists.txt
v01d: I can switch to singular, but I was worried it could be interpreted (by a
ROS user) as an input (i.e.: to have manual control). If you'd prefer
singular, I can change it.

Regarding the buttons field. I did not include it since pixhawk seems to
send zero. Should I add it anyway?

mavros/src/plugins/manual_controls.cpp
```diff
@@ -0,0 +1,70 @@
...
+  ManualControlsPlugin() :
+    rc_nh("~manual_controls"),
+    uas(nullptr)
+	{ };
+
+	void initialize(UAS &uas_)
+	{
+		uas = &uas_;
+
+    controls_pub = rc_nh.advertise<mavros_msgs::ManualControls>("controls", 10);
```
vooon: As i know QGC may use that message for sending joystick data. So it might be useful to add possibility to send that message.

In that case need better name for topic. Something like `~manual_control/to` and `~manual_control/from`.

mavros/src/plugins/manual_controls.cpp
```diff
@@ -0,0 +1,70 @@
+/**
+ * @brief ManualControls plugin
+ * @file manual_controls.cpp
+ * @author Matias Nitsche <mnitsche@dc.uba.ar>
+ *
+ * @addtogroup plugin
+ * @{
+ */
+/*
```
vooon: Please add your copyright string.

mavros/mavros_plugins.xml
```diff
@@ -56,5 +56,8 @@
 	<class name="actuator_control" type="mavplugin::ActuatorControlPlugin" base_class_type="mavplugin::MavRosPlugin">
 		<description>Send direct controls values to the actuators</description>
 	</class>
+  <class name="manual_controls" type="mavplugin::ManualControlsPlugin" base_class_type="mavplugin::MavRosPlugin">
```
vooon: Wrong indentation: 2 spaces instead of tab.

mavros/src/plugins/manual_controls.cpp
v01d: Ok. I have only coded publishing the data sent by the autopilot anyway.

mavros/launch/px4_pluginlists.yaml
```diff
@@ -5,6 +5,10 @@ plugin_blacklist:
 - image_pub
 - vibration
 - distance_sensor
+#- setpoint_position
```
vooon: Not acceptable. please remove that from this PR.

mavros/scripts/mavparam
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
```
vooon: Changing default python is not case of that PR. Please revert that.

Anyway scripts should work under py3 too, but that require to build ros with experimental py3 support.

mavros_msgs/msg/ManualControls.msg
```diff
@@ -0,0 +1,7 @@
+# Manual Control state, all values [-1000,1000]
+std_msgs/Header header
+int16 x
+int16 y
+int16 z
+int16 r
+# buttons is unused, pixhawk does not publish anything here
```
vooon: Please add buttons.

mavros/scripts/mavparam
v01d: Sorry, it seems some local changes I did not wanted to include in the PR
were included. Should I make a new pull request?

mavros_msgs/msg/ManualControls.msg
```diff
@@ -0,0 +1,7 @@
+# Manual Control state, all values [-1000,1000]
+std_msgs/Header header
+int16 x
+int16 y
+int16 z
+int16 r
```
vooon: I think on ROS level we may use float32 and normalize to [-1.0…1.0].

vooon:
Also please use [uncrustify as mentioned in readme](https://github.com/mavlink/mavros/blob/master/mavros/README.md#contributing). Current cpp file is misindented.

PR #378, opened 2015-08-28
Add MANUAL_CONTROL handling with new plugin

v01d:
Cleaned up pull request #377 (I messed up my repo in the process of fixing).

### DIFF
--- a/mavros/src/plugins/manual_control.cpp
+++ b/mavros/src/plugins/manual_control.cpp
@@ -1,0 +1,73 @@
+/**
+ * @brief ManualControls plugin
+ * @file manual_controls.cpp
+ * @author Matias Nitsche <mnitsche@dc.uba.ar>
+ *
+ * @addtogroup plugin
+ * @{
+ */
+/*
+ * * Copyright 2015 Matias Nitsche.
+ *
+ * This file is part of the mavros package and subject to the license terms
+ * in the top-level LICENSE file of the mavros repository.
+ * https://github.com/mavlink/mavros/tree/master/LICENSE.md
+ */
+
+#include <mavros/mavros_plugin.h>
+#include <pluginlib/class_list_macros.h>
+
+#include <mavros_msgs/ManualControl.h>
+
+namespace mavplugin {
+/**
+ * @brief Manual Control plugin
+ */
+class ManualControlPlugin : public MavRosPlugin {
+public:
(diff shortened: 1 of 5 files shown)


mavros/src/plugins/manual_control.cpp
```diff
@@ -0,0 +1,73 @@
...
+	ros::Publisher control_pub;
+
+	/* -*- rx handlers -*- */
+
+	void handle_manual_control(const mavlink_message_t *msg, uint8_t sysid, uint8_t compid) {
+		mavlink_manual_control_t manual_control;
+		mavlink_msg_manual_control_decode(msg, &manual_control);
+
+		mavros_msgs::ManualControl manual_control_ros;
+		manual_control_ros.header.stamp = ros::Time::now();	// correct?
```
vooon: Yes, correct because message don't have any timestamp.

vooon:
Ok, now it looks good to merge. But maybe you implement send code too?

v01d:
I'm not aware that this mavlink message goes both ways. I think it is only
sent by the autopilot, and not received. But I'm not sure.

vooon:
You may check QGC code. In SITL/HIL mode it has ability to send joystick data. I think via that message.

v01d:
To be honest, at the moment I'm not sure If I can manage time to look into
that. I can do it bit further on and provide a separate pull request.

vooon:
Ok.

v01d:
@vooon I'm just testing this using running mavros on my odroid connected to the pixhawk via the TELEM port and I can't see the MANUAL_CONTROL stream there. However, if I connect to the pixhawk from my laptop using USB cable, I do see the stream. Do you know what could be the cause?

vooon:
I think stream rate not configured on TELEM. See [`rc.usb`][1] file.

[1]: https://github.com/PX4/Firmware/blob/9365c5a4383d8702e66b5f0e3f95a1acb3fd5cac/ROMFS/px4fmu_common/init.d/rc.usb

vooon:
Hmm, now stream config for usb is moved to mavlink app, but you still can configure streams via `extras.txt`

v01d:
Ok, thanks!

On Mon, Aug 31, 2015 at 12:24 PM, Vladimir Ermakov <notifications@github.com

PR #391, opened 2015-09-22
Mavros library depends on mavros_msgs headers

jonbinney:
Adding this dependency makes sure that mavros_msgs message headers are generated before the mavros library is built, since it needs those headers.

Otherwise you get intermittent build errors when building using multiple catkin_make threads, such as:
```
[  4%] Building CXX object third_party/mavros/mavros/CMakeFiles/mavros.dir/src/lib/uas_data.cpp.o
In file included from /home/odroid/ros/src/third_party/mavros/mavros/include/mavros/utils.h:21:0,
                 from /home/odroid/ros/src/third_party/mavros/mavros/src/lib/uas_data.cpp:18:
/home/odroid/ros/src/third_party/mavros/mavros_msgs/include/mavros_msgs/mavlink_convert.h:17:33: fatal error: mavros_msgs/Mavlink.h: No such file or directory
 #include <mavros_msgs/Mavlink.h>
                                 ^
compilation terminated.
```

### DIFF
--- a/mavros/CMakeLists.txt
+++ b/mavros/CMakeLists.txt
@@ -87,6 +87,7 @@ target_link_libraries(mavros
   ${catkin_LIBRARIES}
   ${Boost_LIBRARIES}
 )
+add_dependencies(mavros mavros_msgs_generate_messages_cpp)
 
 add_library(mavros_plugins
   src/plugins/dummy.cpp


vooon:
Hmm, i'm never got this error. Perhaps because `catkin` builds `<depend>` before.

jonbinney:
Do you use catkin tools? In that case yes, it builds each package as their own cmake workspace, and orders them using `<depend>`. catkin_make builds everything in one cmake workspace so you need to get the cmake dependencies just right or else you get intermittent errors.

vooon:
Yes, catkin_tools is more friendly and handy than catkin_make.

PR #393, opened 2015-09-28
Local takeoff/land commands

tinito:
I've added support for local takeoff/land, following the implementation of global ones.


mavros/scripts/mavcmd
```diff
@@ -178,6 +213,60 @@ def do_land_cur_gps(args):
     if not done_evt.wait(10.0):
         fault("Something went wrong. Topic timed out.")
 
+def do_takeoff_cur_local(args):
+    done_evt = threading.Event()
+    def pose_cb(msg):
+        print("Taking-off from current pose: X:", msg.pose.position.x, "Y:", msg.pose.position.y)
+        print_if(args.verbose, "With desired X:", args.z,
```
vooon: s/X/Z/

mavros/src/plugins/waypoint.cpp
```diff
@@ -121,6 +121,8 @@ class WaypointItem {
 		case MAV_CMD_NAV_RETURN_TO_LAUNCH:	return "RTL";
 		case MAV_CMD_NAV_LAND:		return "LAND";
 		case MAV_CMD_NAV_TAKEOFF:	return "TAKEOFF";
+		case MAV_CMD_NAV_LAND_LOCAL:	return "LAND_LOCAL";
```
vooon: `s/_LOCAL/\ LOCAL/`

mavros_msgs/srv/CommandTOLLocal.srv
```diff
@@ -0,0 +1,10 @@
+# Common type for Take Off and Landing for local (NED) navigation
+
+float32 min_pitch	# used by takeoff
+float32 yaw
+float32 x
+float32 y
+float32 z
```
vooon: Better to use `geometry_msgs/Vector3` here.

mavros/src/plugins/command.cpp
```diff
@@ -335,6 +340,27 @@ class CommandPlugin : public MavRosPlugin {
...
+				res.success, res.result);
+	}
+
+	bool land_local_cb(mavros_msgs::CommandTOLLocal::Request &req,
+			mavros_msgs::CommandTOLLocal::Response &res) {
+		return send_command_long_and_wait(false,
+				MAV_CMD_NAV_LAND_LOCAL, 1,
+				0, 0, 0,
+				req.yaw,
+				req.x, req.y, req.z,
```
vooon: ROS ENU -> FCU NED transformation is required here.

vooon:
See comments, and does it works?

mhkabir:
Hi @tinito We'll need a corresponding flightstack implementation to support these. I'm working on it with some others, so if you're interested to contribute, it'd be great!

mavros_msgs/srv/CommandTOLLocal.srv
tinito: geometry_msgs/Vector3 (vector) or geometry_msgs/Point (point in the 3D space)?
I would go for:
```
float32 min_pitch	# used by takeoff
float32 yaw
geometry_msgs/Point position
```

tinito:
@vooon: it "works" meaning that I receive the commands 23 and 24 on the FCU which rejects them as not supported, as we need the firmware side code

@mhkabir: I know, and yes I would like to contribute, I really need offboard local takeoff/land so I will work on it. I've asked to join the px4dev group but I'm still waiting for approval.
Is there any public branch I can work on?

mavros/src/plugins/command.cpp
tinito: Done, just check if it is correct please

mavros/src/plugins/command.cpp
```diff
@@ -335,6 +340,37 @@ class CommandPlugin : public MavRosPlugin {
 				res.success, res.result);
 	}
 
+	bool takeoff_local_cb(mavros_msgs::CommandTOLLocal::Request &req,
+			mavros_msgs::CommandTOLLocal::Response &res) {
+		/* Convert from NED to ENU */
+		float x = req.y;
+		float y = req.x;
+		float z = -req.z;
```
vooon: Nope, please use `UAS::transform_frame_enu_ned(Eigen::Vector3d(x, y, z))`.

mavros_msgs/srv/CommandTOLLocal.srv
vooon: Ok. Yes, Point better here.

mavros/src/plugins/command.cpp
tinito: Done, I didn't know about that, thanks.

mavros_msgs/srv/CommandTOLLocal.srv
tinito: Done, I'll go to the lab to test it in a few hours

mhkabir:
@tinito We haven't actually started working on it yet.

@AndreasAntener and I are working on it. Maybe we can schedule a chat/call? We'll let you know the potential pitfalls and things of note that have come up in our discussions. We came up with a mostly solid workflow/roadmap which would get local takeoff/land working. Time is the only thing that is short :P

Can you please message me on Gitter (www.gitter.im) ? :)

tinito:
Ok, I've tested it and communication works, I get the right messages in mavlink.
Started working on FCU support today.

TSC21:
@tinito good to know you are picking up on this. I specifically added `MAV_CMD_NAV_LAND_LOCAL` and `MAV_CMD_NAV_TAKEOFF_LOCAL` to this same purpose, but I came to have no time to continue the implementation on the FCU side.
Let me know if you need assistance in the mean time. It would be good if you could sync the implementation with the usage of `LANDING_TARGET` message, which could be used to issue a landing behavior whenever a landing target is detect and the conditions are gathered so the multicopter can safely land.

tinito:
@TSC21 I've planned to implement missions too, but digging into the code I've noticed that we first need to add support for takeoff/land setpoints and integrate them in mavros to test first offboard. Then I'll go into missions and I'll look to the LANDING_TARGET message too.

@vooon I've added support for SETPOINT_TYPE_TAKEOFF/LAND on FCU side (see https://github.com/PX4/Firmware/compare/master...tinito:local_tol ), but now we need to add the corresponding implementation on mavros side. With @AndreasAntener we agreed to use two more bits of `set_position_target_local_ned.type_mask` to identify TAKEOFF/LAND setpoints.

Now we need to add support on mavros side, I see several options:
- add specific topics for takeoff/land setpoints, handling `geometry_msgs/PoseStamped` messages
- add the setpoint type to messages (not nice, it would be not compatible with most existing code)
- add a topic (a service maybe) to specify the status (takeoff/inair/land)
Which way would you prefer?
Any other solution is welcome of course.

vooon:
I think another topic for setpoits are ok.
But really there better to hear @TSC21, @mhkabir and other setpoint users.

AndreasAntener:
@tinito Thanks for digging into this :)
Another option would be to use a separate topic like you suggested, but create a new message for it that contains PoseStamped plus a way to specify the complete bitmask. I got stuck on one or two occasions where it would have been great to be able to do this (mostly for testing though). For simplicity the message could contain a few constants with the normally needed bitmasks. What do you guys think?

LorenzMeier:
@tinito The PX4 mailing list is this one: https://groups.google.com/forum/#!forum/px4users just join.

AndreasAntener:
@tinito And in addition to my previous answer: I don't think this needs a separate plugin because that just creates a lot of copied code from the original one. You could extend the `setpoint_plugin` and add another topic, with the message I suggested (assuming this fits into the plugin philosophy, @vooon could comment on that :) ).

We just discussing the offboard interface in general, and exposing the bitmask in ROS seems like a good idea: https://github.com/PX4/Firmware/pull/3004#issuecomment-148916149

vooon:
Usually i prefer unix principle: one program for one task. But because that task fits to just extend existing code id prefer that way.

I reviewed code of setpoint_takeoff_land and not understood what it should do? I don't see any difference from setpoint_position. It's almost identical. Also land/takeoff do not use any options to limit accelerations (fields unused), but at least you must define low acceleration on landing.

Code duplication is bad thind, so now that PR not pass :)

@AndreasAntener exposing type_mask require non-standard message type. That's might be ok, but if we really need that.

AndreasAntener:
> exposing type_mask require non-standard message type. That's might be ok, but if we really need that.

@vooon how's your philosophy on non-standard messages then? ;) ROS is basically not MAVLink compliant.. The upside to it is that people don't have to switch. They can just use the same topic as before, but if they need fine grained control they can switch to the new topic with a non-standard message. Do you see another solution? I'm not a big fan of things like separate commands/topics with which you can then influence the additional property (bitmask) because it always leads to race conditions (the same problem there is with attitude/thrust..)

vooon:
@AndreasAntener i'm thinking to use Message Filters to solve attitude/thrust issue. But we need to check that it really works.

Another topic/message is not issue, but i'm think that already exist interfaces is preferable.

TSC21:
@tinito are you with free time to rebase this to latest master, considering latest code structure? @mhkabir any input to implement support on the FW side?

TSC21:
@tinito ping!

TSC21:
I'm closing this for now. This will be reopened in another PR after fixes and testing.

PR #395, opened 2015-10-08
Publish extended state to ROS

AndreasAntener:
Forwards the newly introduced extended system state (available in PX4 now) to ROS.

### DIFF
--- a/mavros/src/plugins/sys_status.cpp
+++ b/mavros/src/plugins/sys_status.cpp
@@ -18,6 +18,7 @@
 #include <pluginlib/class_list_macros.h>
 
 #include <mavros_msgs/State.h>
+#include <mavros_msgs/ExtendedState.h>
 #include <mavros_msgs/BatteryStatus.h>
 #include <mavros_msgs/StreamRate.h>
 #include <mavros_msgs/SetMode.h>
@@ -404,6 +405,7 @@ public:
 		uas->sig_connection_changed.connect(boost::bind(&SystemStatusPlugin::connection_cb, this, _1));
 
 		state_pub = nh.advertise<mavros_msgs::State>("state", 10, true);
+		extended_state_pub = nh.advertise<mavros_msgs::ExtendedState>("extended_state", 10);
 		batt_pub = nh.advertise<mavros_msgs::BatteryStatus>("battery", 10);
 		rate_srv = nh.advertiseService("set_stream_rate", &SystemStatusPlugin::set_rate_cb, this);
 		mode_srv = nh.advertiseService("set_mode", &SystemStatusPlugin::set_mode_cb, this);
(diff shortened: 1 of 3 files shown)


mavros_msgs/msg/ExtendedState.msg
```diff
@@ -0,0 +1,15 @@
...
+
+uint8 VTOL_STATE_UNDEFINED = 0
+uint8 VTOL_STATE_TRANSITION_TO_FW = 0
+uint8 VTOL_STATE_TRANSITION_TO_MC = 0
+uint8 VTOL_STATE_MC = 0
+uint8 VTOL_STATE_FW = 0
+
+uint8 LANDED_STATE_UNDEFINED = 0
+uint8 LANDED_STATE_ON_GROUND = 0
+uint8 LANDED_STATE_IN_AIR = 0
```
vooon: All zeroes?!?

mavros/src/plugins/sys_status.cpp
```diff
@@ -614,6 +618,18 @@ class SystemStatusPlugin : public MavRosPlugin
 		hb_diag.tick(hb.type, hb.autopilot, state_msg->mode, hb.system_status);
 	}
 
+	void handle_extended_sys_state(const mavlink_message_t *msg, uint8_t sysid, uint8_t compid) {
+		mavlink_extended_sys_state_t state;
+		mavlink_msg_extended_sys_state_decode(msg, &state);
+
+		auto state_msg = boost::make_shared<mavros_msgs::ExtendedState>();
+		state_msg->header.stamp = ros::Time::now();
```
vooon: Sad that message don't have timestamp.

mavros_msgs/msg/ExtendedState.msg
```diff
@@ -0,0 +1,15 @@
+# Extended autopilot state
```
vooon: I think to add link to mavlink doc. e.g.

https://pixhawk.ethz.ch/mavlink/#EXTENDED_SYS_STATE

AndreasAntener: Added

mavros/src/plugins/sys_status.cpp
AndreasAntener: What do you mean?

mhkabir: Should've had a timestamp in the message definition.

vooon: Just i think that all mavlink messages should have timestamps, but really only half has some stamps. And more over that stamps has different types.

Anyway mavlink 2 propose change of that. But this thread is offtopic.

AndreasAntener: Ok, I was wondering because the other status messages I looked at had none ;) . So it sounds like we just leave this off in that case and deal with it properly on mavlink 2

PR #420, opened 2015-11-02
Added velocity output of FCU's local position estimate to local_position plugin

scott-eddy:
This adds a TwistStamped velocity topic to the output of the local_position node.  The velocity messages contains a vector3 of linear velocities in the ENU frame and a vector3 of the angular velocitires in the ENU frame.  Note that the angular velocity is expressed in the ENU frame for consistency though I would venture to guess people will quite possibly confuse this.  I'd be happy to change it to the body frame if people think that is best.  

The published topics of the plugin are now:
local_position/local/position(geometry_msgs/PoseStamped)
local_position/local/velocity(geometry_msgs/TwistStamped)

### DIFF
--- a/mavros/src/plugins/local_position.cpp
+++ b/mavros/src/plugins/local_position.cpp
@@ -41,15 +41,14 @@ public:
 		uas = &uas_;
 
 		// general params
-		
 		lp_nh.param<std::string>("frame_id", frame_id, "fcu");
 		// tf subsection
 		lp_nh.param("tf/send", tf_send, true);
 		lp_nh.param<std::string>("tf/frame_id", tf_frame_id, "local_origin");
 		lp_nh.param<std::string>("tf/child_frame_id", tf_child_frame_id, "fcu");
 
-		local_position = lp_nh.advertise<geometry_msgs::PoseStamped>("local/position", 10);
-		local_velocity = lp_nh.advertise<geometry_msgs::TwistStamped>("local/velocity", 10);
+		local_position = lp_nh.advertise<geometry_msgs::PoseStamped>("pose", 10);
+		local_velocity = lp_nh.advertise<geometry_msgs::TwistStamped>("velocity", 10);
 	}
 
 	const message_map get_rx_handlers() {
--- a/mavros/src/plugins/local_position.cpp
+++ b/mavros/src/plugins/local_position.cpp
@@ -20,6 +20,7 @@
 #include <eigen_conversions/eigen_msg.h>
 
 #include <geometry_msgs/PoseStamped.h>
+#include <geometry_msgs/TwistStamped.h>
 #include <geometry_msgs/TransformStamped.h>
 
 namespace mavplugin {
@@ -46,7 +47,8 @@ public:
 		lp_nh.param<std::string>("tf/frame_id", tf_frame_id, "local_origin");
 		lp_nh.param<std::string>("tf/child_frame_id", tf_child_frame_id, "fcu");
 
-		local_position = lp_nh.advertise<geometry_msgs::PoseStamped>("local", 10);
+		local_position = lp_nh.advertise<geometry_msgs::PoseStamped>("local/position", 10);
+		local_velocity = lp_nh.advertise<geometry_msgs::TwistStamped>("local/velocity", 10);
 	}
 
 	const message_map get_rx_handlers() {
@@ -60,6 +62,7 @@ private:
 	UAS *uas;
 
 	ros::Publisher local_position;
+	ros::Publisher local_velocity;
 
 	std::string frame_id;		//!< frame for Pose
 	std::string tf_frame_id;	//!< origin for TF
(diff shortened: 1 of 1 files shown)


mavros/src/plugins/local_position.cpp
```diff
@@ -46,7 +47,8 @@ class LocalPositionPlugin : public MavRosPlugin {
 		lp_nh.param<std::string>("tf/frame_id", tf_frame_id, "local_origin");
 		lp_nh.param<std::string>("tf/child_frame_id", tf_child_frame_id, "fcu");
 
-		local_position = lp_nh.advertise<geometry_msgs::PoseStamped>("local", 10);
+		local_position = lp_nh.advertise<geometry_msgs::PoseStamped>("local/position", 10);
+		local_velocity = lp_nh.advertise<geometry_msgs::TwistStamped>("local/velocity", 10);
```
vooon: Nope, changing topic names usually return with lots questions. While sometime that is absolutely required, but now i'd prefer only addition.

I think "velocity" is enough (`~local_position/velocity`).

mavros/src/plugins/local_position.cpp
```diff
@@ -68,19 +71,32 @@ class LocalPositionPlugin : public MavRosPlugin {
 
 	void handle_local_position_ned(const mavlink_message_t *msg, uint8_t sysid, uint8_t compid) {
 		mavlink_local_position_ned_t pos_ned;
+		Eigen::Vector3d angular_velocity;
```
vooon: Please do not split this `mavlink_*` line pair.

mavros/src/plugins/local_position.cpp
```diff
@@ -68,19 +71,32 @@ class LocalPositionPlugin : public MavRosPlugin {
...
 		mavlink_msg_local_position_ned_decode(msg, &pos_ned);
 
 		auto position = UAS::transform_frame_ned_enu(Eigen::Vector3d(pos_ned.x, pos_ned.y, pos_ned.z));
+		auto velocity = UAS::transform_frame_ned_enu(Eigen::Vector3d(pos_ned.vx, pos_ned.vy, pos_ned.vz));
 		auto orientation = uas->get_attitude_orientation();
+		auto imu_data = uas->get_attitude_imu();
+		tf::vectorMsgToEigen(imu_data->angular_velocity, angular_velocity);
+		angular_velocity = UAS::transform_frame_ned_enu(angular_velocity);
```
vooon: Why?! IMU data is already converted. You may simply copy geometry_msgs::Vector3.

Also because you get whole imu message there no reason for getting orientation, it is inside sensor_msgs::IMU.

mavros/src/plugins/local_position.cpp
```diff
@@ -68,19 +71,32 @@ class LocalPositionPlugin : public MavRosPlugin {
...
+		
 
 		auto pose = boost::make_shared<geometry_msgs::PoseStamped>();
+		auto twist = boost::make_shared<geometry_msgs::TwistStamped>();
 
 		pose->header = uas->synchronized_header(frame_id, pos_ned.time_boot_ms);
+		twist->header = uas->synchronized_header(frame_id, pos_ned.time_boot_ms);
```
vooon: I think you may simply copy header from `pose`.

mavros/src/plugins/local_position.cpp
```diff
@@ -68,19 +71,32 @@ class LocalPositionPlugin : public MavRosPlugin {
...
 
 		tf::pointEigenToMsg(position, pose->pose.position);
 		pose->pose.orientation = orientation;
 
+		tf::vectorEigenToMsg(velocity,twist->twist.linear);
+		tf::vectorEigenToMsg(angular_velocity,twist->twist.angular);
+		
+
```
vooon: Unneeded whitespace.

mavros/src/plugins/local_position.cpp
mhkabir: I like this topic naming though. Makes sense.

mavros/src/plugins/local_position.cpp
scott-eddy: I will correct this, I assumed the IMU data would be represented in the body frame.  Also the orientation was left over from the last implementation but I'll only grab the one.

mavros/src/plugins/local_position.cpp
scott-eddy: I'm also partial to this topic naming.  As an end user it would be confusing to have local_position/local and local_position/velocity, as "local" really doesn't relate to a physically tangible state.  I'd be happy to update documentation if it meant we could keep this naming (i.e. the ros wiki, PX4 Devguide, PX4 wiki, and github README).

However I know topic renaming can be a nightmare so I would be okay with changing to "local" and "velocity" if otherwise it meant this PR got rejected.

vooon: I just do not see reason for local_position/**local**/velocity. Obviously it is already *local*.
Maybe it may have sense to rename `local_position/local` to `local_position/pose`. But really you see topic name and it's type so i suppose keeping current name require less effort.

scott-eddy: I see your point about local being redundant.  Though I think that local_position/local is a bad name.  If local_position/**local**/velocity is redundant than local_position/local is really bad naming because it is still obviously local but doesn't tell you what it is.

I would vote to move to
```
local_position/pose
local_position/velocity
```
@vooon does that work?

vooon: Yes.

mavros/src/plugins/local_position.cpp
```diff
@@ -40,13 +41,15 @@ class LocalPositionPlugin : public MavRosPlugin {
 		uas = &uas_;
 
 		// general params
+
```
vooon: Unneeded whitespace.

scott-eddy:
@vooon I think this is good to go now

vooon:
Yes. Thank you.

PR #425, opened 2015-11-09
Remove warning about rc_io support if not APM

AndreasAntener:
**untested** @scott-eddy

### DIFF
--- a/mavros/src/plugins/rc_io.cpp
+++ b/mavros/src/plugins/rc_io.cpp
@@ -213,7 +213,7 @@ private:
 	}
 
 	void override_cb(const mavros_msgs::OverrideRCIn::ConstPtr req) {
-		if (!uas->is_ardupilotmega())
+		if (!uas->is_ardupilotmega() && !uas->is_px4())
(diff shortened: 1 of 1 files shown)


AndreasAntener:
Support in PX4 for implemented a while back: https://github.com/PX4/Firmware/blob/9647f0622c22b9b1dd3603970a75bb3006a056e7/src/modules/mavlink/mavlink_receiver.cpp#L1045

mavros/src/plugins/rc_io.cpp
```diff
@@ -213,9 +213,6 @@ class RCIOPlugin : public MavRosPlugin {
 	}
 
 	void override_cb(const mavros_msgs::OverrideRCIn::ConstPtr req) {
-		if (!uas->is_ardupilotmega())
```
vooon: I prefer change condition to:
```cpp
if (!uas->is_ardupilotmega() && !uas->is_px4())
```

AndreasAntener: Makes sense

vooon:
Thank you!

PR #428, opened 2015-11-15
WIP: transform yaw and yaw rate from ENU to NED

AndreasAntener:
@vooon My integration tests found a bug in `setpoint_raw`. In `setpoint_position` the orientation (quaternion) is handled by the transform methods. But we don't have a Q in `setpoint_raw`, we just copy yaw/yaw_rate. It needs to be inverted. I only changed it for `local_cb` to test and to show you, do you have a suggestion how to fix this properly? I'm a little lost on the transformation templates ;)

### DIFF
--- a/mavros/src/plugins/setpoint_raw.cpp
+++ b/mavros/src/plugins/setpoint_raw.cpp
@@ -77,6 +77,8 @@ private:
 		auto position = UAS::transform_frame_ned_enu(Eigen::Vector3d(tgt.x, tgt.y, tgt.z));
 		auto velocity = UAS::transform_frame_ned_enu(Eigen::Vector3d(tgt.vx, tgt.vy, tgt.vz));
 		auto af = UAS::transform_frame_ned_enu(Eigen::Vector3d(tgt.afx, tgt.afy, tgt.afz));
+		float yaw = UAS::transform_frame_yaw_ned_enu(tgt.yaw);
+		float yaw_rate = UAS::transform_frame_yaw_ned_enu(tgt.yaw_rate);
 
 		auto target = boost::make_shared<mavros_msgs::PositionTarget>();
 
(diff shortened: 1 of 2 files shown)


vooon:
Hmm, i don't think that we need to build quaternion just for yaw inversion. But patch may be simplifyed.

mavros/src/plugins/setpoint_raw.cpp
```diff
@@ -196,7 +199,7 @@ class SetpointRawPlugin : public MavRosPlugin,
 				position.x(), position.y(), position.z(),
 				velocity.x(), velocity.y(), velocity.z(),
 				af.x(), af.y(), af.z(),
-				req->yaw, req->yaw_rate);
```
vooon: Why not:
```cpp
-(req->yaw), -(req->yaw_rate)); // enu-ned transformation
```

vooon:
Also perhaps global part should be fixed same way. And do not forget loopback topics!

mavros/src/plugins/setpoint_raw.cpp
AndreasAntener: I just wanted to make the transformation really obvious, like the others. A library function would be good I think. Will add one and see how it looks, and also update the rest

vooon: You may add something like that to UAS:
```cpp
// "private" method
static inline double transform_frame_yaw(double yaw) {
   return -yaw;
}

// public API, like transform_frame_xxx_yyy()
static inline double transform_frame_yaw_enu_ned();
static inline double transform_frame_yaw_ned_enu();
```

vooon: I don't think that adding overloaded function here is good idea - with vector3 or quaternion you know what expect from transforms. With single double - nope (it is roll, pitch or yaw?).

vooon: Alternatively we may do:
```cpp
static inline double transform_frame_yaw(double yaw) {
    return transform_framee(Eigen::Vector3d(0, 0, yaw)).z();
}
```

But it is obviously inefficient.

AndreasAntener:
@vooon updated

vooon:
Looks ok. Merged, Thans!

PR #435, opened 2015-11-20
Altitude plugin

AndreasAntener:
@vooon Added a new plugin for the altitude message. PX4 implementation is in the queue. https://github.com/PX4/Firmware/pull/3234

### DIFF
--- a/mavros/src/plugins/altitude.cpp
+++ b/mavros/src/plugins/altitude.cpp
@@ -1,0 +1,72 @@
+/**
+ * @brief Altitude plugin
+ * @file altitude.cpp
+ * @author Andreas Antener <andreas@uaventure.com>
+ *
+ * @addtogroup plugin
+ * @{
+ */
+/*
+ * Copyright 2015 Vladimir Ermakov.
+ *
+ * This file is part of the mavros package and subject to the license terms
+ * in the top-level LICENSE file of the mavros repository.
+ * https://github.com/mavlink/mavros/tree/master/LICENSE.md
+ */
+
+#include <mavros/mavros_plugin.h>
+#include <pluginlib/class_list_macros.h>
+
+#include <mavros_msgs/Altitude.h>
+
+namespace mavplugin {
+/**
+ * @brief Altitude plugin.
+ */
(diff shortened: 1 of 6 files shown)


mavros/src/plugins/altitude.cpp
```diff
@@ -0,0 +1,72 @@
+/**
+ * @brief Altitude plugin
+ * @file altitude.cpp
+ * @author Andreas Antener <andreas@uaventure.com>
+ *
+ * @addtogroup plugin
+ * @{
+ */
+/*
+ * Copyright 2015 Vladimir Ermakov.
```
vooon: Why me?

AndreasAntener: Thanks, copy paste and such..

mavros_msgs/msg/Altitude.msg
```diff
@@ -0,0 +1,12 @@
...
+#
+# https://pixhawk.ethz.ch/mavlink/#ALTITUDE
+
+std_msgs/Header header
+
+float32 altitude_monotonic
+float32 altitude_amsl
+float32 altitude_local
+float32 altitude_relative
+float32 altitude_terrain
```
vooon: Why to repeat `altitude`?

Altitude.monotonic || Altitude.terrain?

vooon: @LorenzMeier how old this message? Why it lack timestamp?

AndreasAntener: You're right, makes no sense

AndreasAntener:
@vooon @LorenzMeier About the missing timestamp: is it just to factor out transmission delays? Or about syncing sensor data? (the later would require multiple timestamps then since the data comes from multiple sources)

LorenzMeier:
Transmission. Let's add it

AndreasAntener:
@vooon This is good to go with latest mavlink

vooon:
Ok, i will release mavlink update.

- Indigo: https://github.com/ros/rosdistro/pull/9867
- Jade:

PR #443, opened 2015-11-26
Added Optitrack Motive full compatibility to mocap position

rosn:
Implemented Optitrack Motive coordinates uses (zxy)
   + Added /mavros/mocap/use_motive_zxy param (true use zxy ENU coordiates, false use xyz ENU coordinates)
   + Added auto transform from zxy to xyz (if use_motive_zxy set to true)

### DIFF
--- a/mavros/launch/apm_config.yaml
+++ b/mavros/launch/apm_config.yaml
@@ -120,6 +120,8 @@ mocap:
   # select mocap source
   use_tf: false   # ~mocap/tf
   use_pose: true  # ~mocap/pose
+  # select Optitrack Motive coordinates
+  use_motive_zxy: false # Set true if you use Motive
 
 # px4flow
 px4flow:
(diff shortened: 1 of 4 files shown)


vooon:
Not sure that i want this swap magick. May you write that as applying rotation?
Define rotation by `Eigen::AngleAxis` and then apply in to Quaternion and Point.

rosn:
I understand that I'll see what I can do I'm not familiar with Eigen lib so might take me some time.

tonybaltovski:
Just some thoughts.  Should this not be done prior to send the data to mavros?  It seems more reasonable to have the node interfacing with Motive to convert to the ROS standard.  The coordinate system recently changed in Motive 1.7.  In addition, shouldn't the pose be local for the mocap?

rosn:
The data are sent using VRPN technology so the problem still be the same, if Optitrack could change that it would be cool but they won't. the think is would you like to have full compatibility with Mocap tech or not. If not I'll use my own branch but I think it should be usefull for mavros to have these functionnalities.

rosn:
I got a message from optitrack, telling me the way to specify Z as the Up value for the ENU. Be careful this only apply the transformation to the data beeing sent. So you can close and reject this pull. Thank you for your attention.

PR #445, opened 2015-12-03
Updated local position subscription topic for setpoint demo

AndreasAntener:
@vooon Maybe we need to include this script somehow in a test so it doesn't break regularly..


vooon:
Not sure hot to do that. Thank you for fix.

Issue #470, opened 2016-01-18
Accessing body frame velocities


scott-eddy:
Hi all,
I am having trouble extracting the PX4 velocity estimate expressed in the body frame.  First let me start off by saying that I'm aware of #438  and think that this is a different problem.  So, looking at the local_position plugin we read in the FCU's estimate of velocity expressed in NED coordinates, as well as extract the vehicle's attitude WRT to the NED frame.  We then transform the velocity to be with respect to the NWU frame (as discussed in #438).  Additionally, the attitude quaternion is changed to represent the vehicle attitude WRT to the NWU frame. 

So, if my custom node does the inverse transformation of the reported velocity with the reported quaternion I will get the vehicle's velocity estimate in the NED frame.  Extrapolating further if I want the vehicle's velocity in the body frame I will need to rotate the quaternion to once again describe the vehicle's attitude WRT the NED frame and transform the NED veloicity with this new quaternion.  Is that correct? 

I don't think that having the vehicle's  body frame velocity is specific to my use case and could be handled within MAVROS as a nav_msgs/Odometry publisher.  This change would obviously require strict definitions for child and header frame ids, but I think it would be minimally impactful if a new frame called "body" was introduced.  Then nav_msgs/Odometry could be pose of the vehicle in the ROS frame (again NWU according to #438 and ENU according to documentation) and the velocity of the vehicle in the body frame.

scott-eddy:
Fixed in #473.  Closing

Issue #438, opened 2015-11-24
Coordinate Frame Mismatches


austinsteamboat:
So I still have an issue with all of this ENU NED business. I took the pixhawk out with a laptop and a GPS device, monitored the /mavros/local_position/local topic, and walked around. Taking the arrow on the pixhawk to be forward flight direction I walked North and saw the x-field of the local position estimate in Mavros increase. I walked East and saw the y-field decrease. I raised the laptop and saw the z-field increase. This indicates to me that XYZ maps to NWU, not NED, and not ENU. I've jumped through a lot of hoops to correct this throughout the code-base I'm building for quadrotor control. I guess I just think someone should indicate this better in the documentation or correct the documentation or correct it in the code.

vooon:
Coordinate frame transforms is much painful topic. Perhaps current one rule for all transforms is wrong, but i want detailed investigation and full list of message - frame used.

That much tied to firmware side, and because we support both PX4 and APM (Copter, Plane and Rover) that little difficult to support (seek & workaround) differences.

For most messages current rule (rotate X + 180°) is right,
while proper ENU-NED require rotate X +180° & Y +90°.

austinsteamboat:
I guess to be clear, it would be ENU if the pixhawk was mounted sideways. I'll be commanding set-points in a local xyz coordinate frame so from a standpoint of pure translation I'm not so concerned, but this could have an effect when I'm giving yaw setpoints. So maybe a different question would be:

1. Is 0 yaw aligned with true North? And is that North aligned with the arrow on the pixhawk's body or is it actually assuming we're mounted sideways? 

Thanks,
Austin

LorenzMeier:
@austinsteamboat Do you have a robotics / academic background? The reason I ask is that I'd like to drill down on this one and get it properly evaluated. I have followed the conversion conversations loosely but I feel I need to spend some quality time on this and help to get it the rigour it requires.

@tfoote If I walk in the earth-fixed ROS coordinate frame into the zero degree orientation, in which axis (X, Y, Z) direction am I walking? I just want to be sure there is no fundamental confusion on our end.

How do you put your ENU frame into GPS coordinates? Where is zero degrees (north or east?)?

austinsteamboat:
Hi Lorenz,
[...]

PR #473, opened 2016-01-18
WIP: Add odometry message to local position plugin & coordinate frame review

scott-eddy:
This PR will add a nav_msgs/Odometry output to the local position plugin, solving #470 directly.  Additionally #438 will be addressed by this PR with an in-depth analysis of the current coordinate frames used, transformations between frames, and making these frames as [REP 105](http://www.ros.org/reps/rep-0105.html) and [REP 103](http://www.ros.org/reps/rep-0103.html) compliant with minimal impact on the current user base (e.g. with changing as few frame names as possible).

Coordinate Frame Examination
======
- [x] Review frames currently in use
    - [x] Determine if renaming frames to be REP 105 compliant is worth the disruption to current user base
    - [x] Review rotations between current frames in [uas_frame_conversions.cpp]
(https://github.com/mavlink/mavros/blob/master/mavros/src/lib/uas_frame_conversions.cpp)
    - [x] Update frame naming and rotation methods as necessary

Odometry Message Publication
======
- [x] Publish Odometry message 
    - [x] Determine proper frames to be placed in header and child frame within the message
    - [x] Calculate linear and angular velocities in the associated child frame
    - [x] Determine way in which to provide the pose and twist covariances from the local position message

Documentation
======
[Here](https://drive.google.com/open?id=1bDhaozrUu9F915T58WGzZeOM-McyU20dwxX-NRum1KA) is a discussion of the proposed changes and why they are being made.

To help facilitate ensuring frames are properly rotated I've [begun tabulating](https://docs.google.com/spreadsheets/d/1LnsWTblU92J5_SMinTvBvHJWx6sqvzFa8SKbn8TXlnU/edit?usp=sharing) each plugin's data and what frame it is represented in on the FCU.  Not being and Arducopter developer I will not be diving into  frame representations on that firmware stack but have included it anyway if someone wants to take the lead on it.

### DIFF
--- a/mavros/include/mavros/mavros_uas.h
+++ b/mavros/include/mavros/mavros_uas.h
@@ -429,8 +429,8 @@ public:
 	/**
 	 * @brief Transform representation of attitude from 1 frame to another
 	 * (e.g. transfrom attitude from representing  from base_link -> NED
-	                 to representing base_link -> ENU)
-
+	 *               to representing base_link -> ENU)
+	 *
 	 * General function. Please use specialized enu-ned and ned-enu variants.
 	 */
 	static Eigen::Quaterniond transform_orientation(const Eigen::Quaterniond &q, const STATIC_TRANSFORM transform);
@@ -475,7 +475,7 @@ public:
 	 */
 	template<class T>
 	static inline T transform_orientation_ned_enu(const T &in) {
-		return transform_orientation(in,STATIC_TRANSFORM::NED_TO_ENU);
+		return transform_orientation(in, STATIC_TRANSFORM::NED_TO_ENU);
 	}
 
 	/**
@@ -484,7 +484,7 @@ public:
 	 */
 	template<class T>
 	static inline T transform_orientation_enu_ned(const T &in) {
-		return transform_orientation(in,STATIC_TRANSFORM::ENU_TO_NED);
+		return transform_orientation(in, STATIC_TRANSFORM::ENU_TO_NED);
 	}
 
 	/**
@@ -493,7 +493,7 @@ public:
 	 */
 	template<class T>
 	static inline T transform_orientation_aircraft_baselink(const T &in) {
-		return transform_orientation(in,STATIC_TRANSFORM::AIRCRAFT_TO_BASELINK);
+		return transform_orientation(in, STATIC_TRANSFORM::AIRCRAFT_TO_BASELINK);
 	}
 
 	/**
@@ -502,7 +502,7 @@ public:
 	 */
 	template<class T>
 	static inline T transform_orientation_baselink_aircraft(const T &in) {
-		return transform_orientation(in,STATIC_TRANSFORM::BASELINK_TO_AIRCRAFT);
+		return transform_orientation(in, STATIC_TRANSFORM::BASELINK_TO_AIRCRAFT);
 	}
 
 	/**
@@ -511,7 +511,7 @@ public:
 	 */
 	template<class T>
 	static inline T transform_frame_ned_enu(const T &in) {
-		return transform_static_frame(in,STATIC_TRANSFORM::NED_TO_ENU);
+		return transform_static_frame(in, STATIC_TRANSFORM::NED_TO_ENU);
 	}
 
 	/**
@@ -520,7 +520,7 @@ public:
 	 */
 	template<class T>
 	static inline T transform_frame_enu_ned(const T &in) {
-		return transform_static_frame(in,STATIC_TRANSFORM::ENU_TO_NED);
+		return transform_static_frame(in, STATIC_TRANSFORM::ENU_TO_NED);
 	}
 
 	/**
@@ -529,7 +529,7 @@ public:
 	 */
 	template<class T>
 	static inline T transform_frame_aircraft_baselink(const T &in) {
-		return transform_static_frame(in,STATIC_TRANSFORM::AIRCRAFT_TO_BASELINK);
+		return transform_static_frame(in, STATIC_TRANSFORM::AIRCRAFT_TO_BASELINK);
 	}
 
 	/**
@@ -538,7 +538,7 @@ public:
 	 */
 	template<class T>
 	static inline T transform_frame_baselink_aircraft(const T &in) {
-		return transform_static_frame(in,STATIC_TRANSFORM::BASELINK_TO_AIRCRAFT);
+		return transform_static_frame(in, STATIC_TRANSFORM::BASELINK_TO_AIRCRAFT);
 	}
 
 	/**
@@ -547,7 +547,7 @@ public:
 	 */
 	template<class T>
 	static inline T transform_frame_aircraft_ned(const T &in,const Eigen::Quaterniond &q) {
-		return transform_frame(in,q);
+		return transform_frame(in, q);
 	}
 
 	/**
@@ -556,7 +556,7 @@ public:
 	 */
 	template<class T>
 	static inline T transform_frame_ned_aircraft(const T &in,const Eigen::Quaterniond &q) {
-		return transform_frame(in,q);
+		return transform_frame(in, q);
 	}
 
 	/**
@@ -565,7 +565,7 @@ public:
 	 */
 	template<class T>
 	static inline T transform_frame_aircraft_enu(const T &in,const Eigen::Quaterniond &q) {
-		return transform_frame(in,q);
+		return transform_frame(in, q);
 	}
 
 	/**
@@ -574,7 +574,7 @@ public:
 	 */
 	template<class T>
 	static inline T transform_frame_enu_aircraft(const T &in,const Eigen::Quaterniond &q) {
-		return transform_frame(in,q);
+		return transform_frame(in, q);
 	}
 
 	/**
@@ -583,7 +583,7 @@ public:
 	 */
 	template<class T>
 	static inline T transform_frame_enu_baselink(const T &in,const Eigen::Quaterniond &q) {
-		return transform_frame(in,q);
+		return transform_frame(in, q);
 	}
 
 	/**
@@ -592,7 +592,7 @@ public:
 	 */
 	template<class T>
 	static inline T transform_frame_baselink_enu(const T &in,const Eigen::Quaterniond &q) {
-		return transform_frame(in,q);
+		return transform_frame(in, q);
 	}
 
 	/**
--- a/mavros/include/mavros/mavros_uas.h
+++ b/mavros/include/mavros/mavros_uas.h
@@ -3,7 +3,7 @@
  * @file mavros_uas.h
  * @author Vladimir Ermakov <vooon341@gmail.com>
  * @author Eddy Scott <scott.edward@aurora.aero>
- * 
+ *
  * @addtogroup nodelib
  * @{
  */
@@ -419,17 +419,17 @@ public:
 	/**
 	 * @brief Orientation transform options when applying rotations to data
 	 */
-	enum class STATIC_TRANSFORM: uint8_t {
-		NED_TO_ENU, //! will change orinetation from being expressed WRT NED frame to WRT ENU frame
-		ENU_TO_NED,  //! change from expressed WRT ENU frame to WRT NED frame
-		AIRCRAFT_TO_BASELINK, //! change from expressed WRT aircraft frame to WRT to baselink frame
-		BASELINK_TO_AIRCRAFT //! change from expressed WRT baselnk to WRT aircraft
+	enum class STATIC_TRANSFORM : uint8_t {
+		NED_TO_ENU,	//!< will change orinetation from being expressed WRT NED frame to WRT ENU frame
+		ENU_TO_NED,	//!< change from expressed WRT ENU frame to WRT NED frame
+		AIRCRAFT_TO_BASELINK,	//!< change from expressed WRT aircraft frame to WRT to baselink frame
+		BASELINK_TO_AIRCRAFT	//!< change from expressed WRT baselnk to WRT aircraft
 	};
 
 	/**
 	 * @brief Transform representation of attitude from 1 frame to another
 	 * (e.g. transfrom attitude from representing  from base_link -> NED
-	 		 to representing base_link -> ENU)
+	                 to representing base_link -> ENU)
 
 	 * General function. Please use specialized enu-ned and ned-enu variants.
 	 */
@@ -458,7 +458,7 @@ public:
 	 * General function. Please use specialized enu-ned and ned-enu variants.
 	 */
 	static Eigen::Vector3d transform_static_frame(const Eigen::Vector3d &vec, const STATIC_TRANSFORM transform);
-	
+
 	/**
 	 * @brief Transform convariance expressed in one frame to another
 	 *
--- a/mavros/include/mavros/mavros_uas.h
+++ b/mavros/include/mavros/mavros_uas.h
@@ -2,7 +2,8 @@
  * @brief MAVROS Plugin context
  * @file mavros_uas.h
  * @author Vladimir Ermakov <vooon341@gmail.com>
- *
+ * @author Eddy Scott <scott.edward@aurora.aero>
+ * 
  * @addtogroup nodelib
  * @{
  */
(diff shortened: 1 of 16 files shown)


scott-eddy:
To address the first point of this I have created the following [document](https://docs.google.com/document/d/1bDhaozrUu9F915T58WGzZeOM-McyU20dwxX-NRum1KA/edit?usp=sharing).

@vooon, @LorenzMeier, @mhkabir.  Input would be appreciated

andre-nguyen:
https://github.com/mavlink/mavros/blob/master/mavros/src/lib/uas_frame_conversions.cpp#L24

Would it not be sufficient to just change

```C++
static const Eigen::Quaterniond FRAME_ROTATE_Q = UAS::quaternion_from_rpy(M_PI, 0.0, 0.0);
```
to 
```C++
static const Eigen::Quaterniond FRAME_ROTATE_Q = UAS::quaternion_from_rpy(M_PI, 0.0, M_PI_2);
```

?

scott-eddy:
@andre-nguyen this change would be sufficient for correcting the change from the ENU to NED frame.  However, the reluctance to change this is because some messages (e.g. ones for IMU data) are actually expressed in the body frame *not* the NED frame as @mhkabir pointed out in #438.  

If nothing else we need to have a body frame -> ENU frame and NED -> ENU frame conversion.  But by making the FRAME_ROTATE_Q variable an enum we will be able to define an arbitrary number of useful rotations such as NED -> body frame or ENU -> custom frame.  Also by making the transform depend on an enumeration if a specific transform was found to have a bug in it, then the entire API wouldn't need to be re-worked, just that specific rotations.

andre-nguyen:
Good point

scott-eddy:
One limitation I'm running into is that FRAME_ROTATE_Q by nature has to be a static rotation.  This makes the NED -> body frame transformation not fit into the scope of the frame transformation lib as the quaternion from NED -> body, and for that matter ENU -> body, will always be changing.  

I don't know how best to approach that piece, I think that the lib should remain a library of strict frame to frame definitions and that each plugin can convert things into the body frame as necessary.

mavros/src/lib/uas_frame_conversions.cpp
```diff
@@ -21,24 +21,75 @@ using namespace mavros;
 // Eigen based functions
 
 //! +PI rotation around X (Roll) axis give us ROS or FCU representation
-static const Eigen::Quaterniond FRAME_ROTATE_Q = UAS::quaternion_from_rpy(M_PI, 0.0, 0.0);
+static Eigen::Quaterniond FRAME_ROTATE_Q;
 
 //! Transform for vector3
-static const Eigen::Transform<double, 3, Eigen::Affine> FRAME_TRANSFORM_VECTOR3(FRAME_ROTATE_Q);
+//static const Eigen::Transform<double, 3, Eigen::Affine> FRAME_TRANSFORM_VECTOR3(FRAME_ROTATE_Q);
```
vooon: It is ok for research, but later i think better to make two consts: `FRAME_ROTATE_LOCAL_Q`, `FRAME_ROTATE_PLATFORM_Q`.

Also `platform`? Not `base`?

mavros/include/mavros/mavros_uas.h
```diff
@@ -416,35 +416,45 @@ class UAS {
 	}
 
 	/**
+	 * @brief Frame transform options when applying rotations to data
+	 */
+	enum TRANSFORM_TYPE{
```
vooon: Maybe new feature:
```cpp
enum class TransformType: uint8_t {
...
}
```

mavros/include/mavros/mavros_uas.h
```diff
@@ -416,35 +416,45 @@ class UAS {
...
+		ENU_TO_NED
+	};
+
+	/**
 	 * @brief Transform frame between ROS and FCU. (Vector3d)
 	 *
 	 * General function. Please use specialized enu-ned and ned-enu variants.
 	 */
-	static Eigen::Vector3d transform_frame(const Eigen::Vector3d &vec);
+	static Eigen::Vector3d transform_frame(const Eigen::Vector3d &vec, const TRANSFORM_TYPE &transform);
```
vooon: "By reference" not needed for small types like enum: `const TransformType type`.

mavros/src/plugins/imu_pub.cpp
```diff
@@ -203,9 +203,9 @@ class IMUPubPlugin : public MavRosPlugin {
 		mavlink_attitude_t att;
 		mavlink_msg_attitude_decode(msg, &att);
 
-		auto orientation = UAS::transform_frame_ned_enu(
+		auto orientation = UAS::transform_frame_platform_enu(
 				UAS::quaternion_from_rpy(att.roll, att.pitch, att.yaw));
-		auto gyro = UAS::transform_frame_ned_enu(
+		auto gyro = UAS::transform_frame_platform_enu(
```
vooon: I'm not sure about naming: we transform from FCU body to ROS body frame.

mavros/src/plugins/local_position.cpp
```diff
@@ -63,36 +69,60 @@ class LocalPositionPlugin : public MavRosPlugin {
...
 		auto angular_velocity = uas->get_attitude_angular_velocity();
 
+		// Here the orientation quaternion is describing the rotation from the body frame 
+		// to the ENU frame.  We also have the velocity expressed in the ENU frame.  We
+		// Therefore need to calculate the inverse quaternion to find the transformation
+		// from the ENU frame to the body frame.  Then we can apply this rotation to the 
+		// velocity expressed in the ENU frame to get the body frame velocity.  
+		auto enu_to_body_quad = Eigen::Quaterniond(orientation.w, orientation.x, orientation.y, orientation.z).inverse();
+		auto enu_angular_velocity = Eigen::Vector3d(angular_velocity.x, angular_velocity.y, angular_velocity.z);
+		Eigen::Transform<double, 3, Eigen::Affine> enu_to_body_rot(enu_to_body_quad);
```
vooon: I think `Eigen::Affine3d` is the same type.
Also you may use `eigen_conversions/eigen_msg.h` for `enu_*`.

mavros/src/plugins/local_position.cpp
```diff
@@ -101,7 +131,7 @@ class LocalPositionPlugin : public MavRosPlugin {
 			transform.child_frame_id = tf_child_frame_id;
 
 			transform.transform.rotation = orientation;
```
vooon: Check please.

mavros/src/plugins/imu_pub.cpp
scott-eddy: This still confuses me.  The only time that transforming data expressed in the body-fixed frame into the ENU frame with a roll of 180 degrees is correct is in when the body-fixed frame's X axis is aligned with East.  

In fact, data that is **truly** expressed in the body-fixed frame should never be rotated by a static quaternion, it should be rotated by the  attitude calculated by the FCU.  Since the attitude represents the dynamic transform from the body-fixed frame to a fixed inertial frame (e.g. NED or ENU), the attitude should be rotated by a static transform to facilitate different frame expressions. 

For example consider the case where the acceleration data is given in the body-fixed frame, and the FCU computes the attitude of the body-fixed frame with respect to the NED frame.  If the user wants the acceleration data reported in the ENU frame they need to rotate the body-fixed frame quaternion to now express the attitude of the vehicle with respect to the ENU frame.  They then need to transform the acceleration data with this new quaternion.  

I think that this should be the fundamental way transformations from the body-fixed frame should work.  If data is already expressed in the NED frame (e.g. the velocities reported by the local_position plugin) then a static transform should be applied.    

Am I wrong here?

mhkabir: You're absolutely correct actually.

I came to the conclusion that a static rotation matrix isn't quite right.
Your approach seems correct to me.

scott-eddy:
In light of the need of dynamic rotations we should decide if we want all rotations to be handled by tf2 within mavros or if we want to continue implementing custom transforms.  Please see new proposed changes in the [documentation](https://drive.google.com/open?id=1bDhaozrUu9F915T58WGzZeOM-McyU20dwxX-NRum1KA)

vooon:
Check [REP103][1], at least IMU should be reported in ros-body frame. And for that static rotation around Roll (X) is good enough.
Also real local-ned to local-enu same static transform. So only "ned-body" or "fcu-body" frame to "local-enu" require dynamic transform.

And here we may use TF, define several transforms:

Static:
- `fcu_local` → `ros_local`: ROLL +180° YAW +90°
- `fcu_body` → `ros_body`: ROLL +180°

Dynamic:
- `fcu_body` → `ros_local`

Then later anyone may check applied conversions in TF tree. But that approach cost in CPU usage than internal Eigen calculations.

mhkabir:
@scott-eddy Looks to me that @vooon suggestion is good  :)

scott-eddy:
Sounds good to add the rotation from fcu_body to ros_body.  Today I'll walk through and clean up all of the naming conventions and try to document it somewhere for clarity.

I'm still a little hazy on the desired implementation.  Are we going to:
1. Use tf internally to perform all transformations.
2. Use Eigen internally for transforms, to save CPU, and only report transformations in the tf tree when setting tf_send to true?

2 would be more in line with how things are done now, and since CPU is generally limited on companion computer's I would vote we do it that way.

vooon:
`tf_send` param not needed - static transform sent only once.

Perhaps 2-option is more effecient (since we do not need to resolve TF on each iteration), but if you can please add switch for static/tf selection.
I don't think that it is hard since we should use same code for body-local tf.

scott-eddy:
@vooon see what you think about that latest push.  It includes a transform from aircraft (which is the fcu body-fixed frame) to base_link (which is the ros body-fixed frame).  It also corrects the IMU publication.

mhkabir:
With your latest push, what's the "aircraft" frame and the "base_link" frame referring to?

scott-eddy:
@mhkabir base_link is the body-fixed frame expressed in ROS frame as per [REP 105](http://www.ros.org/reps/rep-0105.html) and [REP 103](http://www.ros.org/reps/rep-0103.html).  The "aircraft" frame is the [PX4 body-fixed frame](https://pixhawk.org/dev/know-how/frames_of_reference).  

I wanted to avoid calling one frame body and one base_link as both are rigidly attached to the vehicle.  "aircraft" was chosen because the body-fixed frame for aircraft is usually defined with Z Down.  Open to other naming conventions but it seems to make sense.  

Obviously when all is said and done documentation needs to be developed, we should decide where to house it and how to inform other programs (e.g. PX4, Arducopter)

scott-eddy:
Testing with 4a591db looks good.  I ran some tests with the aircraft on the bench with the following:
* Start with aircraft on wheels (e.g. roll, pitch ==0)
* rotate to right wing down
* rotate to left wing down
* rotate to wheels
* rotate to nose down
* rotate to nose up.
* rotate to wheels

This yields the following acceleration profile:
1-22_base_link_accels


While not intuitive, this is the correct rotation of the reported accelerations.  I would have expected that when on wheels, one g would show up in the base_link frame as -9.8[m/s/s] as the g is aligned with the down direction.  However, as explained to me in [PX4/Firmware #3551](https://github.com/PX4/Firmware/issues/3551) this is not the correct way to think about accelerometer measurements.  That said, the ROS sensor_msgs/IMU wants linear accelerations in the base_link frame.  So, if g is aligned with an axis do we report it as a negative or positive g (in m/s/s of course).  @voon, @mhkabir thoughts?

The bench test also provided the following angular rate profile:
1-22_base_link_angular_rates

These rates are correct for the base_link frame:
* positive x rate from wheels to right wing down, negative rate from right wing down to left wing, positive to return to wheels
* positive y rate from wheels to nose down, negative from nose down to nose up, positive to return to wheels

scott-eddy:
Still some problem with extracting velocity in base_link frame.  Here I moved the vehicle back in forth in the base_link x axis and rotated around the z.  Resulting velocities should be a sine wave in the x axis only but there is clearly a y component over time.  

1-22_body_velocities

scott-eddy:
Change didn't fix it
1-22_no_inverse

mhkabir:
@scott-eddy Could you check the following accelerations quickly : 

* Measure +9.81 meters per second squared for the Z axis.
* If the FCU is rolled +90 degrees (left side up), the acceleration should be +9.81 meters per second squared for the Y axis.
* If the FCU is pitched +90 degrees (front side down), it should read -9.81 meters per second squared for the X axis.

scott-eddy:
@mhkabir those are the accelerations measured with this.  See above plot.

scott-eddy:
Tested today that the orientation transformation from NED->aircraft to ENU->base_link is correct.  After that I confirmed that rotating the reported NED velocities to the base_link frame results in accurate body-frame velocities.  I shook the vehicle in the x axis while rotating it about the Z, then shook in the Y while rotating back the other way about the Z.  I repeated this as follows:
* body X velocity
* body Y velocity
* body X velocity 
* body Y velocity
* body X velocity
The below plot confirms the rotations working as the body x and y velocities are nicely split.  
1-25_seperated_body_vel

scott-eddy:
@vooon @mhkabir I'm confident the rotations are all working correctly now.  How do we get this merged in?  I can open a new PR with changes rebased to master.  Also, I don't have a mocap facility to test some of the modules there.  

What other testing needs to be performed?

mavros/launch/px4.launch
```diff
@@ -2,7 +2,7 @@
 	<!-- vim: set ft=xml noet : -->
 	<!-- example launch script for PX4 based FCU's -->
 
-	<arg name="fcu_url" default="/dev/ttyACM0:57600" />
+	<arg name="fcu_url" default="/dev/ttyTHS1:921600" />
```
mhkabir: Local change got in.

mavros/launch/px4_pluginlists.yaml
```diff
@@ -4,7 +4,7 @@ plugin_blacklist:
 # extras
 - image_pub
 - vibration
-- distance_sensor
+# - distance_sensor
```
mhkabir: Local change

mavros/src/plugins/imu_pub.cpp
```diff
@@ -66,11 +69,16 @@ class IMUPubPlugin : public MavRosPlugin {
...
+		// we rotate the data from the aircraft-frame to the base_link frame.
+		// Additionally we report the orientation of the vehicle to describe the 
+		// transformation from the ENU frame to the base_link frame (ENU <-> base_link).  
+		// THIS ORIENTATION IS NOT THE SAME AS THAT REPORTED BY THE FCU (NED <-> aircraft)  
+		imu_nh.param<std::string>("frame_id", frame_id, "base_link"); 
 		imu_nh.param("linear_acceleration_stdev", linear_stdev, 0.0003);		// check default by MPU6000 spec
 		imu_nh.param("angular_velocity_stdev", angular_stdev, 0.02 * (M_PI / 180.0));	// check default by MPU6000 spec
 		imu_nh.param("orientation_stdev", orientation_stdev, 1.0);
 		imu_nh.param("magnetic_stdev", mag_stdev, 0.0);
+		imu_nh.param("transform_debug",transform_debug,false);
```
mhkabir: Maybe remove this for release?

mavros/src/plugins/local_position.cpp
```diff
@@ -2,7 +2,8 @@
  * @brief LocalPosition plugin
  * @file local_position.cpp
  * @author Vladimir Ermakov <vooon341@gmail.com>
- * @author Glenn Gregory
+ * @author Glenn
```
mhkabir: Unintended change, looks like

mavros/src/plugins/local_position.cpp
```diff
@@ -100,11 +136,45 @@ class LocalPositionPlugin : public MavRosPlugin {
...
 			transform.child_frame_id = tf_child_frame_id;
 
-			transform.transform.rotation = orientation;
-			tf::vectorEigenToMsg(position, transform.transform.translation);
+			transform.transform.rotation = enu_orientation_msg;
+			tf::vectorEigenToMsg(enu_position, transform.transform.translation);
 
 			uas->tf2_broadcaster.sendTransform(transform);
 		}
+		if (tf_debug){
```
mhkabir: Maybe remove this for release and make it a unit test instead?

mavros/src/plugins/setpoint_attitude.cpp
```diff
@@ -50,8 +50,8 @@ class SetpointAttitudePlugin : public MavRosPlugin,
 		sp_nh.param("reverse_throttle", reverse_throttle, false);
 		// tf params
 		sp_nh.param("tf/listen", tf_listen, false);
-		sp_nh.param<std::string>("tf/frame_id", tf_frame_id, "local_origin");
-		sp_nh.param<std::string>("tf/child_frame_id", tf_child_frame_id, "attitude");
+		sp_nh.param<std::string>("tf/frame_id", tf_frame_id, "map");
```
mhkabir: Please check. It seems to me that this should be world.

mavros/src/plugins/setpoint_position.cpp
```diff
@@ -45,8 +45,8 @@ class SetpointPositionPlugin : public MavRosPlugin,
 
 		// tf params
 		sp_nh.param("tf/listen", tf_listen, false);
-		sp_nh.param<std::string>("tf/frame_id", tf_frame_id, "local_origin");
-		sp_nh.param<std::string>("tf/child_frame_id", tf_child_frame_id, "setpoint");
+		sp_nh.param<std::string>("tf/frame_id", tf_frame_id, "map");
```
mhkabir: Same as above.

mavros_extras/src/plugins/vision_speed_estimate.cpp
```diff
@@ -84,6 +84,7 @@ class VisionSpeedEstimatePlugin : public MavRosPlugin {
 	void send_vision_speed(const geometry_msgs::Vector3 &vel_enu, const ros::Time &stamp) {
 		Eigen::Vector3d vel_;
 		tf::vectorMsgToEigen(vel_enu, vel_);
+		//Transform from ENU to NED frame
 		auto vel = UAS::transform_frame_enu_ned(vel_);
```
mhkabir: No changes in this plugin?

mhkabir:
@scott-eddy I just did a review, and it's looking great overall! Just a couple of nitpicks :) Also, please rebase to master. 

I should be able to test this with vision or laser odometry in a couple of days time.

scott-eddy:
@mhkabir thanks for reviewing this.  I'll rebase this tomorrow and address the comments above.

mavros_extras/src/plugins/vision_speed_estimate.cpp
scott-eddy: No change here.  The enu->ned transform is correct and that function call's API was unchanged.

mavros/src/plugins/setpoint_attitude.cpp
scott-eddy: "map" is the correct naming, "world" was incorrect in the local_position plugin

mavros/src/plugins/local_position.cpp
scott-eddy: removed

scott-eddy:
@vooon @mhkabir this is rebased to master and ready for final testing and merging.

mavros/include/mavros/mavros_uas.h
```diff
@@ -416,43 +417,182 @@ class UAS {
...
+		AIRCRAFT_TO_BASELINK, //! change from expressed WRT aircraft frame to WRT to baselink frame
+		BASELINK_TO_AIRCRAFT //! change from expressed WRT baselnk to WRT aircraft
+	};
+
+	/**
+	 * @brief Transform representation of attitude from 1 frame to another
+	 * (e.g. transfrom attitude from representing  from base_link -> NED
+	 		 to representing base_link -> ENU)
+
+	 * General function. Please use specialized enu-ned and ned-enu variants.
```
vooon: Please add `@param[in]` or `@param[out]` and `@returns` doxygen tags.

mavros/include/mavros/mavros_uas.h
```diff
@@ -416,43 +417,182 @@ class UAS {
...
 
 	/**
-	 * @brief Transform frame between ROS and FCU. (Vector3d)
+	 * @brief Orientation transform options when applying rotations to data
+	 */
+	enum class STATIC_TRANSFORM: uint8_t {
+		NED_TO_ENU, //! will change orinetation from being expressed WRT NED frame to WRT ENU frame
+		ENU_TO_NED,  //! change from expressed WRT ENU frame to WRT NED frame
+		AIRCRAFT_TO_BASELINK, //! change from expressed WRT aircraft frame to WRT to baselink frame
+		BASELINK_TO_AIRCRAFT //! change from expressed WRT baselnk to WRT aircraft
```
vooon: Inline doxygen should be `//!<`.

mavros/src/lib/uas_frame_conversions.cpp
```diff
@@ -18,26 +19,132 @@
...
+static const Eigen::Quaterniond AIRCRAFT_BASELINK_Q = UAS::quaternion_from_rpy(M_PI, 0.0, 0.0);
 
-//! +PI rotation around X (Roll) axis give us ROS or FCU representation
-static const Eigen::Quaterniond FRAME_ROTATE_Q = UAS::quaternion_from_rpy(M_PI, 0.0, 0.0);
 
 //! Transform for vector3
-static const Eigen::Transform<double, 3, Eigen::Affine> FRAME_TRANSFORM_VECTOR3(FRAME_ROTATE_Q);
+//static const Eigen::Transform<double, 3, Eigen::Affine> FRAME_TRANSFORM_VECTOR3(FRAME_ROTATE_Q);
+
+
```
vooon: Why not to add local, maybe inline function with STATIC_TRANSFORM switch?

```cpp
static const Eigen::Quaterniond &get_static_transform_q(const STATIC_TRANSFORM tr)
{
switch {
default:
    ROS_ASSERT(false);
}
}

File: mavros/src/lib/uas_frame_conversions.cpp [hunk 18]
```diff
@@ -18,26 +19,132 @@
...
+	//be seen http://www.euclideanspace.com/maths/algebra/realNormedAlgebra/quaternions/transforms/
+		switch(transform){
+		case STATIC_TRANSFORM::NED_TO_ENU:
+		case STATIC_TRANSFORM::ENU_TO_NED:{
+			return NED_ENU_Q * q;
+			break;
+		}
+		case STATIC_TRANSFORM::AIRCRAFT_TO_BASELINK:
+		case STATIC_TRANSFORM::BASELINK_TO_AIRCRAFT:{
+			return q * AIRCRAFT_BASELINK_Q;
```
vooon: You sure that in one case Q * q and in other q * Q?

mavros/src/lib/uas_frame_conversions.cpp
```diff
@@ -18,26 +19,132 @@
...
+			break;
+		}
+		default:{
+			//We don't know how to express the attitude WRT an undefined
```
vooon: Best what we can: quit. That is programming error, so `ROS_ASSERT(false)`.

mavros/src/lib/uas_frame_conversions.cpp
```diff
@@ -18,26 +19,132 @@
...
+		}
+	}
+}
+
+Eigen::Vector3d UAS::transform_static_frame(const Eigen::Vector3d &vec, const UAS::STATIC_TRANSFORM transform)
+{
+	switch(transform){
+		case STATIC_TRANSFORM::NED_TO_ENU:
+		case STATIC_TRANSFORM::ENU_TO_NED:{
+			Eigen::Affine3d FRAME_TRANSFORM_VECTOR3(NED_ENU_Q);
```
vooon: Similar to Q, add static const Affines and add selector helper.

mavros/src/plugins/imu_pub.cpp
```diff
@@ -203,12 +210,19 @@ class IMUPubPlugin : public MavRosPlugin {
...
 
-		auto orientation = UAS::transform_frame_ned_enu(
-				UAS::quaternion_from_rpy(att.roll, att.pitch, att.yaw));
-		auto gyro = UAS::transform_frame_ned_enu(
+		//Here we have rpy describing the rotation: aircraft->NED
+		//We need to change this to aircraft->ENU
+		//And finally change it to baselink->ENU
+		auto enu_baselink_orientation = UAS::transform_orientation_aircraft_baselink(
+				UAS::transform_orientation_ned_enu(
+				UAS::quaternion_from_rpy(att.roll, att.pitch, att.yaw)));
```
vooon: Looks little weird.

vooon:
I added few notes, but looks good.
And also please apply uncrustify.

vooon:
Unit test passed: https://travis-ci.org/mavlink/mavros/jobs/104938859#L4639

mavros/src/lib/uas_frame_conversions.cpp
scott-eddy: Yes, the [link in the comment](http://www.euclideanspace.com/maths/algebra/realNormedAlgebra/quaternions/transforms/) sheds some insight as to why.

I've also verified with extensive testing.

mhkabir:
@scott-eddy Could you just address those last few nitpicks? This is looking great! :) Almost ready for prime-time. I'm currently giving this a sprint with my VIO system.
 
Thanks!!

mhkabir:
Okay... We have a crash here, that I discovered while testing : 
```
/home/artemis/artemis_ws/devel/lib/mavros/mavros_node: symbol lookup error: /home/artemis/artemis_ws/devel/lib//libmavros_extras.so: undefined symbol: _ZN6mavros3UAS15transform_frameERKN5Eigen10QuaternionIdLi0EEE
```

tonybaltovski:
Related to `undefined symbol: mavros::UAS::transform_frame(Eigen::Quaternion<double, 0> const&)`

mavros/include/mavros/mavros_uas.h
scott-eddy: Add doxygen tags to each of the inline functions?  Even though they're missing now

scott-eddy:
@mhkabir @tonybaltovski It looks like I'm passing a quaternion to a transform_frame function somewhere instead of to the new transform_orientation.

scott-eddy:
@all,help would be appreciated on this one.  Its not obvious to me where this is happening.  What plugins are you running when you have experience the crash?  Can you try to nail it down to the one where the problem is arising from?

vooon:
@mhkabir did you tried clean all? Perhaps that come from old `*.o` files.
```
catkin clean --all
catkin build
```

vooon:
@scott-eddy replace `ROS_ASSERT(false)` with `ROS_BREAK()`: http://docs.ros.org/diamondback/api/rosconsole/html/assert_8h.html

mavros/include/mavros/mavros_uas.h
```diff
@@ -416,43 +417,182 @@ class UAS {
...
-	static Covariance6d transform_frame(const Covariance6d &cov);
+	static Covariance6d transform_static_frame(const Covariance6d &cov, const STATIC_TRANSFORM transform);
+
+	/**
+	 * @brief Transform from attitude represented WRT NED frame to attitude
+	 *		  represented WRT ENU frame
+	 */
+	template<class T>
+	static inline T transform_orientation_ned_enu(const T &in) {
+		return transform_orientation(in,STATIC_TRANSFORM::NED_TO_ENU);
```
vooon: Hmm, uncrustify should add space after `,`, but it don't.

mavros/include/mavros/mavros_uas.h
```diff
@@ -416,43 +417,182 @@ class UAS {
...
+		AIRCRAFT_TO_BASELINK,	//!< change from expressed WRT aircraft frame to WRT to baselink frame
+		BASELINK_TO_AIRCRAFT	//!< change from expressed WRT baselnk to WRT aircraft
+	};
+
+	/**
+	 * @brief Transform representation of attitude from 1 frame to another
+	 * (e.g. transfrom attitude from representing  from base_link -> NED
+	                 to representing base_link -> ENU)
+
+	 * General function. Please use specialized enu-ned and ned-enu variants.
```
vooon: Break of `*` "wall" looks ugly, but i may fix that later.

mavros/src/lib/uas_frame_conversions.cpp
```diff
@@ -18,26 +19,135 @@
...
+		return NED_ENU_AFFINE;
+		break;
+	}
+	case UAS::STATIC_TRANSFORM::AIRCRAFT_TO_BASELINK:
+	case UAS::STATIC_TRANSFORM::BASELINK_TO_AIRCRAFT: {
+		return AIRCRAFT_BASELINK_AFFINE;
+		break;
+	}
+	default: {
+		ROS_ASSERT(false);
```
vooon: I found that proper macro is `ROS_BREAK()`.

mavros/src/lib/uas_frame_conversions.cpp
```diff
@@ -18,26 +19,135 @@
...
+		//cov_out = FRAME_ROTATE_Q * cov_in * FRAME_ROTATE_Q.inverse();
+		// from comments on github about tf2_sensor_msgs.h
+		cov_out = cov_in * AIRCRAFT_BASELINK_Q;
+		return cov_out_;
+		break;
+	}
+	default: {
+		//We don't know the transform between the static frames.
+		//throw error and return given covariance
+		ROS_ERROR("Requested Static Frame Conversion Unkown");
```
vooon: `ROS_BREAK()`

mavros/src/lib/uas_frame_conversions.cpp
```diff
@@ -18,26 +19,135 @@
...
-Eigen::Quaterniond UAS::transform_frame(const Eigen::Quaterniond &q)
+UAS::Covariance3d UAS::transform_static_frame(const Covariance3d &cov, const UAS::STATIC_TRANSFORM transform)
 {
-	return FRAME_ROTATE_Q * q * FRAME_ROTATE_Q.inverse();
+	switch (transform) {
+	case STATIC_TRANSFORM::NED_TO_ENU:
+	case STATIC_TRANSFORM::ENU_TO_NED: {
+		Covariance3d cov_out_;
+		EigenMapConstCovariance3d cov_in(cov.data());
+		EigenMapCovariance3d cov_out(cov_out_.data());
```
vooon: Move map preparation out of switch - duplication of code.
Also i'm not sure that `{}` are necessary in that switches.

vooon:
Also check please `UAS::transform_frame_yaw()`.

mhkabir:
That would be imu_pub for me, I think. Or maybe vision_pose.

Only those two are active for me.

scott-eddy:
@mhkabir do you get the error after running catkin clean --all and rebuilding.  imu_pub doesn't produce errors on my end so if you can get it repeatably then it must be in vision_pose

mhkabir:
@scott-eddy Yep, vision_pose. I'm listening to a TF from my VIO system, note.

scott-eddy:
@mhkabir I'll look into this in more depth sometime soon, will ping you when I think I have found something.

@voon, I will address the comments you made above.

scott-eddy:
@mhkabir The only place that I can see an Eigen::Quaternion within vision pose is [here](https://github.com/scott-eddy/mavros/blob/base_link/mavros_extras/src/plugins/vision_pose_estimate.cpp#L118).  However, the function is transform_orientation which should be okay.

@tonybaltovski doesn't Eigen::Quaternion<double, 0> define a quaternion with dimension = 0?  Perhaps that contributes to the problem.

vooon:
CI failed. `ROS_BREAK();`, not `ROS_BREAK(false);`.
Also try to add `#include <ros/assert.h>`

scott-eddy:
Fixed ROS_BREAK, did not have to `#include <ros/asser.h>`

vooon:
@scott-eddy @mhkabir seems most problems are fixed, so what we need is testing?

vooon:
TODO: update README.

scott-eddy:
@vooon I still have not tracked down the problem that @mhkabir is having.  I think that is the lingering issue

scott-eddy:
@mhkabir I tested this with vision_pose running, sending a geometry_msgs/PoseStamped from rqt at a rate of 10Hz with no problem.  I can't seem to duplicate the bug you have seen

scott-eddy:
@mhkabir can you send me a gist of your config and pluginlist files

scott-eddy:
@vooon I just spoke with @mhkabir, he was able to trace his problem to old build files on his system.  We've both been able to run this with no issues.  It is good to go.

AlexisTM:
Dear, It seems that the problem is still up. 

I am sending a "almost null yaw" quaternion : `(x:0, y:0, z:0, w:1)` to /mavros/vision_pose/pose (with external heading to vision) and the resulting quaternion in /mavros/local/position is almost 180° `(x:0, y:0, z:-1, w:0)`

Issue #218, opened 2015-02-15
Support for multicast UDP bridge


vooon:
Planning now.

vooon:
Nope, i don't see that multicast is required. Instead i implemented broadcast discovery mode (`udp-b://`).

PR #477, opened 2016-01-21
Add allow_udp_broadcast option

i-v-s:
This adds `allow_udp_broadcast` option to mavros. This feature required, for example, to use qgroundcontrol on any device, connected to companion computer by wifi, without any tuning. Also, this fixes format string bug on startup of `mavros_node`.

### DIFF
--- a/libmavconn/include/mavconn/interface.h
+++ b/libmavconn/include/mavconn/interface.h
@@ -173,7 +173,7 @@ public:
 	 *         or throw @a DeviceError if error occured.
 	 */
 	static Ptr open_url(std::string url,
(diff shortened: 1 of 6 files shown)


vooon:
This could not work because before MAVConnUDP get remote address it will silent.
Next: i do not like idea to adding another parameter, it should be in URL. And finally better to implement multicast (#218), not broadcast.

i-v-s:
I don't know why, but it is working very good. I write address as `<param name="gcs_url" value="udp://:14556@192.168.2.255:14550" />`, and execute mavros on the companion computer. Companion computer also works as WiFi access point and DHCP server for the network 192.168.2.X. If you don't want to add another parameter, you can allow broadcast by default. When we set remote address as X.X.X.255, this automatically means that we will use broadcast.

PR #491, opened 2016-02-03
configurable vehicle model

chataign:
changes to mavros_extras/copter_visualization to make vehicle model configurable:
*  number of rotors
*  arm length
*  body width/height

(package now depends on TF for tf::createQuaternionMsgFromYaw but this could be removed)


vooon:
@chataign thank you for PR. Could you provide rviz screenshot?

chataign:
copter_visualization

chataign:
I updated the mavros_extras wiki to reflect the latest changes

PR #496, opened 2016-02-04
use MarkerArray for vehicle model

chataign:
Doing some tests I found that the rendering of the rotors in rviz can get a bit jittery. I think this is caused by sending out lots of markers all the time. I have now changed it to use MarkerArray instead and the rendering seems much nicer.

Would be good if someone could test to confirm.

Cheers!

### DIFF
--- a/mavros_extras/src/copter_visualization.cpp
+++ b/mavros_extras/src/copter_visualization.cpp
@@ -20,6 +20,9 @@
 
 #include <geometry_msgs/PoseStamped.h>
 #include <visualization_msgs/Marker.h>
+#include <visualization_msgs/MarkerArray.h>
+
+#include <boost/shared_ptr.hpp>
 
 // parameters
 static std::string fixed_frame_id;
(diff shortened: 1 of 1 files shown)


mhkabir:
Nice! I should be able to give this a try with my VIO setup tomorrow.

vooon:
That PR is conflicting with master. Please rebase to master and run again.
Also please do `uncrustify` as said in [readme](https://github.com/mavlink/mavros/tree/master/mavros#contributing).

mavros_extras/src/copter_visualization.cpp
```diff
@@ -20,6 +20,9 @@
 
 #include <geometry_msgs/PoseStamped.h>
 #include <visualization_msgs/Marker.h>
+#include <visualization_msgs/MarkerArray.h>
+
+#include <boost/shared_ptr.hpp>
```
vooon: Already included in `<ros/ros.h>`.

PR #499, opened 2016-02-05
Configurable vehicle model

chataign:
re-submitting pull request -- using MarkerArray instead of a series of Marker for vehicle model.

this is my first time contributing to mavros so please let me know if I'm still missing something.


vooon:
Still conflicting with master :(

PR #500, opened 2016-02-05
Configurable vehicle model

chataign:
re-re-submitting pull request -- using MarkerArray instead of a series of Marker for vehicle model.

hopefully I got the merging right this time!


vooon:
Thank you, @chataign.

PR #507, opened 2016-02-15
Treat submarine vehicles like copter vehicles

joshvillbrandt:
We are creating a new flavor of the APM software stack in conjunction with the drone code community called APM:Sub. The code is currently being tracked at [bluerobotics/ardupilot](https://github.com/bluerobotics/ardupilot/tree/master/ArduSub). At the moment, this is the only change required to get mavros to work on our Pixhawk + APM:Sub stack. We'd be very thankful if you can pull this into the official mavros repo and include it in the next release.

Thanks!

CC @rjehangir


vooon:
Thank you.

PR #509, opened 2016-02-16
added visualization for local setpoints

chataign:
listen to "local_setpoint" topic (should be remapped to /mavros/setpoint_position/local) and display setpoint position as arrow using rviz Marker on "wp_markers" topic

waypoint2

### DIFF
--- a/mavros_extras/src/copter_visualization.cpp
+++ b/mavros_extras/src/copter_visualization.cpp
@@ -30,6 +30,7 @@ static double marker_scale;
 // merker publishers
 ros::Publisher track_marker_pub;
 ros::Publisher vehicle_marker_pub;
+ros::Publisher wp_marker_pub;
 
 boost::shared_ptr<visualization_msgs::MarkerArray> vehicle_marker;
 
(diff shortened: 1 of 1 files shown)


mhkabir:
Can we have a 'ghost'  transparent version of the model install of the arrow?

chataign:
We could but I think that would add unnecessary rendering which could eventually slow down rviz. Also essential information such as orientation wouldn't be as easy to see as with an arrow. Ultimately it might get a bit confusing if you have many markers / vehicle is moving fast / etc.

mhkabir:
Alright then. Accepted. :) 
Would you mind sending in a PR to PX4/Devguide with information on how to setup the whole interface in the current state? Thanks!

chataign:
Okay I'll try to get the PX4 dev guide updated.

Also I wanted to update the mavros wiki for copter_visualization but I don't know how to do it?

vooon:
Thank you!

To update wiki you need to register and then ask permission https://github.com/ros-infrastructure/roswiki/issues/139 (damn spamers!).
Then please use text editor and not WYSIWYG - it ruine node/plugin templated descriptions.

chataign:
@vooon What do I need to register to in order to edit the wiki? Sorry I'm a complete newbie :)

chataign:
@vooon ignore my last post, I worked it out

PR #514, opened 2016-02-29
Plugin for manual control using sensor_msgs::Joy messages

chataign:
I created a plugin to generate Mavlink manual_control messages from sensor_msgs::Joy input. It is based on the ROS node in px4/src/platforms/ros/nodes/manual_input

Mappings between buttons and the corresponding modes can be configured, as well as the button used to arm/disarm. The plugin has to handle modes & arming because PX4 seems to ignore the buttons flags in the manual_control message (see MavlinkReceiver::handle_message_manual_control) Ideally we would simply forward the button flags via Mavlink and let PX4 handle them.

Maybe this code is too specific, and not generic enough to make sense as a plugin, in which case it shouldn't be merged into master

### DIFF
--- a/mavros_extras/mavros_plugins.xml
+++ b/mavros_extras/mavros_plugins.xml
@@ -23,4 +23,8 @@
         <class name="cam_imu_sync" type="mavplugin::CamIMUSyncPlugin" base_class_type="mavplugin::MavRosPlugin">
 		<description>Publish camera trigger data for synchronisation of IMU and camera frames</description>
 	</class>
+	</class>
+        <class name="joy_control" type="mavplugin::JoyControlPlugin" base_class_type="mavplugin::MavRosPlugin">
+		<description>Listen to sensor_msgs::Joy and convert into Mavlink manual control messages</description>
+	</class>
 </library>
--- a/mavros_extras/mavros_plugins.xml
+++ b/mavros_extras/mavros_plugins.xml
@@ -20,7 +20,7 @@
 	<class name="vibration" type="mavplugin::VibrationPlugin" base_class_type="mavplugin::MavRosPlugin">
 		<description>Publish VIBRATION message data from FCU.</description>
 	</class>
-		<class name="cam_imu_sync" type="mavplugin::CamIMUSyncPlugin" base_class_type="mavplugin::MavRosPlugin">
+	<class name="cam_imu_sync" type="mavplugin::CamIMUSyncPlugin" base_class_type="mavplugin::MavRosPlugin">
 		<description>Publish camera trigger data for synchronisation of IMU and camera frames</description>
 	</class>
 	<class name="joy_control" type="mavplugin::JoyControlPlugin" base_class_type="mavplugin::MavRosPlugin">
--- a/mavros_extras/mavros_plugins.xml
+++ b/mavros_extras/mavros_plugins.xml
@@ -23,8 +23,7 @@
         <class name="cam_imu_sync" type="mavplugin::CamIMUSyncPlugin" base_class_type="mavplugin::MavRosPlugin">
 		<description>Publish camera trigger data for synchronisation of IMU and camera frames</description>
 	</class>
-	</class>
-        <class name="joy_control" type="mavplugin::JoyControlPlugin" base_class_type="mavplugin::MavRosPlugin">
+    <class name="joy_control" type="mavplugin::JoyControlPlugin" base_class_type="mavplugin::MavRosPlugin">
 		<description>Listen to sensor_msgs::Joy and convert into Mavlink manual control messages</description>
 	</class>
 </library>
--- a/mavros_extras/mavros_plugins.xml
+++ b/mavros_extras/mavros_plugins.xml
@@ -20,10 +20,10 @@
 	<class name="vibration" type="mavplugin::VibrationPlugin" base_class_type="mavplugin::MavRosPlugin">
 		<description>Publish VIBRATION message data from FCU.</description>
 	</class>
-        <class name="cam_imu_sync" type="mavplugin::CamIMUSyncPlugin" base_class_type="mavplugin::MavRosPlugin">
+		<class name="cam_imu_sync" type="mavplugin::CamIMUSyncPlugin" base_class_type="mavplugin::MavRosPlugin">
 		<description>Publish camera trigger data for synchronisation of IMU and camera frames</description>
 	</class>
-    <class name="joy_control" type="mavplugin::JoyControlPlugin" base_class_type="mavplugin::MavRosPlugin">
+	<class name="joy_control" type="mavplugin::JoyControlPlugin" base_class_type="mavplugin::MavRosPlugin">
 		<description>Listen to sensor_msgs::Joy and convert into Mavlink manual control messages</description>
 	</class>
 </library>
--- a/mavros_extras/src/plugins/joy_control.cpp
+++ b/mavros_extras/src/plugins/joy_control.cpp
@@ -24,13 +24,14 @@
 #include <mavros_msgs/State.h>
 #include <mavros_msgs/SetMode.h>
 
-using namespace std;
-
 ////////////////////////////////////////////////////////////////////////////////
 ////////////////////////////////////////////////////////////////////////////////
 
 namespace mavplugin {
-struct Control
+/**
+ * @brief Properties for joystick control axis
+ */
+struct JoyControl
 {
 	int axis;
 	double scale;
@@ -42,6 +43,17 @@ struct Control
 		if ( axis < 0 || axis >= joy->axes.size() ) throw std::out_of_range("failed to map axis");
 		double val = joy->axes[axis] + offset;
 		return ( val > deadzone || val < -deadzone ? val * scale : 0.0 );
+	}
+
+	void load_params( ros::NodeHandle &priv_nh, const std::string& name,
+			int def_axis = 0, double def_scale = 1, double def_offset = 0, double def_deadzone = 0 )
+	{
+		auto nh = ros::NodeHandle( priv_nh, name );
+
+		nh.param( "axis",		axis,		def_axis		);
+		nh.param( "scale",		scale,		def_scale		);
+		nh.param( "offset",		offset,		def_offset		);
+		nh.param( "deadzone",	deadzone,	def_deadzone	);
 	}
 };
 
@@ -55,16 +67,14 @@ class JoyControlPlugin : public MavRosPlugin
 {
 	ros::NodeHandle mavros_nh_, plugin_nh_;
 	UAS *uas_;
-	mavros_msgs::State state_;
-	mavlink_message_t ctrl_msg_;
+	mavros_msgs::State::ConstPtr curr_state_;
 
-	Control x_, y_, z_, r_;
+	JoyControl x_, y_, z_, r_;
 
 	int arm_button_;
-	vector<int> mode_mapping_;
-	vector<string> mode_names_;
+	std::vector<int> mode_mapping_;
+	std::vector<std::string> mode_names_;
 
-	ros::Timer send_msg_timer_;
 	ros::ServiceClient arming_srv_, set_mode_srv_;
 	ros::Subscriber joystick_sub_, state_sub_;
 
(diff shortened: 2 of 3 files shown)


mavros_extras/src/plugins/joy_control.cpp
```diff
@@ -0,0 +1,266 @@
+/**
+ * @brief JoyControl plugin
+ * @file joy_control.cpp
+ * @author Francois Chataigner <chataign@gmail.com>
+ *
+ * @addtogroup plugin
+ * @{
+ */
+/*
+ * This file is part of the mavros package and subject to the license terms
```
vooon: Please add your copyright string.

mavros_extras/mavros_plugins.xml
```diff
@@ -23,4 +23,7 @@
         <class name="cam_imu_sync" type="mavplugin::CamIMUSyncPlugin" base_class_type="mavplugin::MavRosPlugin">
 		<description>Publish camera trigger data for synchronisation of IMU and camera frames</description>
 	</class>
+    <class name="joy_control" type="mavplugin::JoyControlPlugin" base_class_type="mavplugin::MavRosPlugin">
```
vooon: Looks like 4 spaces instead of tab.

mavros_extras/src/plugins/joy_control.cpp
```diff
@@ -0,0 +1,266 @@
...
+
+#include <thread>
+#include <mutex>
+using namespace std;
+
+////////////////////////////////////////////////////////////////////////////////
+////////////////////////////////////////////////////////////////////////////////
+
+template<typename Service>
+bool call_service( const string& service_name, const typename Service::Request& req )
```
vooon: Why not to use `ros::ServiceClient`?
[tutorial link](http://wiki.ros.org/ROS/Tutorials/WritingServiceClient%28c%2B%2B%29#roscpp_tutorials.2BAC8-Tutorials.2BAC8-WritingServiceClient.Writing_the_Client_Node)

mavros_extras/src/plugins/joy_control.cpp
```diff
@@ -0,0 +1,266 @@
...
+	void initialize(UAS &uas )
+	{
+		uas_ = &uas;
+
+		int rc_rate_hz;
+
+		nh_.param("rc_rate_hz",     rc_rate_hz,      10 );
+
+		nh_.param("arming_service", arming_service_, string("/mavros/cmd/arming") );
+		nh_.param("mode_service",   mode_service_,   string("/mavros/set_mode") );
```
vooon: Wrong! You should use '~set_mode` etc.

vooon: Also i thing that `nh_.param<std::string>()` looks better.

mavros_extras/src/plugins/joy_control.cpp
```diff
@@ -0,0 +1,266 @@
...
+
+		nh_.param("mode_mapping",   mode_mapping_,  {3,4} );
+		nh_.param("mode_names",     mode_names_,    {"MANUAL","ALTCTL","OFFBOARD"} );
+
+		nh_.param("arm_button",     arm_button_,    2   );
+
+		nh_.param("x_axis",         x_.axis,        1   );
+		nh_.param("x_scale",        x_.scale,       1.0 );
+		nh_.param("x_offset",       x_.offset,      0.0 );
+		nh_.param("x_deadzone",     x_.deadzone,    0.2 );
```
vooon: Why not move that in subnamespace? E.g. yaml:
```yaml
joy_control:
  arm_button: 2
  x:
    axis: 1
    scale: 1.0
    offset: 0.0
    deadzone: 0.2

  y:
    ...
```

vooon: Also you may place that loaders into `struct Control` to not repeat.

mavros_extras/src/plugins/joy_control.cpp
```diff
@@ -0,0 +1,266 @@
...
+		nh_.param("z_scale",        z_.scale,       1.0 );
+		nh_.param("z_offset",       z_.offset,      0.0 );
+		nh_.param("z_deadzone",     z_.deadzone,    0.0 );
+
+		nh_.param("r_axis",         r_.axis,        4   );
+		nh_.param("r_scale",        r_.scale,       1.0 );
+		nh_.param("r_offset",       r_.offset,      0.0 );
+		nh_.param("r_deadzone",     r_.deadzone,    0.2 );
+
+		joystick_sub_ = nh_.subscribe("/joy", 10, &JoyControlPlugin::handle_joystick, this );
```
vooon: Remove `/`.

mavros_extras/src/plugins/joy_control.cpp
```diff
@@ -0,0 +1,266 @@
...
+		state_sub_ = nh_.subscribe("/mavros/state", 10, &JoyControlPlugin::handle_state, this );
+
+		thread_.reset( new std::thread( &JoyControlPlugin::thread_call, this, rc_rate_hz ) );
```
vooon: I'm usually set thread name before storing in `std::thread` (utils).

mavros_extras/src/plugins/joy_control.cpp
```diff
@@ -0,0 +1,266 @@
...
+		nh_.param("z_offset",       z_.offset,      0.0 );
+		nh_.param("z_deadzone",     z_.deadzone,    0.0 );
+
+		nh_.param("r_axis",         r_.axis,        4   );
+		nh_.param("r_scale",        r_.scale,       1.0 );
+		nh_.param("r_offset",       r_.offset,      0.0 );
+		nh_.param("r_deadzone",     r_.deadzone,    0.2 );
+
+		joystick_sub_ = nh_.subscribe("/joy", 10, &JoyControlPlugin::handle_joystick, this );
+		state_sub_ = nh_.subscribe("/mavros/state", 10, &JoyControlPlugin::handle_state, this );
```
vooon: Use `~state`.

mavros_extras/src/plugins/joy_control.cpp
```diff
@@ -0,0 +1,266 @@
...
+	}
+
+private:
+	ros::NodeHandle nh_;
+	UAS *uas_;
+	ros::Subscriber joystick_sub_, state_sub_;
+	mavros_msgs::State state_;
+
+	std::mutex msg_mutex_;
+	shared_ptr<std::thread> thread_;
```
vooon: Why? `std::thread` has `.swap()` method: https://github.com/mavlink/mavros/blob/master/mavros/include/mavros/setpoint_mixin.h#L73

mavros_extras/src/plugins/joy_control.cpp
```diff
@@ -0,0 +1,266 @@
...
+	mavlink_message_t ctrl_msg_;
```
vooon: Why? Use local variable.

mavros_extras/src/plugins/joy_control.cpp
```diff
@@ -0,0 +1,266 @@
...
+	{
+		ros::Rate rate(rate_hz);
+
+		// We need to keep sending messages at a fixed
+		// rate or PX4 will trigger RC timeouts
+
+		while (1)
+		{
+			msg_mutex_.lock();
+			UAS_FCU(uas_)->send_message(&ctrl_msg_);
```
vooon: Bad idea! At least you abuse drop packet counter buy sending same seq nr.

vooon: Perhaps better do not use thread at all. Joy node has param for minimal rate.

vooon: Also if Joy falls you will send same data and FCU timeout newer occurs.

vooon:
Travis uses GCC 4.6: your problem: https://travis-ci.org/mavlink/mavros/jobs/112583245#L4751

mavros_extras/src/plugins/joy_control.cpp
```diff
@@ -0,0 +1,266 @@
...
+		nh_.param("arming_service", arming_service_, string("/mavros/cmd/arming") );
+		nh_.param("mode_service",   mode_service_,   string("/mavros/set_mode") );
+
+		// Mode switches on remotes can be reported a single flag, or as a binary
+		// combination of several flags: eg. mode1=[0,0]=0, mode2=[1,0]=1, mode3=[0,1]=2
+		// mode_mapping={3,4} means a binary combination of the 4th and 5th values
+		// in the sensor_msgs::Joy buttons array
+
+		nh_.param("mode_mapping",   mode_mapping_,  {3,4} );
+		nh_.param("mode_names",     mode_names_,    {"MANUAL","ALTCTL","OFFBOARD"} );
```
vooon: Travis falls here. Maybe newer GCC compile that, but not 4.6.

vooon:
Also please uncrustify.

chataign:
@vooon Thanks for reviewing the code! Much appreciated. I will check your comments today and update the code. After that do you think yourself or someone else would be able to test it?

chataign:
@vooon I just committed my changes, I think I addressed all your comments. As you suggested I removed the thread, and I'm now sending the mavlink messages at a fixed rate using a timer. I left the params for the axis/scale/offset/deadzone of each controller as they were, because I find they make it easier to see all default values together. I didn't understand your comment about using yaml for params, if you can give explain I'll have a look again. Thanks!

mavros_extras/mavros_plugins.xml
```diff
@@ -20,7 +20,10 @@
 	<class name="vibration" type="mavplugin::VibrationPlugin" base_class_type="mavplugin::MavRosPlugin">
 		<description>Publish VIBRATION message data from FCU.</description>
 	</class>
-        <class name="cam_imu_sync" type="mavplugin::CamIMUSyncPlugin" base_class_type="mavplugin::MavRosPlugin">
+		<class name="cam_imu_sync" type="mavplugin::CamIMUSyncPlugin" base_class_type="mavplugin::MavRosPlugin">
```
vooon: Why? I think that was unintended change.

mavros_extras/src/plugins/joy_control.cpp
```diff
@@ -0,0 +1,237 @@
...
+		, uas_(nullptr) {}
+
+	void initialize(UAS &uas )
+	{
+		uas_ = &uas;
+
+		int msg_rate;
+
+		plugin_nh_.param<int>("msg_rate", msg_rate, 10 );
+		plugin_nh_.param<int>("arm_button", arm_button_, 2 );
```
vooon: You do not need to specify type in each `NodeHandle::param<>()`, for most calls C++ type out works fine.

mavros_extras/src/plugins/joy_control.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+		// Mode switches on remotes can be reported a single flag, or as a binary
+		// combination of several flags: eg. mode1=[0,0]=0, mode2=[1,0]=1, mode3=[0,1]=2
+		// mode_mapping={3,4} means a binary combination of the 4th and 5th values
+		// in the sensor_msgs::Joy buttons array
+
+		plugin_nh_.param< vector<int> >("mode_mapping", mode_mapping_, {3,4} );
+		plugin_nh_.param< vector<string> >("mode_names", mode_names_, {"MANUAL","ALTCTL","OFFBOARD"} );
```
vooon: That was exception. Here you should specify type.

mavros_extras/src/plugins/joy_control.cpp
```diff
@@ -0,0 +1,237 @@
...
+		plugin_nh_.param<int>      ("r_axis",       r_.axis,        4   );
+		plugin_nh_.param<double>   ("r_scale",      r_.scale,       1.0 );
+		plugin_nh_.param<double>   ("r_offset",     r_.offset,      0.0 );
+		plugin_nh_.param<double>   ("r_deadzone",   r_.deadzone,    0.2 );
+
+		arming_srv_ = mavros_nh_.serviceClient<mavros_msgs::CommandBool>("cmd/arming");
+		set_mode_srv_ = mavros_nh_.serviceClient<mavros_msgs::SetMode>("set_mode");
+
+		arming_srv_.waitForExistence();
+		//set_mode_srv_.waitForExistence();
```
vooon: No! You may lock plugin loading! No locks in `initialize()` allowed.

mavros_extras/src/plugins/joy_control.cpp
```diff
@@ -0,0 +1,237 @@
...
+#include <mavros_msgs/State.h>
+#include <mavros_msgs/SetMode.h>
+
+using namespace std;
+
+////////////////////////////////////////////////////////////////////////////////
+////////////////////////////////////////////////////////////////////////////////
+
+namespace mavplugin {
+struct Control
```
vooon: Inside `class JoyControllPlugin` that ok, but here please add namespace or rename to `JoyControl`.

mavros_extras/src/plugins/joy_control.cpp
```diff
@@ -0,0 +1,237 @@
...
+#include <pluginlib/class_list_macros.h>
+
+#include <sensor_msgs/Joy.h>
+
+#include <mavros_msgs/SetMode.h>
+#include <mavros_msgs/CommandBool.h>
+#include <mavros_msgs/State.h>
+#include <mavros_msgs/SetMode.h>
+
+using namespace std;
```
vooon: I'm usually do not do that. clang says that better list all used classes...

mavros_extras/src/plugins/joy_control.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+/**
+ * @brief Joystick Control plugin
+ */
+class JoyControlPlugin : public MavRosPlugin
+{
+	ros::NodeHandle mavros_nh_, plugin_nh_;
+	UAS *uas_;
+	mavros_msgs::State state_;
+	mavlink_message_t ctrl_msg_;
```
vooon: That was still there. Just send it in `handle_joystick()` and forget!

vooon:
I can't test now. So better if you can check it yourself.

About yaml. That was standard rosparam file storage. I thing that you can reduce code duplication by moving that to `struct Control`. And perhaps it is little more readable.

```cpp
struct JoyAxis {
    void load_params(ros::NodeHandle &priv_nh_, std::string name, size_t axdf) {
        auto nh_ = ros::NodeHandle(priv_nh_, name);

        nh_.param<int>("axis", axis, axdf);
        nh_.param("offset", offset, 0.0);
    }

  size_t axis;
  double offset;
}
...
x_.load_params(pnh, "x", 0);:w
```

vooon:
Next thing: you do not need timer at all! Just encode message and send immediately!
FCU has safety timer - it prevents you from uncontrollable flight when your joystick node falls for some reason.

chataign:
@vooon I made more changes based on your last comments:
* renamed `Control` struct into `JoyAxis` and moved params loading into structure
* removed timer and sending mavlink message directly in `handle_joystick()`
* no longer waiting for arming/mode services in `initialize()` but warn user if services are not up

I tested the plugin with a Futaba flight controller and a Logitech gamepad. In the end I had to introduce an additional `JoyButton` structure to handle button inputs -- apologies if this change will require more reviewing on your part. This is because on the Futaba controller button states can be represented by multiple flags ([0,0], [0,1], etc), while on the gamepad each button has its own flag.

I found that the best way to configure the plugin for a given controller is to use a yaml file as you suggested. For example my configuration for the logitech gamepad looks like this:

```
arm: { switches: [0], values: [1] }
num_modes: 3
mode1: { switches: [2], values: [0], custom_mode: "MANUAL" }
mode2: { switches: [3], values: [1], custom_mode: "ALTCTL" }
mode3: { switches: [1], values: [1], custom_mode: "OFFBOARD" }
x: { axis: 3, scale: 1000 }
y: { axis: 2, scale: -1000 }
z: { axis: 1, scale: 2000 }
r: { axis: 0, scale: -1000 }
```

And for the futaba controller:

```
arm: { switches: [2], values: [1] }
num_modes: 3
mode1: { switches: [3,4], values: [0,0], custom_mode: "MANUAL" }
mode2: { switches: [3,4], values: [1,0], custom_mode: "ALTCTL" }
mode3: { switches: [3,4], values: [0,1], custom_mode: "OFFBOARD" }
x: { axis: 1, scale: 1000 }
y: { axis: 0, scale: -1000 }
z: { axis: 2, scale: 2000 }
r: { axis: 4, scale: -1000 }
```

then I load the configuration in my launch file using:

```
<rosparam param="joy_control" command="load" file="$(find package)controllers/logitech_gamepad.yaml" />
```

Note that the number of nodes is specified because I don't think it is possible to load arbitrary structures the standard ROS param methods. There are probably ways to manage it but I thought it would be overkill for this plugin.

mavros_extras/src/plugins/joy_control.cpp
```diff
@@ -0,0 +1,262 @@
...
+	}
+
+	bool load( ros::NodeHandle &nh, const std::string& button_ns, ModeCallback mode_callback )
+	{
+		auto button_nh = ros::NodeHandle( nh, button_ns );
+
+		std::string custom_mode;
+		if ( !button_nh.getParam( "custom_mode", custom_mode ) ) return false;
+
+		return load( nh, button_ns, (Callback) boost::bind( mode_callback, custom_mode ) );
```
vooon: Why not `std::bind`?

mavros_extras/src/plugins/joy_control.cpp
```diff
@@ -0,0 +1,262 @@
...
+
+		int num_modes=0;
+		plugin_nh_.getParam( "num_modes", num_modes );
+
+		for ( int i=0; i< num_modes; ++i )
+		{
+			std::string mode_ns = "mode" + std::to_string(i+1);
+			if ( button.load( plugin_nh_, mode_ns, mode_callback ) ) buttons_.push_back(button);
+			else ROS_ERROR_STREAM("failed to read button for mode=" << mode_ns );
+		}
```
vooon: You can request list of param keys and then iterate over it.

vooon:
You may request all keys form param server. Here is example [distance_sensor][1].

You should load parameters before node start into `/mavros/joy_control`.
And i'm not sure does parameter cleanup remove all from `/mavros` or not.
Usually for that i use `<fcu>_config.yaml` from `mavros/launch` (or modified copy in robot bringup package).
And i think you should add your default config (same as in plugin code) to that files.

[1]: https://github.com/mavlink/mavros/blob/master/mavros_extras/src/plugins/distance_sensor.cpp#L117

vooon:
Oh, and i forgot to ask why not this approach:

1. make simple ManualControl plugin (with single topic)
2. hack `mavteleop` to support it

chataign:
@vooon I tried modifying `mavteleop` as you suggested and this is a much better approach. With only a few changes I was able to send RC overrides allowing me to arm, change modes and control the drone. I will test this with a gamepad before I submit my changes for review. Thanks again for your help

chataign:
@vooon this pull request should be discarded if you're happy with modifying `mavteleop` instead

PR #517, opened 2016-03-02
Fix px4flow quality

gauxonz:
The entry 'quality' under the topic "px4flow/raw/optical_flow_rad" is always zero.The reason for this is in mavros/mavros_extras/src/plugins/px4flow.cpp the quality in the message is not forwarded to the topic. Please add one line
flow_rad_msg->quality = flow_rad.quality;
at around line 117. I tested it and this will fix it well.

### DIFF
--- a/mavros_extras/src/plugins/px4flow.cpp
+++ b/mavros_extras/src/plugins/px4flow.cpp
@@ -88,14 +88,14 @@ private:
 		 */
 		auto int_xy = UAS::transform_frame_aircraft_baselink(
 				Eigen::Vector3d(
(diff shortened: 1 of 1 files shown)


mhkabir:
Thanks, but when the format changes?

vooon:
Thank you!

Is that second commit uncrustify?

gauxonz:
@vooon Yes. One commit after the add of code, then the second one caused by running uncrustify (it add several tabs in that code and i don't know why @mhkabir , but it doesn't affect the function anyway) before pushing to my repo,

PR #518, opened 2016-03-07
modified mavteleop to support RC modes

chataign:
update to `mavteleop` to allow support modes in -rc mode
* modes are read from the `~rc_modes` in yaml format eg. with `<rosparam>` (see examples below)
* tested with [Futaba flight controller](http://www.realflight.com/products/gpmz4520.html) and [Logitech gamepad](http://gaming.logitech.com/en-roeu/product/f710-wireless-gamepad)
* removed support for arm/disarm in RC mode as it is already available over RC with stick in lower-right position

### DIFF
--- a/mavros_extras/scripts/mavteleop
+++ b/mavros_extras/scripts/mavteleop
@@ -99,8 +99,6 @@ rc_channels = {
     'throttle': RCChan('throttle', 2, 0.0)
 }
 
-rc = OverrideRCIn() # must be global to retain mode value
-
 def arm(args, state):
     try:
         command.arming(value=state)
@@ -134,11 +132,12 @@ def rc_override_control(args):
         v.load_param()
 
     rc_modes = RCMode.load_param()
+	rc = OverrideRCIn()
 
     override_pub = rospy.Publisher(mavros.get_topic("rc", "override"), OverrideRCIn, queue_size=10)
 
     def joy_cb(joy):
-        global rc
+        global rc # must be global to retain mode value
 
         # get axes normalized to -1.0..+1.0 RPY, 0.0..1.0 T
         roll = get_axis(joy, 'roll')
--- a/mavros_extras/scripts/mavteleop
+++ b/mavros_extras/scripts/mavteleop
@@ -137,8 +137,6 @@ def rc_override_control(args):
     override_pub = rospy.Publisher(mavros.get_topic("rc", "override"), OverrideRCIn, queue_size=10)
 
     def joy_cb(joy):
-        global rc # must be global to retain mode value
-
         # get axes normalized to -1.0..+1.0 RPY, 0.0..1.0 T
         roll = get_axis(joy, 'roll')
         pitch = get_axis(joy, 'pitch')
--- a/mavros_extras/scripts/mavteleop
+++ b/mavros_extras/scripts/mavteleop
@@ -44,6 +44,28 @@ class RCChan(object):
         # warn: limit check
         return arduino_map(pos, self.min_pos, 1.0, self.min, self.max)
 
+class RCMode(object):
+	def __init__( self, name, joy_flags, rc_channel, rc_value ):
+		self.name = name
+		self.joy_flags = joy_flags
+		self.rc_channel = rc_channel
+		self.rc_value = rc_value
+		
+	@staticmethod
+	def load_param(ns='~rc_modes/'):
+		yaml = rospy.get_param(ns)
+		return [ RCMode( name, data['joy_flags'], data['rc_channel'], data['rc_value'] ) 
+			for name,data in yaml.iteritems() ]
+		
+	def is_toggled(self,joy):
+		for btn,flag in self.joy_flags:
+			if joy.buttons[btn] != flag:
+				return False
+		return True
+		
+	def apply_mode(self,joy,rc):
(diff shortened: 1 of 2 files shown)


chataign:
Logitech gamepad
```
axes_map:
  'roll': 2
  'pitch': 3
  'yaw': 0
  'throttle': 1
rc_modes:
  offboard:
    rc_channel: 4
    rc_value: 1000
    joy_flags: [ [2,1] ]
  auto:
    rc_channel: 4
    rc_value: 1500
    joy_flags: [ [3,1] ]
  manual:
    rc_channel: 4
    rc_value: 2000
    joy_flags: [ [0,1] ]
```

chataign:
Futaba flight controller

```
axes_map:
  'roll': 0
  'pitch': 1
  'yaw': 4
  'throttle': 2
axes_scale:
  'roll': 1.5
  'pitch': 1.5
  'yaw': 1.5
  'throttle': 1.5
rc_modes:
  offboard:
    rc_channel: 4
    rc_value: 1000
    joy_flags: [ [3,0], [4,1] ]
  auto:
    rc_channel: 4
    rc_value: 1500
    joy_flags: [ [3,1], [4,0] ]
  manual:
    rc_channel: 4
    rc_value: 2000
    joy_flags: [ [3,0], [4,0] ]
```

mavros_extras/scripts/mavteleop
```diff
@@ -109,9 +133,13 @@ def rc_override_control(args):
     for k, v in rc_channels.iteritems():
         v.load_param()
 
+	rc_modes = RCMode.load_param()
```
vooon: `pep8` or `pyflake` should alert here: mix of tabs and spaces is a very bad thing in python.

Python indent should be done by 4 spaces only.

mavros_extras/scripts/mavteleop
```diff
@@ -77,6 +99,8 @@ rc_channels = {
     'throttle': RCChan('throttle', 2, 0.0)
 }
 
+rc = OverrideRCIn() # must be global to retain mode value
```
vooon: I'm not a fan of global variables where we can avoid it.

Store required value in RCMode class, anyway you apply that on each turn.

vooon:
Also you should change `mavros_extras/launch/f710_joy.yaml`.

And i'm not sure how good idea to change flight mod via override (e.g. PX4 uses several channels for switches).
I'm think that MANUAL_CONTROL is little better, at least PX4 has safety timeout (and before 1.0 they do not support overrides at all).

mavros_extras/scripts/mavteleop
chataign: The RC message needs to be a global variable in this case, to store the last mode selected using the gamepad as the buttons are immediately released (unlike the flight controller). However it can be moved inside `rc_override_control()` to minimize scope.

chataign:
Are you happy to merge as it is or do you want me to try using MANUAL_CONTROL instead?

vooon:
I do not much like setting mode via the override, but if it works - ok. `MANUAL_CONTROL` **may be** better.

andrejpan:
Hey,
could somebody explain how were you able to make configuration file for these two gamepads? I am trying to use Sony ps3 gamepad, but more or less nothing happening. Just arm/disarm of motors works.

chataign:
I haven't tried using `mavteleop` with that gamepad but there's no reason why it shouldn't work.

Here's how I've been setting it up:

1. Run `joy_node`, `rostopic echo /joy` topic. Check what your buttons and axes are mapped to in the `sensor_msgs::Joy` message
2. Create a `,yaml` file for your gamepad where you define `axes_map`, `axes_scales` and `rc_modes`. For each RC mode specify the mapping between the flag in `/joy/buttons` and the channel value in the `mavros::OverrideRCIn` message. echo `/mavros/rc/override` to check that all your mappings work as expected
3. Pass the `.yaml` file to `mavteleop` using `rosparam` in your launch file. For example: `<rosparam param="mavteleop" command="load" file="$(find <package>)/launch/sony_ps3.yaml" />`. Note that this needs to be in the `mavros` node (this might be a bug)
4. Run `mavros`, `mavteleop` and your simulator (presumably) from your launch file. At this point you should be able to arm, and take off manually
```
<node name="mavteleop" pkg="mavros_extras" type="mavteleop" args="-rc" output="$(arg log_output)" ns="mavros">
	<remap from="joy" to="/joy"/>
</node>

```
5. Setting the correct `OverrideRCIn` channel values to switch between flight modes (`MANUAL`/`ALTCTL`/`OFFBOARD`/etc) isn't straightforward (at least with PX4). Check the Pixhawk documentation or even the source code.
6. It is entirely possible that there are bugs or things to clarify. Please let me know how you get on!

andrejpan:
Hey,
tnx for the help. I hacked some code here:
 https://gitlab9.in.tum.de/pangerca/mavros/commit/232b84704a100c241239108e8eb06d0aed29cb1e

For the task I want to navigate drone with a ps3 gamepad and because it is working with ros there only way that I send messages to drone via mavros, correct? This means that I need to put drone in to OFFBOARD mode and then select attitude, velocity or position control? Why is there also rc override mode or what it does? I am reading documentation, but I really have problem understanding it.

PR #519, opened 2016-03-11
Fix track visualization

chataign:
track visualization in `copter_visualization` was constantly creating new markers (`marker.id` constantly incremented) which would eventually clog rviz and slow in down.

introduced `max_track_size` parameter to limit number of track markers, also used `CUBE_LIST` marker type instead of `CUBE` for efficiency as it allows rviz to batch rendering.

http://wiki.ros.org/rviz/DisplayTypes/Marker#Cube_List_.28CUBE_LIST.3D6.29

### DIFF
--- a/mavros_extras/src/copter_visualization.cpp
+++ b/mavros_extras/src/copter_visualization.cpp
@@ -26,43 +26,48 @@
 static std::string fixed_frame_id;
 static std::string child_frame_id;
 static double marker_scale;
-static int max_track_size=100;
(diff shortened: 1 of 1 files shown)


vooon:
@chataign looks fine, but please run `uncrustify` as described in [contributing readme no. 7][1].

[1]: https://github.com/mavlink/mavros/tree/master/mavros#contributing

vooon:
@chataign thank you!

PR #533, opened 2016-04-20
Fix for kinetic std::isnan.

jgoppert:


vooon:
Ok, CI passed after manual restart. Thank you!

Issue #536, opened 2016-04-29
[setpoint_raw] attitude_cb orientation not transformed!


vooon:
@justinthomas found a bug at line 255: https://github.com/mavlink/mavros/blob/master/mavros/src/plugins/setpoint_raw.cpp#L255

PR #537, opened 2016-04-29
Fix bug #536 with orientation in setpoint_raw

justinthomas:
Fixes a bug where the ned_desired_orientation was not actually passed into set_attitude_target. Instead, the desired_orientation (wrong frame) was passed.

This addresses #536


vooon:
CI ok. Thank you for bug hunting!

Issue #553, opened 2016-05-28
MAVLink 2.0 support required


LorenzMeier:
@vooon In order to not stall the further development and addition of new messages we will need MAVLink 2.0 support (not for signing and v2.0 format, but for the MAVLink 2.0 binaries which support v1.0 and v2.0 messaging).

vooon:
Lorenz, #543 & #545. But first i need some testing target like SITL.
As i know APM already merged to master, so i hope to go with #543 soon.

LorenzMeier:
PX4 has MAVLink 2.0 support in master for about two weeks as well.

vooon:
MAVLink 2.0 support is done, but: only Kinetic and require C++11 generator from `mavlink2-cxx11`.

Issue #522, opened 2016-03-14
libmavconn: Generate mavlink_dialect.h


vooon:
Since ROS version of mavlink package provide list of supported dialects we may generate that file.

At least that allow more easy inclusion of custom dialects.

vooon:
Not sure on status, since in mavlink 2.0 we may support all dialects at once.

samacumen:
Guys, how do we create this mavlink_dialect.h file. This is the only file missing in this project and I am trying to compile libMavConn library (without ROS - as it is independent of it). Any lead/instructions on how to generate the mavlink_dialect.h will be helpful.

Thanks!

Issue #286, opened 2015-05-03
Use `mavlink_frame_char()` in libmavconn for better forwarding


vooon:
See https://github.com/mavlink/mavlink/commit/e485a67e4e9c1dd2e96acc67a229daf49893aaa1

But we need to find how to pass additional bad crc flag in MAVConnInterface API.

vooon:
Done by #554.

PR #554, opened 2016-06-06
Mavlink 2.0 support (WIP)

### LINKED ISSUE #543: Support MAVLink 2.0
MAVLink 2.0 megred: mavlink/mavlink#535

- [x] libmavconn
- [x] mavros_msgs
- [x] mavros_node
- [ ] std plugins
- [ ] extras

vooon:
This is WIP PR for Mavlink 2.0 support.

Fixes #543 (#553), #522, #286.


vooon:
I decided to make C++ MAVLink library: mavlink/mavlink#568

vooon:
I found bug with memory leak. Test - simple echo server (`test_hang.cpp`).
It appears only when `send_message()` was called from that object IO thread, for example `message_received` slot send something back.

Also that bug exists in current master too.

vooon:
`libmavconn` - done. API redone, Tx queue now limited to 1k messages, no more sig-slot usage.
Only C++11 std and boost::asio (perhaps it may work with standalone C++11 version of ASIO library).

Message sending now looks more like ROS publish:
```cpp
mavlink::common::msg::HEARTBEAT hb{};
hb.type = int(mavlink::common::MAV_TYPE::ONBOARD_CONTROLLER);
// ...

link_ptr->send_message(hb);
```

vooon:
All plugins ported. One more thing i do not like - static functions of UAS.

TODO: fix tests, protocol version selection param.

vooon:
Finally! It works!

But not fully tested and require PR https://github.com/mavlink/mavlink/pull/568

vooon:
Last two things for this PR is `UAS::sig_connection_changed` and idea to move message handlers out of FCU io-thread.

But i decided not to use ROS interprocess pub-sub, so lets try as is.

PR #572, opened 2016-07-12
Caktin_lint: correct most errors

khancyr:


mavros_extras/CMakeLists.txt
```diff
@@ -63,16 +63,16 @@ include_directories(
   ${mavlink_INCLUDE_DIRS}
 )
 
-add_library(mavros_extras
+add_library(${PROJECT_NAME}
```
vooon: Hello bugs!

vooon:
Sorry, but i ignore that PR. libmavros depends on gtest, but not rosunit. I do not know why lint say that.
PROJECT_NAME in most cases makes code more ugly.

Only alphabetical sort is ok, but it is absolutely not required.

khancyr:
Ok no problem!

PR #635, opened 2016-12-13
Plugins: add system_status to state message

khancyr:
Used to know the state of APM based drone with Companion Computer.


vooon:
LGTM! Thank you @khancyr!

PR #655, opened 2017-02-14
Update setpoint_velocity to simple twist for compatibility with ROS n…

khancyr:
…av stack

Current setpoint_velocity require twistStamped but all ROS navigation stack and what is ising move_base is using simple Twist (http://wiki.ros.org/move_base?distro=kinetic ). Using simple twist will allow direct mapping from navigation to setpoint_velocity and use ROS controllers

### DIFF
--- a/mavros/src/plugins/setpoint_velocity.cpp
+++ b/mavros/src/plugins/setpoint_velocity.cpp
@@ -20,6 +20,7 @@
 #include <eigen_conversions/eigen_msg.h>
 
 #include <geometry_msgs/TwistStamped.h>
+#include <geometry_msgs/Twist.h>
 
 namespace mavros {
 namespace std_plugins {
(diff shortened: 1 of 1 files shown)


vooon:
I'm not sure that i want to break ROS API. Maybe better add additional subscriber?

khancyr:
@vooon Ok I will rework that tomorrow to just add the twist listener . I submit this before talking to you because I didn't find an example in ROS using cmd_vel with twistStamped, But if it was done like that before, just adding a subcriber will preserve current compatibiliy and add more on ROS side!

vooon:
Old version of plugins has subscriber type switch parameter, but i removed it because of error prone.

khancyr:
@vooon ok, rebase and twist listener added. It is working as intend with ROS nav stack see https://www.youtube.com/watch?v=kYMzVeW-0-E&index=5&list=PL6sCNLbHuYxaLtItzADD5zHxixUtQPSvk as example (I know slam is failing but vel order are well send !)

vooon:
Merged. Thank you!

PR #660, opened 2017-02-23
Vision plugin fixes, Camera trigger fixes

mhkabir:
1. Fixes a long-term bug in the vision plugin. Yaw between ENU and NED will now be rotated correctly. This was already done correctly in the `mocap` plugin (https://github.com/mavlink/mavros/blob/master/mavros_extras/src/plugins/mocap_pose_estimate.cpp#L117), but we were missing it in `vision_pose`.

2. Minor renaming in camera trigger interface to match PX4 naming conventions.

@vooon Please review and merge. These have been tested extensively both in PX4 SITL (with laser-scan odometry) and on hardware.


vooon:
Looks good to me, thank you Kabir.

mhkabir:
Great!

PR #679, opened 2017-03-22
Add two publishers for reached and current/next waypoint

mw46d:
Just an idea;-)

### DIFF
--- a/mavros/src/plugins/waypoint.cpp
+++ b/mavros/src/plugins/waypoint.cpp
@@ -15,6 +15,7 @@
  */
 
 #include <chrono>
+#include <std_msgs/Int16.h>
 #include <condition_variable>
 #include <mavros/mavros_plugin.h>
 #include <pluginlib/class_list_macros.h>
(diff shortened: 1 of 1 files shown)


vooon:
Current wp available in list, reached not needed.

mw46d:
Polling the WP list to find out the current WP is kind of inconvenient when you try to control some ROS actions based on it. For me, it's basically turning on the camera at certain points. But it could be different things;-)

sputnick1124:
I don't think this pull request is entirely unneeded. I have implemented the same in my own system for a little while now [in an almost identical commit](https://github.com/sputnick1124/mavros/commit/eab37ab319ba8d971aa690ac0aa11e1d76ec2be4). It is otherwise not possible to know when the vehicle has actually successfully reached the waypoint but to monitor the vehicle's position yourself and compare the goal to your current position. 
An example use case is if you want to force some action when the last waypoint in the mission has been achieved, simply looking at the `is_current` field of your list of waypoints will not tell you anything and there is no other indication that you have indeed reached your goal. Unless I'm missing something?

vooon:
@mw46d you subscribe to wp list, and get notification when it changes (cb call), no polling.

@sputnick1124 reached may be useful, but it might be unimplementd. You can add additional point before end loiter and check when it switch. All other intermediate reached may be detected by is_current change.

PR #636, opened 2016-12-14
Px4 hil plugins

mzahana:
This branch adds support for MAVLink messages that are used in HIL with PX4 firmware.

Added messages:
- HilGPS, HilControls, HILSensor, HilStateQuaternion, ActuatorControlTarget

Plugins:
- hil_controls, hil_sensor, hil_state_quaternion, hil_gps, actuator_control_target

This is demonstrated with VREP simulator. [Demo video](https://www.youtube.com/watch?v=NoEw4upSE4M&feature=youtu.be)

Issue #529, opened 2016-03-31
GPS and global_position altitude (NavSatFix - WGS84; MAVLink - AMSL)


mantelt:
Hi

I just tried to figure out in which system the altitude is published for gps and global_position messages. As far as I can tell, the altitude of gps_raw_int and global_position_int are only scaled to meters and published in NavSatFix messages in their respective ROS topics.

BUT: while the [mavlink specification](http://mavlink.org/messages/common#GPS_RAW_INT) states that the altitude is in AMSL, the [ROS specification for NavSatFix](http://docs.ros.org/api/sensor_msgs/html/msg/NavSatFix.html) states that it should be published relative to the WGS84 ellipsoid (which to my knowledge is in general not the same).

Any thoughts?

vooon:
Oh, yes, i missed that problem.

mantelt:
I also almost forgot about my issue here. I again looked at the code and think this needs a fix for consistency with the definitions. Are you looking into this?

vooon:
I'm not sure when i can deal with this issue, but at least now the bug is known.

TSC21:
@mantelt can you please issue a PR on this if you have found the solution?

mantelt:
Unfortunately, I did not fix this yet.
A quick search revealed either the solution posted above using GeographicLib, or basically re-implementing that method using the publicly available data of the geoid / ellipsoid to calculate the _undulations_ (see http://earth-info.nga.mil/GandG/wgs84/gravitymod/egm2008/egm08_wgs84.html) or even calculate them whenever needed (code can be found online, e.g. http://stackoverflow.com/questions/22196714/wgs84-geoid-height-altitude-offset-for-external-gps-data-on-ios). 

I would vote for using an already implemented library, but I don't know how the (main) developers (i.e. @vooon) would feel about this.
I guess it can be seen as quite an overkill using a big library such as geographic lib just to do the conversion from height above geoid to height above ellipsoid. On the other hand, it would be a nice one liner to do the correction... (+ some include of course)


BTW; Some easy to understand description of the problem can be found here:
http://www.esri.com/news/arcuser/0703/geoid1of3.html

TSC21:
@mantelt thanks for the research. At my understanding, a standalone function to do the calculation is the best way in this case, so to not overwhelm the dependencies of a library like that, unless there are other functions that can be useful to us. Do you find any other function on GeographicLib that can be useful for our use case?
@vooon that do you think?

[...]

PR #684, opened 2017-04-03
HIL Plugin

TSC21:
This PR fixes some of @mzahana #636 PR. @vooon can you please review it?


TSC21:
In a conversation with @vooon we decided it would be better to merge all the plugin files into a single HIL plugin. Do you see any inconvenient on this @mzahana?

mhkabir:
I like the merged version a lot :)

mzahana:
The merged version is much better. I just did not have the time to do it.
Thanks guys.

mavros/src/plugins/hil.cpp
```diff
@@ -0,0 +1,320 @@
...
+	{ }
+
+	void initialize(UAS &uas_)
+	{
+		PluginBase::initialize(uas_);
+		last_time = ros::Time(0.0);
+		period = ros::Duration(0.025);	// 40hz
+
+		hilStateQuaternion_sub = hil_nh.subscribe("hil_state", 10, &HilPlugin::state_quat_cb, this);
+		hilGPS_sub = hil_nh.subscribe("hilgps", 10, &HilPlugin::gps_cb, this);
```
vooon: Why not `hil_gps`? But a bit ugly `/mavros/hil/hilgps`, maybe remove `hil_` prefixes?

mavros/src/plugins/hil.cpp
```diff
@@ -0,0 +1,320 @@
...
+	ros::Subscriber hilSensor_sub;
+
+	ros::Time last_time;
+	ros::Duration period;
+
+	/* -*- rx handlers -*- */
+
+	void handle_hil_controls(const mavlink::mavlink_message_t *msg, mavlink::common::msg::HIL_CONTROLS &hil_controls) {
+		// Throttle incoming messages to 40hz
+		if ((ros::Time::now() - last_time) < period) {
```
vooon: Is that required? Why not leave full stream?

mavros/src/plugins/hil.cpp
```diff
@@ -0,0 +1,320 @@
...
+		//     cog.outl("hil_controls_msg->%s = hil_controls.%s;" % (f, f))
+		// ]]]
+		hil_controls_msg->roll_ailerons = hil_controls.roll_ailerons;
+		hil_controls_msg->pitch_elevator = hil_controls.pitch_elevator;
+		hil_controls_msg->yaw_rudder = hil_controls.yaw_rudder;
+		hil_controls_msg->throttle = hil_controls.throttle;
+		hil_controls_msg->aux[0] = hil_controls.aux1;
+		hil_controls_msg->aux[1] = hil_controls.aux2;
+		hil_controls_msg->aux[2] = hil_controls.aux3;
+		hil_controls_msg->aux[3] = hil_controls.aux4;
```
vooon: No no no! You change preprocessed code! And perhaps current aux1...aux4 is fine.

```
pip3 install --user cogapp
cog.py -cr hil.cpp
```

mavros/src/plugins/hil.cpp
```diff
@@ -0,0 +1,320 @@
...
+		hil_controls_pub.publish(hil_controls_msg);
+	}
+
+	void handle_hil_actuator_controls(const mavlink::mavlink_message_t *msg, mavlink::common::msg::HIL_ACTUATOR_CONTROLS &hil_actuator_controls) {
+		auto hil_actuator_controls_msg = boost::make_shared<mavros_msgs::HilActuatorControls>();
+
+		hil_actuator_controls_msg->header.stamp = m_uas->synchronise_stamp(hil_actuator_controls.time_usec);
+		for (int i = 0; i < 16; ++i) {
+			hil_actuator_controls_msg->controls[i] = hil_actuator_controls.controls[i];
+		}
```
vooon: `std::copy()`

mavros/src/plugins/hil.cpp
```diff
@@ -0,0 +1,320 @@
...
+		gps.epv = epv;
+		gps.vel = vel;
+		gps.vn = vn;
+		gps.ve = ve;
+		gps.vd = vd;
+		gps.cog = cog;
+		gps.satellites_visible = satellites_visible;
+
+		UAS_FCU(m_uas)->send_message_ignore_drop(gps);
+	}
```
vooon: All that set/send functions are used only once. Why not just put that code inside caller func?

mavros/src/plugins/hil.cpp
```diff
@@ -0,0 +1,320 @@
...
+
+		state_quat.ind_airspeed = ind_airspeed;
+		state_quat.true_airspeed = true_airspeed;
+
+		state_quat.xacc = xacc;
+		state_quat.yacc = yacc;
+		state_quat.zacc = zacc;
+
+		UAS_FCU(m_uas)->send_message_ignore_drop(state_quat);
+	}
```
vooon: And that. It might be useful if it takes Eigen::Vector3d and Quaterniond at input and fills message.
But not this function. Just increase loc.

mavros/src/plugins/hil.cpp
```diff
@@ -0,0 +1,320 @@
...
+		sensor.ymag = ymag;
+		sensor.zmag = zmag;
+		sensor.abs_pressure = abs_pressure;
+		sensor.diff_pressure = diff_pressure;
+		sensor.pressure_alt = pressure_alt;
+		sensor.temperature = temperature;
+		sensor.fields_updated = fields_updated;
+
+		UAS_FCU(m_uas)->send_message_ignore_drop(sensor);
+	}
```
vooon: Same as above.

mavros/src/plugins/hil.cpp
```diff
@@ -0,0 +1,320 @@
...
+				int16_t xacc, int16_t yacc, int16_t zacc) {
+		set_hil_state_quaternion(stamp.toNSec() / 1000,
+					qw, qx, qy, qz,
+					rollspeed, pitchspeed, yawspeed,
+					lat, lon, alt,
+					vx, vy, vz,
+					ind_airspeed,
+					true_airspeed,
+					xacc, yacc, zacc);
+	}
```
vooon: Lets copy again!

mavros/src/plugins/hil.cpp
```diff
@@ -0,0 +1,320 @@
...
+					req->cog,
+					req->satellites_visible);
+	}
+
+	void sensor_cb(const mavros_msgs::HilSensor::ConstPtr &req) {
+		// Throttle incoming messages to 40hz
+		if ((ros::Time::now() - last_time) < period) {
+			return;
+		}
+		last_time = ros::Time::now();
```
vooon: That will conflict with `handle_hil_controls()`

mavros/src/plugins/hil.cpp
```diff
@@ -0,0 +1,320 @@
...
+
+		send_hil_sensor(req->header.stamp,
+					req->acc.x, req->acc.y, req->acc.z,
+					req->gyro.x, req->gyro.y, req->gyro.z,
+					req->mag.x, req->mag.y, req->mag.z,
+					req->abs_pressure.fluid_pressure, req->diff_pressure.fluid_pressure, req->pressure_alt.fluid_pressure,
+					req->temperature.temperature,
+					req->fields_updated);
```
vooon: As i mentioned, just fill mavlink message here. And better to use cog.py to preprocess all that vectors and etc.
That helps with copy-paste problems.

mavros_extras/src/plugins/mocap_fake_gps.cpp
```diff
@@ -145,7 +145,7 @@ class MocapFakeGPSPlugin : public plugin::PluginBase
 		pos.eph = 2;
 		pos.epv = 2;
 		pos.fix_type = 3;
-		pos.satellites_visible = 5;
+		pos.satellites_visible = 10;
 		UAS_FCU(m_uas)->send_message_ignore_drop(pos);
 	}
 };
```
vooon: Why to change that plugin in that PR? Looks like mistake.

mavros_msgs/CMakeLists.txt
```diff
@@ -1,14 +1,15 @@
...
-find_package(catkin REQUIRED COMPONENTS geometry_msgs message_generation std_msgs)
+find_package(catkin REQUIRED COMPONENTS geometry_msgs sensor_msgs message_generation std_msgs)
 
 include_directories(include)
 
 add_message_files(
   DIRECTORY msg
   FILES
   ActuatorControl.msg
+  ActuatorControlTarget.msg
```
vooon: As i said, just use `ActuatorControl.msg`.

mavros_msgs/msg/HilControls.msg
```diff
@@ -10,9 +10,6 @@ float32 roll_ailerons
 float32 pitch_elevator
 float32 yaw_rudder
 float32 throttle
-float32 aux1
-float32 aux2
-float32 aux3
-float32 aux4
+float32[4] aux
```
vooon: Now i don't want to change existing message. Keep API.

mavros_msgs/msg/HilGPS.msg
```diff
@@ -0,0 +1,17 @@
+# HilControls.msg
+#
+# ROS representation of MAVLink HIL_GPS
+# See mavlink message documentation here:
+# https://pixhawk.ethz.ch/mavlink/#HIL_GPS
+
+std_msgs/Header header
+uint8 fix_type
+sensor_msgs/NavSatFix fix
```
vooon: `NavSatFix`? You only uses lat/long/alt triplet. And since all other fields uses same raw types as mavlink it is more logical.
Also fix header.

vooon: Perhaps better to add GeoPointInt.msg:
```
uint32 latitude
uint32 longitude
uint32 altitude
```

mavros_msgs/msg/HilSensor.msg
```diff
@@ -0,0 +1,17 @@
...
+
+std_msgs/Header header
+
+geometry_msgs/Vector3 acc
+geometry_msgs/Vector3 gyro
+geometry_msgs/Vector3 mag
+sensor_msgs/FluidPressure abs_pressure
+sensor_msgs/FluidPressure diff_pressure
+sensor_msgs/FluidPressure pressure_alt
+sensor_msgs/Temperature temperature
```
vooon: All that sensor_msgs/ carry unused header and variance. No reason for that, use generic types.

mavros_msgs/msg/HilStateQuaternion.msg
```diff
@@ -0,0 +1,12 @@
+# HilStateQuaternion.msg
+#
+# ROS representation of MAVLink HIL_STATE_QUATERNION
+# See mavlink message documentation here:
+# https://pixhawk.ethz.ch/mavlink/#HIL_STATE_QUATERNION
+
+std_msgs/Header header
+sensor_msgs/Imu imu
+geometry_msgs/Vector3 linear_velocity
+sensor_msgs/NavSatFix fix
```
vooon: Same idea about NavSatFix, but Imu may be ok (except that i don't like waste of header and variances).

mavros/mavros_plugins.xml
```diff
@@ -65,11 +65,6 @@
...
 		<description>Publish altitude values</description>
 	</class>
-	<class name="hil_controls" type="mavros::std_plugins::HilControlsPlugin" base_class_type="mavros::plugin::PluginBase">
-		<description>Publish HIL controls values (deprecated message)</description>
-	</class>
-	<class name="hil_actuator_controls" type="mavros::std_plugins::HilActuatorControlsPlugin" base_class_type="mavros::plugin::PluginBase">
-		<description>Publish HIL controls values</description>
-	</class>
+	<class name="hil" type="mavros::std_plugins::HilPlugin" base_class_type="mavros::plugin::PluginBase">
+		<description>Publish and subscribe HIL controls, state, GPS and sensor values </description>
```
TSC21: ups forgot it

mavros/src/plugins/actuator_control.cpp
```diff
@@ -17,6 +17,7 @@
 #include <mavros/mavros_plugin.h>
 
 #include <mavros_msgs/ActuatorControl.h>
+#include <mavros_msgs/ActuatorControlTarget.h>
```
TSC21: According to the MAVLINK Common message set, https://pixhawk.ethz.ch/mavlink/#ACTUATOR_CONTROL_TARGET, only `ACTUATOR_CONTROL_TARGET` exists at the moment. So I will replace the `ActuatorControl` with `ActuatorControlTarget` only.

mavros/src/plugins/actuator_control.cpp
```diff
@@ -35,29 +36,49 @@ class ActuatorControlPlugin : public plugin::PluginBase {
 	{
 		PluginBase::initialize(uas_);
 
-		actuator_control_sub = nh.subscribe("actuator_control", 10, &ActuatorControlPlugin::actuator_control_cb, this);
+		actuator_control_target_pub = nh.advertise<mavros_msgs::ActuatorControl>("actuator_control_target", 10);
+		actuator_control_sub = nh.subscribe("actuator_control_target", 10, &ActuatorControlPlugin::actuator_control_cb, this);
```
TSC21: This is based on what I said above.

mavros/src/plugins/actuator_control.cpp
```diff
@@ -35,29 +36,49 @@ class ActuatorControlPlugin : public plugin::PluginBase {
...
+
+	void handle_actuator_control_target(const mavlink::mavlink_message_t *msg, mavlink::common::msg::ACTUATOR_CONTROL_TARGET &actuator_control_target)
+	{
+		auto actuator_control_target_msg = boost::make_shared<mavros_msgs::ActuatorControlTarget>();
+		actuator_control_target_msg->header.stamp = m_uas->synchronise_stamp(actuator_control_target.time_usec);
+
+		actuator_control_target_msg->group_mlx = actuator_control_target.group_mlx;
+		for(int i=0; i<8; i++){
+			actuator_control_target_msg->controls[i]=actuator_control_target.controls[i];
+		}
```
TSC21: sure, why not

mavros/src/plugins/actuator_control.cpp
```diff
@@ -35,29 +36,49 @@ class ActuatorControlPlugin : public plugin::PluginBase {
...
+
+		act.time_usec = req->header.stamp.toNSec() / 1000;
+		act.group_mlx = req->group_mlx;
+		std::copy((req->controls).begin(), (req->controls).end(), act.controls.begin());
 		act.target_system = m_uas->get_tgt_system();
 		act.target_component = m_uas->get_tgt_component();
-		std::copy(req->controls.begin(), req->controls.end(), act.controls.begin());	// std::array = boost::array
 
 		UAS_FCU(m_uas)->send_message_ignore_drop(act);
 	}
```
TSC21: Same as I said above

mavros/src/plugins/hil.cpp
TSC21: yeah I forgot the underscore. will fix

mavros/src/plugins/hil.cpp
TSC21: Well that is @mzahana implementation. @mzahana can you explain why you throttle the message here? Any reason to do it?

mavros/src/plugins/hil.cpp
TSC21: how's that?

mavros/src/plugins/hil.cpp
TSC21: sure

mavros/src/plugins/hil.cpp
TSC21: yeah! didn't thought of that. will do it

mavros/src/plugins/hil.cpp
TSC21: will apply the same as you proposed above

mavros/src/plugins/hil.cpp
TSC21: well having functions to send separate from the calls I think it is a good practice in terms of code perception. But ok I'll include the mavlink message fill on the cb.

mavros_extras/src/plugins/mocap_fake_gps.cpp
TSC21: Don't know if it is a mistake. This comes from @mzahana for some reason, probably something he saw during testing. @mzahana can you explain this change?

mavros_msgs/CMakeLists.txt
TSC21: will replace for `ActuatorControlTarget` to keep the logic from the MAVLINK common message set.

mavros_msgs/msg/HilControls.msg
TSC21: HilControls is a deprecated message... don't see why is this a problem

mavros_msgs/msg/HilGPS.msg
TSC21: Well this has a reason: mainly sims like Gazebo like sensor_msgs more than generic types. I can do two things: 1 is to fill the rest of `NavSatFix` message and keep a sustained and complete message. The other is to reverse to a more generic type - probably the `GeoPointInt` is good

mavros_msgs/msg/HilSensor.msg
TSC21: Same as I said before - keep the consistency of the type of data. Sensor data -> sensor_msgs. Can fill the variances and headers so to make them readable. The overhead is minimal.

mavros_msgs/msg/HilStateQuaternion.msg
TSC21: Same about what I said. Will replace though the `NavSatFix`.

mavros/src/plugins/actuator_control.cpp
vooon: I telling about ROS message, not mavlink. And mavlink has `SET_ACTUATOR_CONTROL_TARGET`. No reason to replace existing msg.

mavros_msgs/CMakeLists.txt
vooon: No! I don't like to break ROS API without strong reason. ActuatorControl here since 2015 and it does SET_ACTUATOR_CONTROL_TARGET for all that time.
And no reason to used different msg for return loop ACTUATOR_CONTROL_TARGET.

mavros_msgs/msg/HilControls.msg
vooon: If it deprecated even more reason not to change our msg.

mavros_msgs/msg/HilGPS.msg
vooon: HilGPS.msg already not sensor_msgs. Give a link to src where sensor_msgs included in other message is more handy.

TSC21:
@vooon looks better to you?

mavros_extras/src/plugins/mocap_fake_gps.cpp
mzahana: It is been a while, but I guess there was a minimum number of Sats in px4 firmware to accept GPS data, at least at the time I tested it.

mzahana:
I used V-REP for HIL testing. I tested the old PR yesterday, and it was working OK. It won't work directly with this new PR. The VREP quad model I have has to be modified to adapt to he new message definitions. I won't be able to do this soon as O am having deadlines.  I can share the VREP model for testing and updates, that would be great. Just email me and I will send the model. Thanks.

mavros_extras/src/plugins/mocap_fake_gps.cpp
TSC21: Ok for me I can leave it. @vooon ?

TSC21:
>I used V-REP for HIL testing. I tested the old PR yesterday, and it was working OK. It won't work directly with this new PR. The VREP quad model I have has to be modified to adapt to he new message definitions. I won't be able to do this soon as O am having deadlines. I can share the VREP model for testing and updates, that would be great. Just email me and I will send the model. Thanks.

Can you share your e-mail please? Also, in your e-mail can you send a quick guide in how to setup V-REP to test the plugin (and where to make the changes), as I never used V-REP before. Thanks

mavros_extras/src/plugins/mocap_fake_gps.cpp
vooon: Can you verify that? For me 10 looks to many, but it's ok if FCU require that.

mavros/src/plugins/hil.cpp
```diff
@@ -134,9 +134,9 @@ class HilPlugin : public plugin::PluginBase {
 		state_quat.rollspeed = req->imu.angular_velocity.x;
 		state_quat.pitchspeed = req->imu.angular_velocity.y;
 		state_quat.yawspeed = req->imu.angular_velocity.z;
-		state_quat.lat = req->geo.latitude;	// @warning geographic_msgs/GeoPoint.msg uses WGS 84 reference ellipsoid
-		state_quat.lon = req->geo.longitude;	// @TODO: Convert to AMSL to be received by the FCU
-		state_quat.alt = req->geo.altitude;	// related to issue #529
+		state_quat.lat = (int32_t)req->geo.latitude;	// @warning geographic_msgs/GeoPoint.msg uses WGS 84 reference ellipsoid
+		state_quat.lon = (int32_t)req->geo.longitude;	// @TODO: Convert to AMSL to be received by the FCU
+		state_quat.alt = (int32_t)req->geo.altitude;	// related to issue #529
```
vooon: WHAT are you doing? That is not how you should convert int32 coord to double (or back)!

TSC21:
We still need to consider #529

TSC21:
Ok for the interested I tested this with V-REP and a model from @mzahana and seems to be working ok. Had to change the `SYS_COMPANION` parameter to 921600 and seemed to work. Msgs are being publish also. I don't have a RC controller with me now to test the commands for the actuators but @mzahana, if you have, let us now. Will post a video soon. (tested this on a MindPX btw)

TSC21:
>Is this to allow e.g. joystick control?

Yes

TSC21:
Now I'm checking this: shouldn't ENU->NED transforms be applied also?

TSC21:
Stupid question - of course we have.

mhkabir:
Let me know when this is ready from your side. Will review after.

mzahana:
I am still testing with VREP. Will let you know the next couple of days.

mhkabir:
Also, I would not merge this before we have good docs on dev.px4.io. 

Thanks for understanding - we need high-quality documentation for this feature to not be lost and become unmaintained. Again, great job @TSC21 @mzahana !

TSC21:
@mzahana ROS API msgs work work in the ENU frame. If you want to send the sensor data, positions, etc, to the FCU, you have to apply the proper transforms so the FCU receives in the NED frame. We have functions on mavros that apply these same transforms, so I just applied them.

mzahana:
Yah. I see that. I already did the transformation in VREP. Will remove the transformation then.

TSC21:
@mhkabir yes sure. thanks. I think as is right now is ready to be reviewed. I also retested it on the sim and seems consistent and stable. Let's wait for @mzahana feedback on his tests.

mavros/src/plugins/hil.cpp
```diff
@@ -217,27 +240,41 @@ class HilPlugin : public plugin::PluginBase {
...
+		sensor.xgyro = gyro.x();
+		sensor.ygyro = gyro.y();
+		sensor.zgyro = gyro.z();
+		sensor.xmag = mag.x();
+		sensor.ymag = mag.y();
+		sensor.zmag = mag.z();
+		// [[[end]]] (checksum: e4bb03f33b73db75bc2d5f1c7595e737)
+		sensor.abs_pressure = req->abs_pressure.fluid_pressure * PASCAL_TO_MILLIBAR;
+		sensor.diff_pressure = req->diff_pressure.fluid_pressure * PASCAL_TO_MILLIBAR;
+		sensor.pressure_alt = req->fields_updated;
```
vooon: Wtf? Looks like error.

mavros/src/plugins/actuator_control.cpp
```diff
@@ -35,25 +35,45 @@ class ActuatorControlPlugin : public plugin::PluginBase {
 	{
 		PluginBase::initialize(uas_);
 
+		actuator_control_target_pub = nh.advertise<mavros_msgs::ActuatorControl>("actuator_control", 10);
 		actuator_control_sub = nh.subscribe("actuator_control", 10, &ActuatorControlPlugin::actuator_control_cb, this);
```
vooon: Again same topic name of subscriber and publisher!

vooon: Lets be similat ro setpoint_raw: actuator_control - sub; target_actuator_control - pub

mavros/src/plugins/hil.cpp
```diff
@@ -0,0 +1,356 @@
...
+		case 10: rcin.chan10_raw	= req->channels[ 9];
+		case  9: rcin.chan9_raw		= req->channels[ 8];
+		case  8: rcin.chan8_raw		= req->channels[ 7];
+		case  7: rcin.chan7_raw		= req->channels[ 6];
+		case  6: rcin.chan6_raw		= req->channels[ 5];
+		case  5: rcin.chan5_raw		= req->channels[ 4];
+		case  4: rcin.chan4_raw		= req->channels[ 3];
+		case  3: rcin.chan3_raw		= req->channels[ 2];
+		case  2: rcin.chan2_raw		= req->channels[ 1];
+		case  1: rcin.chan1_raw		= req->channels[ 0];
```
vooon: Please check: should we set some special value for free channles or not?

mavros_msgs/msg/HilGPS.msg
```diff
@@ -0,0 +1,17 @@
+# HilControls.msg
+#
+# ROS representation of MAVLink HIL_GPS
+# See mavlink message documentation here:
+# https://pixhawk.ethz.ch/mavlink/#HIL_GPS
+
+std_msgs/Header header
+uint8 fix_type
+geographic_msgs/GeoPoint geo
```
vooon: I still don't see a reason for that. I think better to replace with:
```
float64 latitude
float64 longitude
float32 altitude # AMSL
```

mavros_msgs/msg/HilSensor.msg
```diff
@@ -0,0 +1,17 @@
...
+
+std_msgs/Header header
+
+geometry_msgs/Vector3 acc
+geometry_msgs/Vector3 gyro
+geometry_msgs/Vector3 mag
+sensor_msgs/FluidPressure abs_pressure
+sensor_msgs/FluidPressure diff_pressure
+float32 pressure_alt
+sensor_msgs/Temperature temperature
```
vooon: Replace all that sensor_msgs:
```
float32 abs_pressure
float32 diff_pressure
float32 temperature
```

mavros_msgs/msg/HilStateQuaternion.msg
```diff
@@ -0,0 +1,12 @@
+# HilStateQuaternion.msg
+#
+# ROS representation of MAVLink HIL_STATE_QUATERNION
+# See mavlink message documentation here:
+# https://pixhawk.ethz.ch/mavlink/#HIL_STATE_QUATERNION
+
+std_msgs/Header header
+sensor_msgs/Imu imu
+geometry_msgs/Vector3 linear_velocity
+geographic_msgs/GeoPoint geo
```
vooon: Also not needed. Perhaps better msg def:
```
std_msgs/Header header

geometry_msgs/Quaternion orientation
geometry_msgs/Vector3 angular_speed
geometry_msgs/Vector3 linear_acceleration
geometry_msgs/Vector3 linear_speed
float64 latitude
float64 longitude
float32 altitude
float32 ind_airspeed
float32 true_airspeed
```

mavros_msgs/msg/HilGPS.msg
mzahana: I agree here. This also removes the dependency on ```geographic-info``` packages.

mavros_msgs/msg/HilStateQuaternion.msg
TSC21: Why are you always asking for refactoring this? Me and @mzahana already tested this with V-REP sim and it is working

mavros_msgs/msg/HilGPS.msg
TSC21: @vooon you were the first to suggest this. Then you saw the AMSL thing. But ok.
I will chang this later have stuff to do now

mavros_msgs/msg/HilSensor.msg
TSC21: Changing thia implies changing everything on sim side. A bunch of work again because of a detail

mzahana:
I tested this in V-REP. It looks fine in the attitude. I can fly in Manual mode with RC. However, there is something wrong with height data, probably position, too. In the old PR, it was working. But now, the altitude keeps fluctuating with very large step. For example, if the vehicle is on the ground, it fluctuates between 0 and -3000. see attached pic.
hil_altitude_jumps

mzahana:
Here is the jumps in local_pos_x
hil_localx_jumps

mavros/src/plugins/hil.cpp
TSC21: What do you mean?

TSC21:
Those jumps are conversion related (*1E7 on lat and lon). You need to check that you are not applying that directly to local coordinates instead

mavros_msgs/msg/HilGPS.msg
vooon: @TSC21 i suggested to make our `mavros_msgs/GeoPointInt`, not usage of geographinc-info.
Something like:
```
uint32 latitude_1e7
uint32 longitude_1e7
uint32 altitude_mm
```

That msg was smaller than two float64 + one 32, but usage of generic lla fields looks more intuitive.

mavros/src/plugins/hil.cpp
vooon: I mean that if unused channel should be set to UINT16_MAX that approach is not good enough.
If 0 - is ok, then that solution is ok.

mavros_msgs/msg/HilStateQuaternion.msg
vooon: Obviously, because of very bad temper :)

mzahana:
I was able to get rid of the fluctautions. I just implemented the code in ```mocap_fake_gps.cpp``` plugin on the sim side.

The problem now is that height is not correct. If I fly to high altitude, say 20m, it increases slowly then can go back to snall altitude, close to zero.

Also, sometimes,  there is slow convergence in attitude. I am not sure if it is latency issue or px4 firmware issue. Any hints?

TSC21:
> The problem now is that height is not correct. If I fly to high altitude, say 20m, it increases slowly then can go back to small altitude, close to zero.

Do you have a video where we can check this behavior?

> Also, sometimes, there is slow convergence in attitude. I am not sure if it is latency issue or px4 firmware issue. Any hints?

Same as above. Also a log would help

mavros/src/plugins/hil.cpp
```diff
@@ -307,41 +307,39 @@ class HilPlugin : public plugin::PluginBase {
...
 						"FCU receives %lu RC channels, but HIL_RC_INPUTS_RAW can store %zu",
 						req->channels.size(), MAX_CHANCNT);
-
-			channels_count = MAX_CHANCNT;
 		}
 
+		std::array<uint16_t, MAX_CHANCNT> channels;
+		size_t n = req->channels.size() < MAX_CHANCNT ? : MAX_CHANCNT;
+		std::copy(req->channels.begin(), req->channels.begin() + n, channels.begin());
+		std::fill(channels.begin() + n, channels.end(), UINT16_MAX);
```
vooon: Remove `if` at line 314, not needed.
line 321 have bug in limit condition (see my test code). Replace with:

```cpp
size_t n = (req->channels.size() < channels.size()) ? req->channels.size() : channels.size();
```

Test code:
```cpp
#include <iostream>
#include <array>
#include <limits>

int main(int argc, char *argv[])
{
	int a = 0, v;
	std::array<uint16_t, 12> arr{};

	if (argc > 1)
		a = atoi(argv[1]);

	v = (a < arr.size()) ? a : arr.size();

	std::cout << "v: " << v << std::endl;

	auto b = arr.begin();
	std::cout << "b: " << b << " b+v: " << b+v << " is end: " << (b+v == arr.end()) << std::endl;

	std::fill(arr.begin() + v, arr.end(), std::numeric_limits<decltype(arr)::value_type>::max());

	std::cout << "arr:";
	for (auto av : arr)
		std::cout << " " << av;
	std::cout << std::endl;

	return 0;
}
```

mavros/src/plugins/hil.cpp
```diff
@@ -295,7 +295,7 @@ class HilPlugin : public plugin::PluginBase {
 		of.integrated_zgyro = int_gyro.z();
 		of.time_delta_distance_us = req->time_delta_distance_us;
 		of.distance = req->distance;
-		of.quality  = req->quality;
+		of.quality  = req->quality ;
```
vooon: Modification in generated code looks like mistake. Cog will fail here :)
Use `cog.py -xr <file>` that usual `cog.py -cr <file>`

mavros/src/plugins/hil.cpp
vooon: Note: for some reason with that cond worked fine:
```cpp
v = a < 12 ? : 12;
```
But as soon i changed constants with method (which is anyway reduced to constant) it stops, v = 1 or 12.

vooon: Or better to use `std::min()` from `#include <algorithm>`.
That template fn require same type for a & b, but anyway size() is always std::size_t.

```cpp
auto n = std::min(req->channels.size(), channels.size());
```

TSC21:
@mzahana do you have updates on the sim?

mzahana:
Sorry, not yet! I am having piles of work! I will try to steal sometime end
of today, hopefully!

On Apr 12, 2017 2:21 PM, "Nuno Marques" <notifications@github.com> wrote:

@mzahana <https://github.com/mzahana> do you have updates on the sim?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/mavlink/mavros/pull/684#issuecomment-293547641>, or mute
the thread

TSC21:
> Sorry, not yet! I am having piles of work! I will try to steal sometime end of today, hopefully!

Ok np let us know some more results and a log/video would help so we are able to merge this finally.

mzahana:
I am still not able to solve the attitude/altitude convergence issue. I tried latest master with LPE and EKF2, but had same problems. One question regarding altitude: in the ```mavros_msgs/msg/HilSensor.msg ``` msg there is the ```pressure_alt``` field. Should this be simply the altitude from the simulator, or some other value. Currently, I am just setting this to be the altitude of the vehicle computed in the sim.

I made some video and have some log. What is the fastest way to share the log with you?

mzahana:
Here is a link to the video and log,
https://www.dropbox.com/sh/hj0jbpynfh5008h/AACLqYjOBU203J7bsbncCO9sa?dl=0

TSC21:
one thing: change the `SYS_COMPANION` parameter on QGC to 921600 instead of 115200 and apply that baud on the connection to the FCU. This will guarantee a faster data transfer and possibly even fix the convergence problem.

TSC21:
> Should this be simply the altitude from the simulator, or some other value?
Currently, I am just setting this to be the altitude of the vehicle computed in the sim.

The `pressure_alt` field is supposed to be the altitude calculated from the barometer (see http://mavlink.org/messages/common#HIL_SENSOR). So unless you have a simulated pressure sensor on the sim, this value should be close or equal to `local_pos_ned.z`.

mzahana:
I am not actually using TELEM2. Iam using normal usb port

TSC21:
Well the throughput like that will be lower. Don't you have USB<->UART converter you can use?

TSC21:
Btw this https://github.com/PX4/Firmware/blob/ecb2511a7bbad094bff23dd95036eda0656778f0/src/modules/simulator/simulator_mavlink.cpp#L215 is the function that matters on the FCU side in terms of receiving HIL sensor data. As you can see, there seems to be no problem it terms of sensor pass-trough.

TSC21:
Oh I see that I forgot to add `*100` here https://github.com/mavlink/mavros/pull/684/files#diff-d6d9379eca06dca47c809d24983a3e28R151. Will fix it

TSC21:
One thing I can't understand is this: https://github.com/PX4/Firmware/blob/ecb2511a7bbad094bff23dd95036eda0656778f0/src/modules/simulator/simulator_mavlink.cpp#L407. The altitude is divided again by `1E3` but the lat and lon are not divided by `1E7`.

TSC21:
Another thing: https://github.com/PX4/Firmware/blob/ecb2511a7bbad094bff23dd95036eda0656778f0/src/modules/simulator/simulator_mavlink.cpp#L354. We can see that `HIL_GPS` is not even published as a UORB topic, so I it only depends on the `HIL_STATE_QUATERNION` msg.

TSC21:
@mzahana please pull and try latest commit.

mzahana:
@TSC21 

>  We can see that ```HIL_GPS``` is not even published as a UORB topic, so I it only depends on the ```HIL_STATE_QUATERNION``` msg.

That could be one issue. I was only using HIL_SENSOR and HIL_GPS. I had ```HIL_STATE_QUATERNION``` implemented in the sim, but not published yet. For some reason, I switched to ```HIL_GPS```. However, what confuses me is that the number of Satellites that is set in the ```HIL_GPS``` msg appears in QGC. So, some how, it is received and interpreted.

Anyway, I will try ```HIL_STATE_QUATERNION``` and see if it makes difference.

Next, I will try USB<-> UART cable. However, I used to work with jMavSim+Pixhawk with USB cable, and it was working fine. I will keep you posted.

mzahana:
@TSC21 

> One thing I can't understand is this: https://github.com/PX4/Firmware/blob/ecb2511a7bbad094bff23dd95036eda0656778f0/src/modules/simulator/simulator_mavlink.cpp#L407. The altitude is divided again by 1E3 but the lat and lon are not divided by 1E7.

I agree! In the MAVLink description of ```HIL_STATE_QUATERNION ```, [here] (https://pixhawk.ethz.ch/mavlink/#HIL_STATE_QUATERNION), lat/lon are assumed to be multiplied by ```1E7```. However, it is used as is on the px4 side! I am not sure if it is a bug, or intended.

mzahana:
According to the msg definition on px4 side https://github.com/PX4/Firmware/blob/912ed98a28022504ef3ec4bf1bce0fe041f83ac4/msg/vehicle_global_position.msg
lat/lon/alt should be in deg/deg/meters, respectively.
Therefore, https://github.com/PX4/Firmware/blob/ecb2511a7bbad094bff23dd95036eda0656778f0/src/modules/simulator/simulator_mavlink.cpp#L407-L409 should all be scaled down appropriately. I think an issue should be opened in PX4.

mzahana:
There is a mistake here,
https://github.com/TSC21/mavros/blob/ea439de354dde3154babc8ce7fee9623f7b763e3/mavros/src/plugins/hil.cpp#L138
it should be
```state_quat.lat = req->geo.latitude * 1E7``` not ```state_quat.lat = req->geo.altitude * 1E7```

TSC21:
> There is a mistake here,
https://github.com/TSC21/mavros/blob/ea439de354dde3154babc8ce7fee9623f7b763e3/mavros/src/plugins/hil.cpp#L138
it should be
state_quat.lat = req->geo.latitude * 1E7 not state_quat.lat = req->geo.altitude * 1E7

Fixed

mzahana:
@TSC21 I don't think that the ```simulator``` [module](https://github.com/PX4/Firmware/tree/master/src/modules/simulator) is running in the first place. I did not find it in NSH, and it is not compiled either, see the [config file](https://github.com/PX4/Firmware/blob/master/cmake/configs/nuttx_px4fmu-v2_default.cmake).
So, I guess it is not used in HIL!

My guess is that mavlink messages are handled in the [mavlink module](https://github.com/PX4/Firmware/blob/master/src/modules/mavlink/mavlink_receiver.cpp#L307-L337)

Also, I saw [this diagram](https://github.com/PX4/Firmware/tree/master/posix-configs/SITL) that explains the SITL/HIL setup (for JMAVSim), and ```HIL_STATE_QUATERNION``` is not mentioned there!

I guess, whoever worked on the HIL with jMAVSim could help here.

mhkabir:
@mzahana `simulator` is only used for SITL. For HIL, yes, the messages are handled in mavlink_receiver.

mzahana:
Could not figure out the problem. I still get convergence issues. May some one who worked on HIL can help here.

TSC21:
Well SITL using HIL msgs is kinda awkward and led me to think that `simulator` was being used to decode and HIL msgs. Well but nothing was lost since we were able to find and correct a bug on `simulator_mavlink.cpp`. Will continue evaluating the code to find some flaw that may lead to your problem. Btw @mzahana have you already tested using an higher baud-rate trough TELEM?

TSC21:
https://github.com/PX4/Firmware/blob/b2d47adf56cc2157659945afd241aeffdec320b3/src/modules/mavlink/mavlink_receiver.cpp#L1921

No conversion again? @mhkabir ?

TSC21:
@mzahana updated the code to have the right conversions according to the mavlink msg. Please remove all the conversions you have in your sim Lua code so we have a proper match here.

mzahana:
I don't think it's wrong here,
https://github.com/PX4/Firmware/blob/b2d47adf56cc2157659945afd241aeffdec320b3/src/modules/mavlink/mavlink_receiver.cpp#L1921

As the message definition for ```vehicle_gps_position``` matches, see [here](https://github.com/PX4/Firmware/blob/912ed98a28022504ef3ec4bf1bce0fe041f83ac4/msg/vehicle_gps_position.msg)

TSC21:
> 
I don't think it's wrong here,
https://github.com/PX4/Firmware/blob/b2d47adf56cc2157659945afd241aeffdec320b3/src/modules/mavlink/mavlink_receiver.cpp#L1921
As the message definition for vehicle_gps_position matches, see here

Yeah didn't saw that.

A detail: [this](https://github.com/PX4/Firmware/blob/b2d47adf56cc2157659945afd241aeffdec320b3/src/modules/mavlink/mavlink_receiver.cpp#L2143) shouldn't this be `hil_local_pos.z = -(_hil_local_alt0 + hil_state.alt / 1000.0f);`

mzahana:
@TSC21 why you are multiplying ```cog``` by ```1E-1```?
According to MAVLink def [here](https://pixhawk.ethz.ch/mavlink/#HIL_GPS), it's *degrees\*100*. So, should be ```req->cog * 1E2```, no?

mzahana:
Also, I am so confused! Do I need to use ```HIL_STATE_QUATERNION``` or NOT?! If ```HIL_GPS``` is used, why the need for the other one?

mzahana:
How should ```xacc, yacc, zacc``` be published? are these the local accelerations, or accelerometer data?
What is the unit?
https://pixhawk.ethz.ch/mavlink/#HIL_STATE_QUATERNION

TSC21:
> @TSC21 why you are multiplying cog by 1E-1?
According to MAVLink def here, it's degrees*100. So, should be req->cog * 1E2, no?

Yeah forget that I thought that the initial `M_` was for mili but it is for macro. Will fix

mzahana:
Update: I am now publishing ```HIL_STATE_QUATERNION```, in addition to ```HIL_GPS``` and ```HIL_SENSORS```.
The altitude looks good. However, the roll/pitch axes are switched. In other words, If I don’t use ```HIL_STATE_QUATERNION```, but use ```HIL_GPS```, then use my RC to fly the quad, It flies fine. If I publish ```HIL_STATE_QUATERNION```, they get switched. This, I guess, is a transformation issue. I publish gyros/accel to ROS a: x-forward, y-left, z-up. This is ROS convention as described [here](http://www.ros.org/reps/rep-0103.html).

TSC21:
> Also, I am so confused! Do I need to use HIL_STATE_QUATERNION or NOT?! If HIL_GPS is used, why the need for the other one?

`HIL_GPS` doesn't give you the attitude. And from my evaluation, seems to me that `STATE_QUATERNION` is the one that gives most of the info to the FCU, including vehicle global and local position, while the data in `HIL_GPS` is only used to update `vehicle_gps_position`.

TSC21:
> How should xacc, yacc, zacc be published? are these the local accelerations, or accelerometer data?
What is the unit?
https://pixhawk.ethz.ch/mavlink/#HIL_STATE_QUATERNION

Those are accelerations. Accel data goes on `HIL_SENSOR`.

mzahana:
> Those are accelerations. Accel data goes on HIL_SENSOR.

You mean the local frame acceleration, e.g. derivative of linear velocities?

TSC21:
> You mean the local frame acceleration, e.g. derivative of linear velocities?

I suppose so.

TSC21:
Btw what's the status on the lat, lon and alt / local positions on QGC? Are they ok?

mzahana:
> Btw what's the status on the lat, lon and alt / local positions on QGC? Are they ok?

No, there is something weird. If I move the quad in the sim, only lat changes. Lon dos not change!

Also, when I start the quad, the heading is supposed to point to North. However, it points to East, and this is when I publish ```HIL_STATE_QUATERNION```. So, I guess there is something wrong in the transformation~

TSC21:
> So, I guess there is something wrong in the transformation~

In which side? sim or mavros?

mzahana:
I have no idea! There are few things that I am not sure about:
* what is exactly expected to published here
https://github.com/TSC21/mavros/blob/hil_plugins/mavros_msgs/msg/HilStateQuaternion.msg#L8
and what coordinates. Specifically, the acceleration ```xacc/yacc/zacc``` in the
https://pixhawk.ethz.ch/mavlink/#HIL_STATE_QUATERNION
* what is the quaternion to be published, in which coordinate?

TSC21:
>what is exactly expected to published here
https://github.com/TSC21/mavros/blob/hil_plugins/mavros_msgs/msg/HilStateQuaternion.msg#L8
and what coordinates. Specifically, the acceleration xacc/yacc/zacc in the
https://pixhawk.ethz.ch/mavlink/#HIL_STATE_QUATERNION

Accel: https://github.com/PX4/Firmware/blob/b2d47adf56cc2157659945afd241aeffdec320b3/src/modules/mavlink/mavlink_receiver.cpp#L2255-L2260. So as the conversion on mavros is done from baselink frame to aircraft (body frame), you need to publish the accel on baselink frame. Try for one time to remove the conversion [here](https://github.com/mavlink/mavros/pull/684/files#diff-d6d9379eca06dca47c809d24983a3e28R156) and see how it goes. Also the quaternion is relative to the baselink frame on ROS, so mavros converts from baselink to aicraft frame after converting from enu to ned orientation.

mzahana:
I dont know why there is this 90 degrees shift in heading, seen in QGC!! I could not git rid of it!
The heading should point North on start, but it points East!

TSC21:
> I dont know why there is this 90 degrees shift in heading, seen in QGC!! I could not git rid of it!
The heading should point North on start, but it points East!

Ok we will try to figure that out? What about the lat, lon and altitude? Do you have the altitude already ok?

TSC21:
Have you tried to remove [this](https://github.com/mavlink/mavros/pull/684/files#diff-d6d9379eca06dca47c809d24983a3e28R130)?

TSC21:
> A detail: this shouldn't be hil_local_pos.z = -(_hil_local_alt0 + hil_state.alt / 1000.0f); instead?

Also have you tried this out?

mhkabir:
> I dont know why there is this 90 degrees shift in heading, seen in QGC!! I could not git rid of it!
The heading should point North on start, but it points East!

You're missing a transform in a global-frame quaternion somewhere. The fix should be like this : https://github.com/mavlink/mavros/pull/660/commits/f4de2b1ad34ae5071a2d8de7437d12eb6fceff2f. Apply it wherever needed.

TSC21:
> I dont know why there is this 90 degrees shift in heading, seen in QGC!! I could not git rid of it!
The heading should point North on start, but it points East!
You're missing a transform in a global-frame quaternion somewhere. The fix should be like this : f4de2b1. Apply it wherever needed.

https://github.com/mavlink/mavros/pull/684/files#diff-d6d9379eca06dca47c809d24983a3e28R130

TSC21:
@mzahana about the East orientation:
```bash
geo: 
  latitude: 5.0
  longitude: 5.0
  altitude: 0.0150001849979
eph: 2
epv: 2
vel: 0
vn: 0
ve: 0
vd: 0
cog: 90
satellites_visible: 10
---

You see that the current `cog` is with a 90 degrees rotation relative to the world, so that means it is facing East ;)

TSC21:
@mhkabir lgtm?

mhkabir:
I will review fully tomorrow. I have the day off, so should be done quickly.

TSC21:
> I will review fully tomorrow. I have the day off, so should be done quickly.

Yeah it should :D

mhkabir:
@TSC21 Please squash the commits using rebase. Way too many commits at the moment.

TSC21:
> @TSC21 Please squash the commits using rebase. Way too many commits at the moment.

I don't want to "kill" @mzahana commits. Any ways of doing that without cleaning his commits?

mhkabir:
Yes, a rebase won't kill his commits. It'll show @mzahana as author and @TSC21 as rebaser/author. 

Use something like `git rebase -i HEAD~60` and then choose and squash.

TSC21:
@mhkabir it is not letting me squash rebase because of master merges...

mhkabir:
I'm reviewing now. Can one of you document this properly and send in a pull to https://github.com/PX4/Devguide? Will only merge here after docs are in.

TSC21:
> I'm reviewing now. Can one of you document this properly and send in a pull to https://github.com/PX4/Devguide? Will only merge here after docs are in.

https://mzahana.gitbooks.io/indoor-nav-at-risclab/content/setup_hil_with_px4_and_v-rep.html @mzahana do you have some time to update this according to our implementation. @mhkabir I can merge it myself no worries just let me know about any tip on code.

mzahana:
I prefer to document, if and only if everything works end to end.
Otherwise, documentation would be loose. It would be more complete if it is
tested end to end. I also agree to merge after documentation.
Unfortunately, I have deadlines that wont allow me to revisit this soon, as
it is becoming time consuming for me to test and find the bugs. So, I will
go on a slow pace the coming days. Tanks.

TSC21:
I agree on docs but this is stopping the dev of other HIL related issues - example: https://github.com/ethz-asl/rotors_simulator/issues/356 and https://github.com/ethz-asl/rotors_simulator/issues/380

TSC21:
Having this merged will allow to test the plugins with gazebo using rotors. That will be useful for document. But if @mzahana doesn't have time to test/document, this will be staled.

TSC21:
> Unfortunately, I have deadlines that wont allow me to revisit this soon, as
it is becoming time consuming for me to test and find the bugs.

@mzahana please list the bugs/issues you are getting here on the comment list so they can be checked. Thanks

mzahana:
Can somebody who worked on HIL with jMAVSim step in and help here as that setup is supposed to be working. That can help finalizing this PR faster.

TSC21:
@bkueng, @julianoes, @v01d, @kd0aij maybe you can lend an hand here please? This is ROS but also HIL related, the problem seems to be related to some rotation, probably coming from the sim side. @mzahana can you please post here your latest Lua script of V-REP so it can be evaluated by all?

TSC21:
@mzahana did you get a time frame to test V-REP plugin again and solve the problem?

mzahana:
Unfortunately, not yet. And it does not seem to be soon as I have several
deadlines that are eating my whole time. Maybe by the end of this month or
early next month, I will have some time to debug.

TSC21:
@mzahana something interesting about the supposed problem you were getting on the vehicle facing East - you can read on ROS REP 103:

> By the right hand rule, the yaw component of orientation increases as the child frame rotates counter-clockwise, and for geographic poses, yaw is zero when pointing east.

So this is not a problem at all, more like a question of convention which results from applying the transforms from NED<->ENU and aircraft<->base_link.

TSC21:
More:
> This requires special mention only because it differs from a traditional compass bearing, which is zero when pointing north and increments clockwise. Hardware drivers should make the appropriate transformations before publishing standard ROS messages.

TSC21:
I tested this on V-REP and seemed consistent after some V-REP mods on the Lua script. I'm merging it now, having pendent adding documentation to the PX4 Devguide on how to setup the V-REP model (if someone wants to do that for Ardupilot let me know and I can put it available to it also).
Also, I'll be working on interfacing this with Gazebo, which will can also provide some improvements to the HIL interface of `rotors_simulator`. Will have it available soon.

Issue #371, opened 2015-08-10
Add support for HOME_POSITION


vooon:
Introduced: https://github.com/mavlink/mavlink/commit/0c16a3841e8865982800ef9e03e04f9f002db417

tstastny:
@vooon Was this ever put into a plugin? I did not find it. If not, I was planning to do this anyway and could make a pull request. Was thinking it could be its own plugin, as it's not really a waypoint, and not an aircraft position. Let me know if you disagree. Otherwise I'll start today.

vooon:
@tstastny write new plugin.

TSC21:
@tstastny have you implemented this already? If yes, issuing a PR wourld be awesome. Maybe integrate it on `global_position` plugin?

tstastny:
@TSC21 I actually did implement, but on my group's own fork of MAVROS. In December we were quite behind the current master. Now we have caught up. I could make a PR in the next couple of days. (Note I've implemented as a new plugin specifically for `home_position`, as I thought it best to separate a home point from vehicle specific positions in `global_position`).

TSC21:
Yeah makes sense as it also includes the local position. Will be waiting for your PR then. Thank you!

vooon:
Implemented in #695.

PR #688, opened 2017-04-06
HomePositionPlugin: add Plugin to publish current home position

tstastny:
Plugin for publishing home position. Addresses #371.

### DIFF
--- a/mavros/src/plugins/home_position.cpp
+++ b/mavros/src/plugins/home_position.cpp
@@ -1,0 +1,104 @@
+/**
+ * @brief HomePosition plugin 
+ * @file home_position.cpp
+ * @author Thomas Stastny <thomas.stastny@mavt.ethz.ch>
+ *
+ * @Plugin to publish home position mavlink messages in ROS
+ * @addtogroup plugin
+ * @{
+ */
+/*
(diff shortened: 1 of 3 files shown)


mavros/src/plugins/home_position.cpp
```diff
@@ -0,0 +1,104 @@
...
+ */
+
+#include <list>
+
+#include <mavros/mavros_plugin.h>
+#include <pluginlib/class_list_macros.h>
+
+#include <mavros_msgs/HomePosition.h>
+
+namespace mavplugin {
```
TSC21: You are not applying new plugin API. Please check how currently the plugins are implemented and update please

TSC21: Ow now I see: you are applying this to the `indigo-devel`. Please upgrade to Kinetic and launch the PR against master. `indigo-devel` is deprecated.

TSC21:
@tstastny I'm closing this PR 'cause you are applying it to the `indigo-devel`, which we are not currenlty maintaining. Please upgrade do Kinetic and launch PR against `master`. Also add a use case where you applied this plugin (like a video or a log) where we can see this as tested. Thanks!

PR #694, opened 2017-04-15
Add HomePosition plugin

TSC21:
This is a redo of #688 and adjusted to current API. I removed `HomePosition.msg` as I think we can publish the different info in separated topics instead of adding a new msg.

### DIFF
--- a/mavros/src/plugins/safety_area.cpp
+++ b/mavros/src/plugins/safety_area.cpp
@@ -36,6 +36,8 @@ public:
 	{
 		PluginBase::initialize(uas_);
 
+		safety_nh.param<std::string>("frame_id", frame_id, "safety_area");
+
 		bool manual_def = false;
 		double p1x, p1y, p1z,
 			p2x, p2y, p2z;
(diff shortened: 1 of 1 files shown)


TSC21:
It still needs that `geographic_msgs` get updated on Rosdistro (done according to https://github.com/ros-geographic-info/geographic_info/commit/b8043a92e91f38aa09b65b6b31aee96077996bb6#commitcomment-21775820)

vooon:
That is BAD! With that changes we can't build master on Indigo!

TSC21:
> That is BAD! With that changes we can't build master on Indigo!

You mean remove `HomePosition.msg`? I can re-add it or make a pr to `indigo-devel` removing it. It wasn't being used anyway.

TSC21:
We can find a intermediate solution if https://github.com/ros-geographic-info/geographic_info/commit/b8043a92e91f38aa09b65b6b31aee96077996bb6#commitcomment-21777839 doesn't happen. But add a note on Readme to git pull the latest `geographic_msgs` would be a better solution as we will want to progress on the development of newer platforms, even though
 we will want to do minimal maintenance of legacy releases

PR #695, opened 2017-04-16
Add HomePosition plugin

TSC21:
Handles `HOME_POSITION` and `SET_HOME_POSITION` Mavlink messages. Decided to redo #694 and keep `HomePositiom.msg` as I detected there is a `SET_HOME_POSITION` Mavlink msg, so then I added a callback to send it to the stream.


TSC21:
As in #694, waiting for https://github.com/ros-geographic-info/geographic_info/commit/b8043a92e91f38aa09b65b6b31aee96077996bb6#commitcomment-21778673

mavros/src/plugins/home_position.cpp
```diff
@@ -0,0 +1,140 @@
...
+namespace std_plugins {
+/**
+ * @brief home position plugin.
+ *
+ * Publishes home position.
+ */
+class HomePositionPlugin : public plugin::PluginBase {
+public:
+	HomePositionPlugin() :
+		hp_nh("~")
```
vooon: Should be `nh("~home_position")`

mavros/src/plugins/home_position.cpp
```diff
@@ -0,0 +1,140 @@
...
+	{ }
+
+	void initialize(UAS &uas_)
+	{
+		PluginBase::initialize(uas_);
+
+		home_position_pub = hp_nh.advertise<mavros_msgs::HomePosition>("get",10);
```
vooon: Why "get"? And perhaps that should be storing topic like `~mission/waypoints`.
Should some COMMAND_LONG be sent to request HOME_POSITION?

TSC21: > Why "get"? And perhaps that should be storing topic like `~mission/waypoints`.

Since there's only one `home_position`, do you think it makes sense to store it?

> Should some COMMAND_LONG be sent to request HOME_POSITION?

Yes there is. I will implement it.

mavros/src/plugins/home_position.cpp
TSC21: ok

TSC21:
Added `COMMAND_LONG` to request for `HOME_POSITION`.

TSC21:
I removed the baselink<-> aicraft transform as in https://github.com/mavlink/mavros/pull/695:
`q	float[4]	World to surface normal and heading transformation of the takeoff position. Used to indicate the heading and slope of the ground`

mavros/src/plugins/home_position.cpp
vooon: https://github.com/mavlink/mavros/blob/master/mavros/src/plugins/waypoint.cpp#L151 did you see that `true`?
http://docs.ros.org/kinetic/api/roscpp/html/classros_1_1NodeHandle.html#a6b655c04f32c4c967d49799ff9312ac6

So yes, it is not storing, it is latching publisher.

mavros_msgs/CMakeLists.txt
```diff
@@ -40,7 +40,8 @@ add_service_files(
   DIRECTORY srv
   FILES
   CommandBool.srv
-  CommandHome.srv
+  CommandSetHome.srv
+  CommandGetHome.srv
```
vooon: Do not approve. You don't need to change existing API.

mavros/src/plugins/command.cpp
```diff
@@ -333,6 +334,16 @@ class CommandPlugin : public plugin::PluginBase {
...
+	bool get_home_cb(mavros_msgs::CommandGetHome::Request &req,
+			mavros_msgs::CommandGetHome::Response &res)
+	{
+		using mavlink::common::MAV_CMD;
+		return send_command_long_and_wait(false,
+				enum_value(MAV_CMD::GET_HOME_POSITION), 1,
+				0, 0, 0, 0, 0, 0, 0,
+				res.success, res.result);
+	}
+
```
vooon: I don't see why that shortcut is needed? Why not just call generic service?

mavros_msgs/CMakeLists.txt
vooon: Also that is legacy, since you implemented new topic.

TSC21: What do you suggest then? Eliminate the srv call?

mavros/src/plugins/command.cpp
TSC21: What do you mean?

mavros/src/plugins/home_position.cpp
TSC21: Oh it's a latch! Thanks! So should I create a new msg like `HomePositionCurrent.msg`?

TSC21:
> I expect, that after connection was changed you will start timeout timer, which request home several times (maybe w/o limit since obtaining GPS lock may take a while).
Then when you got response you will stop timer.
Eg:
Got CON -> poll request -> got home
Lost CON -> stop poll

Can you exemplify what you mean here please?

vooon:
Will be rebased manually.

TSC21:
thanks

Issue #396, opened 2015-10-09
[setpoint_attitude] Don't work with APM


vooon:
For now this issue is a note to remember.

APM implementation PR: https://github.com/diydrones/ardupilot/pull/2959

vooon:
Possible way to solve this is to use Message Filters to syncronize two subscribers and construct one SET_ATTITUDE_TARGET with quaternion and thrust.

CrazyPopLin:
hi @vooon ，does the setpoint works with APM right now. Thanks.

vooon:
Attitude - nope.

TSC21:
Closing this as this in not MAVROS related.

TSC21:
Closing as solved. APM supports attitude setpoints now. Just use `setpoint_raw` plugin while https://github.com/mavlink/mavros/pull/700 is not merged.

PR #699, opened 2017-05-01
launch: remove setpoint-attitude from apm blacklist

rmackay9:
Ardupilot supports the SET_ATTITUDE_TARGET mavlink message so the setpoint-attitude should be removed from the blacklist.  This has been tested by a user.

### DIFF
--- a/mavros/launch/apm_pluginlists.yaml
+++ b/mavros/launch/apm_pluginlists.yaml
@@ -3,7 +3,6 @@ plugin_blacklist:
 - actuator_control
 - ftp
 - safety_area
-- setpoint_attitude
 # extras
 - image_pub
 - 'mocap_*'


TSC21:
Dear @rmackay9 thanks for pointing that out. Do you have a reference/link where we can check that and its results? Thanks

TSC21:
Ok I checked some references and seems the support exists. For me is ok to merge.

rmackay9:
Thanks. yes there's a guy on the ardupilot gitter chat who has tested it.  The link to that chat is here with links to all our various channels of communication (quite a few!). http://ardupilot.org/copter/docs/common-contact-us.html

We do have a small issue in our support that we need to allow accepting throttle and attitude separately but it's still possible to have ROS send both together so it can work.

TSC21:
> We do have a small issue in our support that we need to allow accepting throttle and attitude separately but it's still possible to have ROS send both together so it can work.

We actually have two different callbacks: one for attitude and one for throttle that we send separately, ate least in `setpoint_attitude`. For `setpoint_raw` there's the possibility of sending both together. I will probably add this option to `setpoint_attitude` plugin also.

rmackay9:
Yes, it's the setpoint-raw that works ok with ardupilot.

TSC21:
So if it's `setpoint_raw` that is working, makes sense to whitelist the `setpoint_attitude``plugin for APM? Or do you have a way of handling it separately?

vooon:
That was a long time todo, to use message filter to synchronize topics and send one message #396.

TSC21:
@vooon I can implement it. I already used Message Filter once for a project.

rmackay9:
So I think because ardupilot supports the setpoint_raw, it makes sense to take this setpoint-attitude off the blacklist.  i.e. partial functionality is all we need to achieve for it to be useful.  I don't think we need to wait for TSC21's fix (although that would be nice).  So can we merge this in?

TSC21:
Will add Message Filter for sunc today

rmackay9:
@TSC21, thanks very much!

TSC21:
@rmackay9 by sending the `SET_ATTITUDE_TARGET` msg, do you also combine the rpy rates with the thrust or only the quaternion with the thrust?

PR #700, opened 2017-05-02
[WIP] Plugins: setpoint_attitude: add sync between thrust and attitude

TSC21:
- [x] Sync between thrust and pose msg
- [x] Sync between thrust and twist msg
- [x] Sync between thrust and transform

### DIFF
--- a/mavros/include/mavros/setpoint_mixin.h
+++ b/mavros/include/mavros/setpoint_mixin.h
@@ -212,11 +212,11 @@ public:
 	 * @param cbp        plugin callback function
 	 */
 	template <class T>
-	void tf2_start(const char *_thd_name, void (D::*cbp)(const geometry_msgs::TransformStamped &, const typename T::ConstPtr &))
+	void tf2_start(const char *_thd_name, message_filters::Subscriber<T> &topic_sub, void (D::*cbp)(const geometry_msgs::TransformStamped &, const typename T::ConstPtr &))
 	{
 		tf_thd_name = _thd_name;
 
-		tf_thread = std::thread([this, cbp]() {
+		tf_thread = std::thread([this, cbp, &topic_sub]() {
 			mavconn::utils::set_this_thread_name("%s", tf_thd_name.c_str());
 
 			mavros::UAS *m_uas_ = static_cast<D *>(this)->m_uas;
@@ -224,8 +224,7 @@ public:
 			std::string &_frame_id = static_cast<D *>(this)->tf_frame_id;
 			std::string &_child_frame_id = static_cast<D *>(this)->tf_child_frame_id;
 
-			message_filters::Subscriber<T> &_topic_sub = static_cast<D *>(this)->sub;
-			tf2_ros::MessageFilter<T> tf2_filter(_topic_sub, m_uas_->tf2_buffer, _frame_id, 10, _sp_nh);
+			tf2_ros::MessageFilter<T> tf2_filter(topic_sub, m_uas_->tf2_buffer, _frame_id, 10, _sp_nh);
 
 			ros::Rate rate(static_cast<D *>(this)->tf_rate);
 			while (ros::ok()) {
--- a/mavros/include/mavros/setpoint_mixin.h
+++ b/mavros/include/mavros/setpoint_mixin.h
@@ -21,6 +21,11 @@
 #include <mavros/mavros_plugin.h>
 
 #include <geometry_msgs/TransformStamped.h>
+#include <mavros_msgs/Thrust.h>
+
+#include "tf2_ros/message_filter.h"
+#include <message_filters/subscriber.h>
+
 
 namespace mavros {
 namespace plugin {
@@ -162,8 +167,10 @@ public:
 template <class D>
 class TF2ListenerMixin {
 public:
-	std::thread tf_thread;
-	std::string tf_thd_name;
+	std::thread tf_thread_1;
+	std::thread tf_thread_2;
+	std::string tf_thd_name_1;
+	std::string tf_thd_name_2;
 
 	/**
 	 * @brief start tf listener
@@ -173,11 +180,11 @@ public:
 	 */
 	void tf2_start(const char *_thd_name, void (D::*cbp)(const geometry_msgs::TransformStamped &) )
 	{
-		tf_thd_name = _thd_name;
+		tf_thd_name_1 = _thd_name;
 		auto tf_transform_cb = std::bind(cbp, static_cast<D *>(this), std::placeholders::_1);
 
-		tf_thread = std::thread([this, tf_transform_cb]() {
-			mavconn::utils::set_this_thread_name("%s", tf_thd_name.c_str());
+		tf_thread_1 = std::thread([this, tf_transform_cb]() {
+			mavconn::utils::set_this_thread_name("%s", tf_thd_name_1.c_str());
 
 			mavros::UAS *m_uas_ = static_cast<D *>(this)->m_uas;
 			std::string &_frame_id = static_cast<D *>(this)->tf_frame_id;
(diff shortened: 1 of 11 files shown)


mavros/src/plugins/setpoint_accel.cpp
```diff
@@ -81,12 +81,12 @@ class SetpointAccelerationPlugin : public plugin::PluginBase,
 		auto accel = ftf::transform_frame_enu_ned(accel_enu);
 
 		set_position_target_local_ned(stamp.toNSec() / 1000000,
-				utils::enum_value(MAV_FRAME::LOCAL_NED),
-				ignore_all_except_a_xyz,
-				Eigen::Vector3d::Zero(),
-				Eigen::Vector3d::Zero(),
-				accel,
-				0.0, 0.0);
+					utils::enum_value(MAV_FRAME::LOCAL_NED),
```
TSC21: I apllied the uncrustify to the wrong file initially so that's why this changed.

mavros/src/plugins/setpoint_attitude.cpp
```diff
@@ -19,9 +19,13 @@
...
 #include <eigen_conversions/eigen_msg.h>
 
+#include <message_filters/subscriber.h>
+#include <message_filters/synchronizer.h>
+#include <message_filters/sync_policies/approximate_time.h>
+
 #include <geometry_msgs/PoseStamped.h>
 #include <geometry_msgs/TwistStamped.h>
-#include <std_msgs/Float64.h>
+#include <mavros_msgs/Float32Stamped.h>
```
TSC21: This was needed as `Float64` doesn't have a stamp so to sync.

TSC21: Since I'm changing this to a custom msg, maybe I should change to `Thrust.msg` only

TSC21:
Seems like APM is also able to handle RPY together with throttle. Will add it then

TSC21:
@rmackay9 FYI

TSC21:
ping @vooon

TSC21:
@vooon please review. thanks

TSC21:
@vooon this is tested. friendly ping!

mavros/include/mavros/setpoint_mixin.h
```diff
@@ -193,7 +200,46 @@ class TF2ListenerMixin {
...
+		});
+	}
+
+	/**
+	 * @brief start tf listener syncronized with mavros_msgs::Thrust publisher topic
+	 *
+	 * @param _thd_name  listener thread name
+	 * @param cbp        plugin callback function
+	 */
+	void tf2_sync_start(const char *_thd_name, void (D::*cbp)(const geometry_msgs::TransformStamped &, const mavros_msgs::Thrust::ConstPtr &))
```
vooon: I don't like that change. Should be better way.

TSC21: This change took me half day to figure a way to work lol If you have a different proposal that works the same way or better I accept it

TSC21: Besides I'm doing some updates on `mavteleop` that require that this PR is merged to work properly. Also, this change is required to `setpoint_attitude` to work as it should, and I don't think this has to be delayed more

TSC21: The only optimization I can think of is generalize the type of message we are syncing the tf listener with, instead of with the thrust only. That way we have a generalized way of syncing different types of topics with the transform.

TSC21:
@vooon I think right now this is the best that we can get right now (a generic function template that syncs any topic with a transform listener), besides now the templates have the same name but different classes/typenames, which in terms of code structure is better on the plugin side (is not a template friend class but a friend class with function templates). And this is working. So unless you have a better proposal, I think this is good to go.

TSC21:
@vooon ok better now I kept the template for the class and only added the template for the `tf2_start` in case of sync needed. Can you review and merge of ok, please? Thanks

mavros/include/mavros/setpoint_mixin.h
```diff
@@ -162,8 +166,10 @@ class SetAttitudeTargetMixin {
 template <class D>
 class TF2ListenerMixin {
 public:
-	std::thread tf_thread;
-	std::string tf_thd_name;
+	std::thread tf_thread_1;
+	std::thread tf_thread_2;
+	std::string tf_thd_name_1;
+	std::string tf_thd_name_2;
```
vooon: Why you need two threads? Only one `tf_start()` should be used.

mavros/src/plugins/setpoint_attitude.cpp
```diff
@@ -46,26 +51,46 @@ class SetpointAttitudePlugin : public plugin::PluginBase,
 
 		bool tf_listen;
 
+		typedef message_filters::sync_policies::ApproximateTime<geometry_msgs::PoseStamped, mavros_msgs::Thrust> SyncPoseThrust;
+		typedef message_filters::sync_policies::ApproximateTime<geometry_msgs::TwistStamped, mavros_msgs::Thrust> SyncTwistThrust;
```
vooon: Use `using` instead `typedef`.
```cpp
using SyncPoseThrust = mesage_filters::sync_polices::ApproximateTime<geometry_msgs::PoseStamped, mavros_msgs::Thrust>;
```

Also may be good to move out of method, e.g. place in `std_plugins::`.

mavros/src/plugins/setpoint_attitude.cpp
```diff
@@ -46,26 +51,46 @@ class SetpointAttitudePlugin : public plugin::PluginBase,
...
+		sp_nh.param("use_quaternion", use_quaternion, false);
+		sp_nh.param("reverse_thrust", reverse_thrust, false);
 		// tf params
 		sp_nh.param("tf/listen", tf_listen, false);
 		sp_nh.param<std::string>("tf/frame_id", tf_frame_id, "map");
 		sp_nh.param<std::string>("tf/child_frame_id", tf_child_frame_id, "aircraft");
 		sp_nh.param("tf/rate_limit", tf_rate, 10.0);
 
+		// thrust msg subscriber to sync
+		message_filters::Subscriber<mavros_msgs::Thrust> thrust_sub(sp_nh, "thrust", 10);
```
vooon: Why subscriber here? How it can survive when initialize() returns?

mavros/src/plugins/setpoint_attitude.cpp
```diff
@@ -46,26 +51,46 @@ class SetpointAttitudePlugin : public plugin::PluginBase,
...
+
 		if (tf_listen) {
 			ROS_INFO_STREAM_NAMED("attitude",
-					"Listen to desired attitude transform "
-					<< tf_frame_id << " -> " << tf_child_frame_id);
-			tf2_start("AttitudeSpTF", &SetpointAttitudePlugin::transform_cb);
+						"Listen to desired attitude transform "
+						<< tf_frame_id << " -> " << tf_child_frame_id);
+
+			tf2_start<mavros_msgs::Thrust>("AttitudeSpTFSync", &SetpointAttitudePlugin::transform_cb);
```
vooon: Message type should be deduced from callback type automagically.

mavros/include/mavros/setpoint_mixin.h
TSC21: If you are using one thread for `tf2_start` in one plugin and another `tf2_start` for another plugin which is using sync (the case of `setpoint_ attitude`), shouldn't two different threads be used?

mavros/src/plugins/setpoint_attitude.cpp
TSC21: Makes sense

mavros/src/plugins/setpoint_attitude.cpp
TSC21: Oh yeah mistake. I will put on `public` and then initialize here, as before.

mavros/src/plugins/setpoint_attitude.cpp
TSC21: so I should remove `<mavros_msgs::Thrust>`?

TSC21:
> but also please update config yaml in launch.

Ok sure.

mavros/include/mavros/setpoint_mixin.h
```diff
@@ -193,7 +199,48 @@ class TF2ListenerMixin {
...
+			ros::Rate rate(static_cast<D *>(this)->tf_rate);
+			while (ros::ok()) {
+				// Wait up to 3s for transform
+				if (m_uas_->tf2_buffer.canTransform(_frame_id, _child_frame_id, ros::Time(0), ros::Duration(3.0))) {
+					try {
+						auto transform = m_uas_->tf2_buffer.lookupTransform(
+								_frame_id, _child_frame_id, ros::Time(0), ros::Duration(3.0));
+
+						tf2_ros::MessageFilter<T> tf2_filter(_topic_sub, m_uas_->tf2_buffer, _frame_id, 10, _sp_nh);
+						tf2_filter.registerCallback(boost::bind(cbp, static_cast<D *>(this), transform, _1));
```
vooon: That looks weird. Creating new filter on each TF rx not good. And `_topic_sub` not initialized, there no topic name.
Do you test that?

mavros/include/mavros/setpoint_mixin.h
vooon: Each plugin instance (object) have their `std::thread` object.
It is not **shared** between plugins, only inside class instance.

mavros/src/plugins/setpoint_attitude.cpp
vooon: Yes, and in theory is should still works :)

mavros/include/mavros/setpoint_mixin.h
TSC21: I had `_top_sub` initialized before in a proper way. I don't know why I changed it. I would roll back and initialize it on plugin and use it here on the template by applying `static_cast<D *>(this)->`. Sorry for that

TSC21: > Creating new filter on each TF rx not good.

Yeah I will put it outside the while loop.

mavros/include/mavros/setpoint_mixin.h
TSC21: Ok will use one only then.

mavros/src/plugins/setpoint_attitude.cpp
TSC21: It didn't work

TSC21:
@vooon done. Do you have a suggestion on how https://github.com/mavlink/mavros/pull/700#discussion-diff-123747442R74 could work?

mavros/src/plugins/setpoint_attitude.cpp
```diff
@@ -106,7 +106,10 @@ class SetpointAttitudePlugin : public plugin::PluginBase,
 	std::string tf_frame_id;
 	std::string tf_child_frame_id;
 
+	message_filters::Subscriber<mavros_msgs::Thrust> sub;
```
vooon: `s/sub/thd_sub/`

TSC21: What do you mean?

vooon: General "sub" bad for later support, thats why i want more specific name like `thd_sub`.
And to solve problem with `tf_start()`, just pass by reference.

Or move whole `tf_start()` code out of mixin, since it looks specific to that plugin.
Or you plan to use that variation somewhere else?

mavros/include/mavros/setpoint_mixin.h
```diff
@@ -200,6 +204,47 @@ class TF2ListenerMixin {
...
+
+		tf_thread = std::thread([this, cbp]() {
+			mavconn::utils::set_this_thread_name("%s", tf_thd_name.c_str());
+
+			mavros::UAS *m_uas_ = static_cast<D *>(this)->m_uas;
+			ros::NodeHandle &_sp_nh = static_cast<D *>(this)->sp_nh;
+			std::string &_frame_id = static_cast<D *>(this)->tf_frame_id;
+			std::string &_child_frame_id = static_cast<D *>(this)->tf_child_frame_id;
+
+			message_filters::Subscriber<T> &_topic_sub = static_cast<D *>(this)->sub;
```
vooon: Pass it via args!

mavros/include/mavros/setpoint_mixin.h
```diff
@@ -200,6 +204,47 @@ class TF2ListenerMixin {
...
+	 *
+	 * @param _thd_name  listener thread name
+	 * @param cbp        plugin callback function
+	 */
+	template <class T>
+	void tf2_start(const char *_thd_name, void (D::*cbp)(const geometry_msgs::TransformStamped &, const typename T::ConstPtr &))
+	{
+		tf_thd_name = _thd_name;
+
+		tf_thread = std::thread([this, cbp]() {
```
vooon:
```cpp
void tf2_start(const char *_thd_name, message_fileters::Subscriber<T> &tsub, ...)
...
th_therad = std::thread([this, cbp, tsub]() {
...
```

PR #706, opened 2017-05-06
WIP: Add odometry plugin.

jgoppert:
This adds an odometry plugin that simplifies interaction with ROS based estimators.

@TSC21 
@mhkabir

### DIFF
--- a/mavros/include/mavros/frame_tf.h
+++ b/mavros/include/mavros/frame_tf.h
@@ -319,9 +319,7 @@ inline Eigen::Quaterniond mavlink_to_quaternion(const std::array<float, 4> &q)
  */
 template<class T, std::size_t SIZE>
 inline void covariance_to_mavlink(const T &cov, std::array<float, SIZE> &covmsg) {
-	for (size_t i = 0; i < SIZE; i++) {
-		covmsg[i] = cov[i];
-	}
+	std::copy(cov.cbegin(), cov.cend(), covmsg.begin());
 }
 }	// namespace ftf
 }	// namespace mavros
--- a/mavros/include/mavros/frame_tf.h
+++ b/mavros/include/mavros/frame_tf.h
@@ -27,7 +27,6 @@
 
 namespace mavros {
 namespace ftf {
-
 //! Type matching rosmsg for covariance 3x3
 using Covariance3d = sensor_msgs::Imu::_angular_velocity_covariance_type;
 
@@ -53,7 +52,6 @@ enum class StaticTF {
 };
 
 namespace detail {
-
 /**
  * @brief Transform representation of attitude from 1 frame to another
  * (e.g. transfrom attitude from representing  from base_link -> NED
@@ -71,14 +69,18 @@ Eigen::Quaterniond transform_orientation(const Eigen::Quaterniond &q, const Stat
 Eigen::Vector3d transform_frame(const Eigen::Vector3d &vec, const Eigen::Quaterniond &q);
 
 /**
- * @brief Transform convariance expressed in one frame to another
+ * @brief Transform 3x3 convariance expressed in one frame to another
  *
  * General function. Please use specialized enu-ned and ned-enu variants.
  */
 Covariance3d transform_frame(const Covariance3d &cov, const Eigen::Quaterniond &q);
 
-// XXX TODO implement that function
-//Covariance6d transform_frame(const Covariance6d &cov, const Eigen::Quaterniond &q);
+/**
+ * @brief Transform 6x6 convariance expressed in one frame to another
+ *
+ * General function. Please use specialized enu-ned and ned-enu variants.
+ */
+Covariance6d transform_frame(const Covariance6d &cov, const Eigen::Quaterniond &q);
 
 /**
  * @brief Transform data experessed in one frame to another frame.
@@ -88,19 +90,22 @@ Covariance3d transform_frame(const Covariance3d &cov, const Eigen::Quaterniond &
 Eigen::Vector3d transform_static_frame(const Eigen::Vector3d &vec, const StaticTF transform);
 
 /**
- * @brief Transform convariance expressed in one frame to another
+ * @brief Transform 3d convariance expressed in one frame to another
  *
  * General function. Please use specialized enu-ned and ned-enu variants.
  */
 Covariance3d transform_static_frame(const Covariance3d &cov, const StaticTF transform);
 
-// XXX TODO implement that function
-//Covariance6d transform_static_frame(const Covariance6d &cov, const StaticTF transform);
+/**
+ * @brief Transform 6d convariance expressed in one frame to another
+ *
+ * General function. Please use specialized enu-ned and ned-enu variants.
+ */
+Covariance6d transform_static_frame(const Covariance6d &cov, const StaticTF transform);
 
 inline double transform_frame_yaw(double yaw) {
 	return -yaw;
 }
-
 }	// namespace detail
 
 // -*- frame tf -*-
@@ -309,5 +314,14 @@ inline Eigen::Quaterniond mavlink_to_quaternion(const std::array<float, 4> &q)
 	return Eigen::Quaterniond(q[0], q[1], q[2], q[3]);
 }
 
+/**
+ * @brief Store Covariance matrix to MAVLink float[n] format
+ */
+template<class T, std::size_t SIZE>
+inline void covariance_to_mavlink(const T &cov, std::array<float, SIZE> &covmsg) {
+	for (size_t i = 0; i < SIZE; i++) {
+		covmsg[i] = cov[i];
+	}
+}
 }	// namespace ftf
 }	// namespace mavros
--- a/mavros/include/mavros/frame_tf.h
+++ b/mavros/include/mavros/frame_tf.h
@@ -349,35 +349,13 @@ inline void covariance_to_mavlink(const T &cov, std::array<float, SIZE> &covmsg)
  */
 template<class T, std::size_t SIZE>
 inline void covariance9d_urt_to_mavlink(const T &cov, std::array<float, SIZE> &covmsg) {
-	std::vector<float> aux;
+	EigenMapConstCovariance9d m(cov.data());
 
-	for (size_t i = 0; i <= 8; i ++) {
-		aux.push_back(cov.data()[i]);
-	}
-	for (size_t i = 10; i <= 17; i ++) {
-		aux.push_back(cov.data()[i]);
-	}
-	for (size_t i = 20; i <= 26; i ++) {
-		aux.push_back(cov.data()[i]);
-	}
-	for (size_t i = 30; i <= 35; i ++) {
-		aux.push_back(cov.data()[i]);
-	}
-	for (size_t i = 40; i <= 44; i ++) {
-		aux.push_back(cov.data()[i]);
-	}
-	for (size_t i = 50; i <= 53; i ++) {
-		aux.push_back(cov.data()[i]);
-	}
-	for (size_t i = 60; i <= 62; i ++) {
-		aux.push_back(cov.data()[i]);
-	}
-	for (size_t i = 70; i <= 71; i ++) {
-		aux.push_back(cov.data()[i]);
-	}
-	aux.push_back(cov.data()[80]);
+	auto out = covmsg.begin();
 
-	std::copy(aux.cbegin(), aux.cend(), covmsg.begin());
+	for (size_t x = 0; x < m.cols(); x++)
+	    for (size_t y = x; y < m.rows(); y++)
+	        *out++ = m(y, x);
 }
 }	// namespace ftf
 }	// namespace mavros
--- a/mavros/include/mavros/frame_tf.h
+++ b/mavros/include/mavros/frame_tf.h
@@ -27,11 +27,14 @@
 
 namespace mavros {
 namespace ftf {
-//! Type matching rosmsg for covariance 3x3
+//! Type matching rosmsg for 3x3 covariance matrix
 using Covariance3d = sensor_msgs::Imu::_angular_velocity_covariance_type;
 
-//! Type matching rosmsg for covariance 6x6
+//! Type matching rosmsg for 6x6 covariance matrix
 using Covariance6d = geometry_msgs::PoseWithCovariance::_covariance_type;
+
+//! Type matching rosmsg for 9x9 covariance matrix
+using Covariance9d = std::array<double, 81>;
 
 //! Eigen::Map for Covariance3d
 using EigenMapCovariance3d = Eigen::Map<Eigen::Matrix<double, 3, 3, Eigen::RowMajor> >;
@@ -40,6 +43,11 @@ using EigenMapConstCovariance3d = Eigen::Map<const Eigen::Matrix<double, 3, 3, E
 //! Eigen::Map for Covariance6d
 using EigenMapCovariance6d = Eigen::Map<Eigen::Matrix<double, 6, 6, Eigen::RowMajor> >;
 using EigenMapConstCovariance6d = Eigen::Map<const Eigen::Matrix<double, 6, 6, Eigen::RowMajor> >;
+
+//! Eigen::Map for Covariance9d
+using EigenMapCovariance9d = Eigen::Map<Eigen::Matrix<double, 9, 9, Eigen::RowMajor> >;
+using EigenMapConstCovariance9d = Eigen::Map<const Eigen::Matrix<double, 9, 9, Eigen::RowMajor> >;
+
 
 /**
  * @brief Orientation transform options when applying rotations to data
@@ -83,6 +91,13 @@ Covariance3d transform_frame(const Covariance3d &cov, const Eigen::Quaterniond &
 Covariance6d transform_frame(const Covariance6d &cov, const Eigen::Quaterniond &q);
 
 /**
+ * @brief Transform 9x9 convariance expressed in one frame to another
+ *
+ * General function. Please use specialized enu-ned and ned-enu variants.
+ */
+Covariance9d transform_frame(const Covariance9d &cov, const Eigen::Quaterniond &q);
+
+/**
  * @brief Transform data experessed in one frame to another frame.
  *
  * General function. Please use specialized enu-ned and ned-enu variants.
@@ -102,6 +117,13 @@ Covariance3d transform_static_frame(const Covariance3d &cov, const StaticTF tran
  * General function. Please use specialized enu-ned and ned-enu variants.
  */
 Covariance6d transform_static_frame(const Covariance6d &cov, const StaticTF transform);
+
+/**
+ * @brief Transform 9d convariance expressed in one frame to another
+ *
+ * General function. Please use specialized enu-ned and ned-enu variants.
+ */
+Covariance9d transform_static_frame(const Covariance9d &cov, const StaticTF transform);
 
 inline double transform_frame_yaw(double yaw) {
 	return -yaw;
@@ -320,6 +342,42 @@ inline Eigen::Quaterniond mavlink_to_quaternion(const std::array<float, 4> &q)
 template<class T, std::size_t SIZE>
 inline void covariance_to_mavlink(const T &cov, std::array<float, SIZE> &covmsg) {
 	std::copy(cov.cbegin(), cov.cend(), covmsg.begin());
+}
+
+/**
+ * @brief Store half upper right triangular of 9d Covariance matrix to MAVLink float[n] format
+ */
+template<class T, std::size_t SIZE>
+inline void covariance9d_urt_to_mavlink(const T &cov, std::array<float, SIZE> &covmsg) {
+	std::vector<float> aux;
+
+	for (size_t i = 0; i <= 8; i ++) {
+		aux.push_back(cov.data()[i]);
+	}
+	for (size_t i = 10; i <= 17; i ++) {
+		aux.push_back(cov.data()[i]);
+	}
+	for (size_t i = 20; i <= 26; i ++) {
+		aux.push_back(cov.data()[i]);
+	}
+	for (size_t i = 30; i <= 35; i ++) {
+		aux.push_back(cov.data()[i]);
+	}
+	for (size_t i = 40; i <= 44; i ++) {
+		aux.push_back(cov.data()[i]);
+	}
+	for (size_t i = 50; i <= 53; i ++) {
+		aux.push_back(cov.data()[i]);
+	}
+	for (size_t i = 60; i <= 62; i ++) {
+		aux.push_back(cov.data()[i]);
+	}
+	for (size_t i = 70; i <= 71; i ++) {
+		aux.push_back(cov.data()[i]);
+	}
+	aux.push_back(cov.data()[80]);
+
+	std::copy(aux.cbegin(), aux.cend(), covmsg.begin());
 }
 }	// namespace ftf
 }	// namespace mavros
--- a/mavros/include/mavros/frame_tf.h
+++ b/mavros/include/mavros/frame_tf.h
@@ -34,7 +34,7 @@ using Covariance3d = sensor_msgs::Imu::_angular_velocity_covariance_type;
 using Covariance6d = geometry_msgs::PoseWithCovariance::_covariance_type;
 
 //! Type matching rosmsg for 9x9 covariance matrix
-using Covariance9d = std::array<double, 81>;
+using Covariance9d = boost::array<double, 81>;
 
 //! Eigen::Map for Covariance3d
 using EigenMapCovariance3d = Eigen::Map<Eigen::Matrix<double, 3, 3, Eigen::RowMajor> >;
--- a/mavros/src/lib/ftf_frame_conversions.cpp
+++ b/mavros/src/lib/ftf_frame_conversions.cpp
@@ -91,22 +91,23 @@ Covariance6d transform_static_frame(const Covariance6d &cov, const StaticTF tran
 	EigenMapCovariance6d cov_out(cov_out_.data());
 
 	Eigen::MatrixXd Affine6dTf(6,6);
-	Eigen::Matrix3d R = Eigen::MatrixXd::Identity(3,3);
+	Eigen::Matrix3d _R = NED_ENU_Q.normalized().toRotationMatrix();
+	Eigen::Matrix3d R_ = AIRCRAFT_BASELINK_Q.normalized().toRotationMatrix();
 
 	switch (transform) {
 	default: {
 	case StaticTF::NED_TO_ENU:
 	case StaticTF::ENU_TO_NED:
-		Affine6dTf << NED_ENU_AFFINE * R, Eigen::MatrixXd::Zero(3, 3),
-			      Eigen::MatrixXd::Zero(3, 3), NED_ENU_AFFINE * R;
+		Affine6dTf << _R, Eigen::MatrixXd::Zero(3, 3),
+			      Eigen::MatrixXd::Zero(3, 3), _R;
 
 		cov_out = Affine6dTf * cov_in * Affine6dTf.transpose();
 		return cov_out_;
 
 	case StaticTF::AIRCRAFT_TO_BASELINK:
 	case StaticTF::BASELINK_TO_AIRCRAFT:
-		Affine6dTf << AIRCRAFT_BASELINK_AFFINE * R, Eigen::MatrixXd::Zero(3, 3),
-			      Eigen::MatrixXd::Zero(3, 3), AIRCRAFT_BASELINK_AFFINE * R;
+		Affine6dTf << R_, Eigen::MatrixXd::Zero(3, 3),
+			      Eigen::MatrixXd::Zero(3, 3), R_;
 
 		cov_out = Affine6dTf * cov_in * Affine6dTf.transpose();
 		return cov_out_;
--- a/mavros/src/lib/ftf_frame_conversions.cpp
+++ b/mavros/src/lib/ftf_frame_conversions.cpp
@@ -98,7 +98,7 @@ Covariance6d transform_static_frame(const Covariance6d &cov, const StaticTF tran
 	case StaticTF::NED_TO_ENU:
 	case StaticTF::ENU_TO_NED:
 		Affine6dTf << NED_ENU_AFFINE * R, Eigen::MatrixXd::Zero(3, 3),
-									Eigen::MatrixXd::Zero(3, 3), NED_ENU_AFFINE * R;
+			      Eigen::MatrixXd::Zero(3, 3), NED_ENU_AFFINE * R;
 
 		cov_out = Affine6dTf * cov_in * Affine6dTf.transpose();
 		return cov_out_;
@@ -106,7 +106,7 @@ Covariance6d transform_static_frame(const Covariance6d &cov, const StaticTF tran
 	case StaticTF::AIRCRAFT_TO_BASELINK:
 	case StaticTF::BASELINK_TO_AIRCRAFT:
 		Affine6dTf << AIRCRAFT_BASELINK_AFFINE * R, Eigen::MatrixXd::Zero(3, 3),
-									Eigen::MatrixXd::Zero(3, 3), AIRCRAFT_BASELINK_AFFINE * R;
+			      Eigen::MatrixXd::Zero(3, 3), AIRCRAFT_BASELINK_AFFINE * R;
 
 		cov_out = Affine6dTf * cov_in * Affine6dTf.transpose();
 		return cov_out_;
@@ -140,7 +140,7 @@ Covariance6d transform_frame(const Covariance6d &cov, const Eigen::Quaterniond &
 	Eigen::Matrix3d R = q.normalized().toRotationMatrix();
 
 	Affine6dTf << R, Eigen::MatrixXd::Zero(3, 3),
-								Eigen::MatrixXd::Zero(3, 3), R;
+		      Eigen::MatrixXd::Zero(3, 3), R;
 
 	cov_out = Affine6dTf * cov_in * Affine6dTf.transpose();
 
--- a/mavros/src/lib/ftf_frame_conversions.cpp
+++ b/mavros/src/lib/ftf_frame_conversions.cpp
@@ -61,8 +61,8 @@ static const auto ZEROM3D = Eigen::Matrix3d::Zero();
 /**
  * @brief Auxiliar matrices to Covariance transforms
  */
-using Affine6dTF = Eigen::Matrix<double, 6, 6>;
-using Affine9dTF = Eigen::Matrix<double, 9, 9>;
+using Affine6dTF = Eigen::Matrix<double, 6, 6, Eigen::RowMajor>;
+using Affine9dTF = Eigen::Matrix<double, 9, 9, Eigen::RowMajor>;
 
 
 Eigen::Quaterniond transform_orientation(const Eigen::Quaterniond &q, const StaticTF transform)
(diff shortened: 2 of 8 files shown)


mavros_extras/src/plugins/odom.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+private:
+	ros::NodeHandle _nh;
+	ros::Subscriber _odom_sub;
+
+	/* -*- callbacks -*- */
+
+	void odom_cb(const nav_msgs::Odometry::ConstPtr &odom)
+	{
+		//ROS_INFO("odometry callback");
```
TSC21: maybe we can clear this?

mavros_extras/src/plugins/odom.cpp
```diff
@@ -0,0 +1,111 @@
...
+		lpos.vx = odom->twist.twist.linear.y;
+		lpos.vy = odom->twist.twist.linear.x;
+		lpos.vz = -odom->twist.twist.linear.z;
+		lpos.ax = 0;
+		lpos.ay = 0;
+		lpos.az = 0;
+		size_t i = 0;
+		for (int row=0; row < 6; row++) {
+			for (int col=row; col < 6; col++) {
+				//ROS_INFO("row: %d, col:%d, i:%d", row, col, i);
```
TSC21: also this...

mavros_extras/src/plugins/odom.cpp
```diff
@@ -0,0 +1,111 @@
...
+				lpos.covariance[i] = odom->pose.covariance[row*6 + col];
+				i += 1;
+			}
+		}
+		UAS_FCU(m_uas)->send_message_ignore_drop(lpos);
```
TSC21: maybe you can put both msg send on the end?

mavros_extras/src/plugins/odom.cpp
```diff
@@ -0,0 +1,111 @@
...
+		lpos.vy = odom->twist.twist.linear.x;
+		lpos.vz = -odom->twist.twist.linear.z;
+		lpos.ax = 0;
+		lpos.ay = 0;
+		lpos.az = 0;
+		size_t i = 0;
+		for (int row=0; row < 6; row++) {
+			for (int col=row; col < 6; col++) {
+				//ROS_INFO("row: %d, col:%d, i:%d", row, col, i);
+				lpos.covariance[i] = odom->pose.covariance[row*6 + col];
```
TSC21: we need to apply frame transform from ENU to NED. Maybe we can discuss this and add it to `frame_tf.h`

mavros_extras/src/plugins/odom.cpp
```diff
@@ -0,0 +1,111 @@
...
+		lpos.y = odom->pose.pose.position.x;
+		lpos.z = -odom->pose.pose.position.z;
+		lpos.vx = odom->twist.twist.linear.y;
+		lpos.vy = odom->twist.twist.linear.x;
+		lpos.vz = -odom->twist.twist.linear.z;
+		lpos.ax = 0;
+		lpos.ay = 0;
+		lpos.az = 0;
+		size_t i = 0;
+		for (int row=0; row < 6; row++) {
```
TSC21: probably better to create a `std::array` and use `std::copy`

mavros_extras/src/plugins/odom.cpp
```diff
@@ -0,0 +1,111 @@
...
+		att.rollspeed =  odom->twist.twist.angular.y;
+		att.pitchspeed =  odom->twist.twist.angular.x;
+		att.yawspeed =  -odom->twist.twist.angular.z;
+
+		att.time_usec = stamp;
+		att.q[0] = odom->pose.pose.orientation.w;
+		att.q[1] = odom->pose.pose.orientation.x;
+		att.q[2] = odom->pose.pose.orientation.y;
+		att.q[3] = odom->pose.pose.orientation.z;
+		for (size_t i=0; i < 9; i++) {
```
TSC21: `std::copy` suits better I think

TSC21: well we can leave it like this as it would require initiate att.covariance also as an array

mavros_extras/src/plugins/odom.cpp
```diff
@@ -0,0 +1,111 @@
...
+		mavlink::common::msg::LOCAL_POSITION_NED_COV lpos{};
+		lpos.time_usec = stamp;
+		// NED to ENU
+		// TODO check these are in the correct frame
+		lpos.x = odom->pose.pose.position.y;
+		lpos.y = odom->pose.pose.position.x;
+		lpos.z = -odom->pose.pose.position.z;
+		lpos.vx = odom->twist.twist.linear.y;
+		lpos.vy = odom->twist.twist.linear.x;
+		lpos.vz = -odom->twist.twist.linear.z;
```
vooon: Use `tft::transform_frame_enu_ned()` from [frame_tf.h][1]

Also i'm more like to use [cog.py][2] to generate msg-to-msg copy code.

[1]: https://github.com/mavlink/mavros/blob/master/mavros/include/mavros/frame_tf.h
[2]: https://pypi.python.org/pypi/cogapp

mavros_extras/src/plugins/odom.cpp
```diff
@@ -0,0 +1,111 @@
...
+		// TODO check these are in the correct frame
+		att.rollspeed =  odom->twist.twist.angular.y;
+		att.pitchspeed =  odom->twist.twist.angular.x;
+		att.yawspeed =  -odom->twist.twist.angular.z;
+
+		att.time_usec = stamp;
+		att.q[0] = odom->pose.pose.orientation.w;
+		att.q[1] = odom->pose.pose.orientation.x;
+		att.q[2] = odom->pose.pose.orientation.y;
+		att.q[3] = odom->pose.pose.orientation.z;
```
vooon: `ftf::quaternion_to_mavlink()`.

mavros_extras/src/plugins/odom.cpp
vooon: I think better to finish `ftf::transform_frame(Covariance6d)` function implementation and use it.

mavros_extras/src/plugins/odom.cpp
TSC21: yeah forgot that transform also

mavros_extras/src/plugins/odom.cpp
TSC21: yeah @vooon that's the intention

mavros_extras/src/plugins/odom.cpp
```diff
@@ -78,15 +78,24 @@ class OdometryPlugin : public plugin::PluginBase {
 
 		lpos.time_usec = stamp;
 
+		// [[[cog:
+		// for f in "xyz":
+		//     cog.outl("lpos.%s = pos_ned.%s();" % (f, f))
+		// for f in "xyz":
+		//     cog.outl("lpos.v%s = lin_vel_ned.%s();" % (f, f))
+		// for f in "xyz":
+		//     cog.outl("lpos.a%s = 0.0;" % f)
```
vooon: Why not:
```python
for a, b in (('', 'pos_ned'), ('v', 'lin_vel_ned'), ('a', 'zero')):
    for f in 'xyz':
        cog.outl("lpos.{a}{f} = {b}.{f}();".format(**locals()))
```

*add 
```cpp
const auto zero = Eigen::Vector3d::Zero();
```

mavros_extras/src/plugins/odom.cpp
```diff
@@ -0,0 +1,138 @@
...
+		att.pitchspeed = ang_vel_ned.y();
+		att.yawspeed = ang_vel_ned.z();
+		// [[[end]]] (checksum: e100d5c18a64c243df616f342f712ca1)
+
+		ftf::quaternion_to_mavlink(q_ned, att.q);
+
+		// TODO: apply ftf::transform_frame(Covariance9d)
+		for (size_t i = 0; i < 9; i++) {
+			att.covariance[i] = odom->pose.covariance[i];
+		}
```
vooon: I assume that `std::copy()` shall correctly move `array<double, 9>` to `array<float, 9>`, because it uses iterators.

```cpp
std::copy(odom->pose.covariance.cbegin(), odom->pose.covariance.cend(), att.covariance.begin());
```

mavros_extras/src/plugins/odom.cpp
TSC21: yeah makes it simples. thanks

mavros/src/lib/ftf_frame_conversions.cpp
```diff
@@ -85,21 +84,35 @@ Covariance3d transform_static_frame(const Covariance3d &cov, const StaticTF tran
...
+		Affine6dTf << NED_ENU_AFFINE * R, Eigen::MatrixXd::Zero(3, 3),
+		Eigen::MatrixXd::Zero(3, 3), NED_ENU_AFFINE * R;
+
+		cov_out = Affine6dTf * cov_in * Affine6dTf.transpose();
+		return cov_out_;
+
+	case StaticTF::AIRCRAFT_TO_BASELINK:
+	case StaticTF::BASELINK_TO_AIRCRAFT:
+		Affine6dTf << AIRCRAFT_BASELINK_AFFINE * R, Eigen::MatrixXd::Zero(3, 3),
+		Eigen::MatrixXd::Zero(3, 3), AIRCRAFT_BASELINK_AFFINE * R;
```
vooon: Uncrustify often breaks indent in eigen matrixes. I think Second line should have identical indent as after `<<`.

mavros_extras/src/plugins/odom.cpp
```diff
@@ -26,21 +25,24 @@ namespace extra_plugins {
...
+	{ }
 
 	void initialize(UAS &uas_)
 	{
 		PluginBase::initialize(uas_);
 
-		// tf params
-		_odom_sub = _nh.subscribe("odom", 10, &OdometryPlugin::odom_cb, this);
+		// general params
+		odom_nh.param("estimator_type", estimator_type, 3); // defaulted to VIO type
```
vooon:
```cpp
using mavlink::common::MAV_ESTIMATOR_TYPE;

// and then

enum_value(MAV_ESTIMATOR_TYPE::VIO);
```

mavros_extras/src/plugins/odom.cpp
```diff
@@ -49,14 +51,15 @@ class OdometryPlugin : public plugin::PluginBase {
 	}
 
 private:
-	ros::NodeHandle _nh;
+	ros::NodeHandle odom_nh;
 	ros::Subscriber _odom_sub;
 
+	int estimator_type;
```
vooon: Why `int` and not `MAV_ESTIMATOR_TYPE`?

mavros_extras/src/plugins/odom.cpp
```diff
@@ -49,14 +51,15 @@ class OdometryPlugin : public plugin::PluginBase {
 	}
 
 private:
-	ros::NodeHandle _nh;
+	ros::NodeHandle odom_nh;
 	ros::Subscriber _odom_sub;
```
vooon: `nh` is fine. I don't like starting `_`, prefer `m_` instead. But that isn't a problem.

mavros_extras/src/plugins/odom.cpp
```diff
@@ -67,16 +70,18 @@ class OdometryPlugin : public plugin::PluginBase {
...
 
 		uint64_t stamp = odom->header.stamp.toNSec() / 1e3;
 
 		// send LOCAL_POSITION_NED_COV
-		mavlink::common::msg::LOCAL_POSITION_NED_COV lpos {};
+		mavlink::common::msg::LOCAL_POSITION_NED_COV lpos{};
 
 		lpos.time_usec = stamp;
+		lpos.estimator_type = estimator_type;
+		ROS_DEBUG_STREAM_NAMED("odom", "Odometry: estimator type: " << utils::to_string_enum<mavlink::common::MAV_ESTIMATOR_TYPE>(estimator_type));
```
vooon: Why here? I think put that into `initialize()` is enough.

mavros_extras/src/plugins/odom.cpp
vooon: I think that not needed. Fill one message + send is fine.

mavros/include/mavros/frame_tf.h
```diff
@@ -309,5 +314,14 @@ inline Eigen::Quaterniond mavlink_to_quaternion(const std::array<float, 4> &q)
...
 
+/**
+ * @brief Store Covariance matrix to MAVLink float[n] format
+ */
+template<class T, std::size_t SIZE>
+inline void covariance_to_mavlink(const T &cov, std::array<float, SIZE> &covmsg) {
+	for (size_t i = 0; i < SIZE; i++) {
+		covmsg[i] = cov[i];
+	}
+}
```
vooon: Why? `std::copy()` does not work?

TSC21:
> Now looks better. One thing that i'm not sure: maybe call it `send_odom`?

As it is I think we could but probably the next set should be add handlers for `LOCAL_POSITION_NED_COV` and `ATTITUDE_QUATERNION_COV` and publish those in a ROS Odometry msg.

mavros_extras/src/plugins/odom.cpp
TSC21: I kept it as `int` so it can be easily received and modified on the `yaml` file.

TSC21: https://github.com/mavlink/mavros/pull/706/commits/b08ffc0a97a030e5ba883d7a97c5d9ecbd10aefa#r115705809

mavros_extras/src/plugins/odom.cpp
TSC21: https://github.com/mavlink/mavros/pull/706/commits/b08ffc0a97a030e5ba883d7a97c5d9ecbd10aefa#r115705809

TSC21:
Ok right now there are two things left to be fixed:
- Understand why is the code giving matrix out of bounds for the covariance
- Understand how to properly apply the rotation for angular velocity to the body frame

@jgoppert will do some tests on the simulation as soon as I can.

mavros_extras/src/plugins/odom.cpp
TSC21: We should give the option to set it on rasparam IMO.

TSC21:
Just a note: the pose is WRT and the velocities are relative to body frame, so currently the transforms are well applied.

mavros/src/lib/ftf_frame_conversions.cpp
```diff
@@ -91,22 +91,23 @@ Covariance6d transform_static_frame(const Covariance6d &cov, const StaticTF tran
 	EigenMapCovariance6d cov_out(cov_out_.data());
 
 	Eigen::MatrixXd Affine6dTf(6,6);
-	Eigen::Matrix3d R = Eigen::MatrixXd::Identity(3,3);
+	Eigen::Matrix3d _R = NED_ENU_Q.normalized().toRotationMatrix();
+	Eigen::Matrix3d R_ = AIRCRAFT_BASELINK_Q.normalized().toRotationMatrix();
```
vooon: Should be `static const` since that R do not change.
`_` in prefix/suffix is bad! Call it `R_NE`, `R_AB`.

Is `normalized()` required?
I tested:
```cpp
auto q1 = NED_ENU_Q;
auto q2 = NED_ENU_Q.normalized();
```

```
q1: [0.000000 0.707107 0.707107 0.000000]
q2: [0.000000 0.707107 0.707107 0.000000]
```

mavros/src/lib/ftf_frame_conversions.cpp
```diff
@@ -91,22 +91,23 @@ Covariance6d transform_static_frame(const Covariance6d &cov, const StaticTF tran
 	EigenMapCovariance6d cov_out(cov_out_.data());
 
 	Eigen::MatrixXd Affine6dTf(6,6);
```
vooon: Variables should be lower case, that not a type/class.

mavros/src/lib/ftf_frame_conversions.cpp
```diff
@@ -91,22 +91,23 @@ Covariance6d transform_static_frame(const Covariance6d &cov, const StaticTF tran
...
+	Eigen::Matrix3d R_ = AIRCRAFT_BASELINK_Q.normalized().toRotationMatrix();
 
 	switch (transform) {
 	default: {
 	case StaticTF::NED_TO_ENU:
 	case StaticTF::ENU_TO_NED:
-		Affine6dTf << NED_ENU_AFFINE * R, Eigen::MatrixXd::Zero(3, 3),
-			      Eigen::MatrixXd::Zero(3, 3), NED_ENU_AFFINE * R;
+		Affine6dTf << _R, Eigen::MatrixXd::Zero(3, 3),
+			      Eigen::MatrixXd::Zero(3, 3), _R;
```
vooon: Why not just add that?
```cpp
static const auto zero3d = Eigen::Matrix3d::Zero();
```

mavros_extras/src/plugins/odom.cpp
```diff
@@ -126,6 +126,10 @@ class OdometryPlugin : public plugin::PluginBase {
 		auto cov_ang_vel = ftf::transform_frame_aircraft_baselink(odom->twist.covariance);
 		ftf::covariance_to_mavlink(cov_ang_vel, att.covariance);
 
+		ROS_INFO("position cov: \n%1.20f \n%1.20f \n%1.20f \n%1.20f \n%1.20f \n%1.20f \n%1.20f",
+		cov_pos_linvel[0], cov_pos_linvel[1], cov_pos_linvel[7],  cov_pos_linvel[14],
+		cov_pos_linvel[21], cov_pos_linvel[28],  cov_pos_linvel[35]);
```
vooon: Why not something like that:

```cpp
ftf::EigenMapCovariance6d map(cov_pos_linvel.data());

ROS_INFO_STREAM("pos cov: " << std::endl << map);
```

And i hope that only for debugging?
Eigen::Map should work like Eigen::Matrix.

mavros/src/lib/ftf_frame_conversions.cpp
TSC21: In https://eigen.tuxfamily.org/dox/classEigen_1_1Quaternion.html:

> Operations interpreting the quaternion as rotation have undefined behavior if the quaternion is not normalized.

mavros/src/lib/ftf_frame_conversions.cpp
TSC21: yeah sure

mavros_extras/src/plugins/odom.cpp
TSC21: not used on the next commit

mavros/src/lib/ftf_frame_conversions.cpp
vooon: But NED_ENU_Q is normalized. Anyway it's fine if se make R static const.

mavros/src/lib/ftf_frame_conversions.cpp
```diff
@@ -21,21 +21,35 @@
...
+/**
+ * @brief 3-D matrices to fill 6-D rotation matrix applied to change covariance matrices coordinate frames
+ */
+static const Eigen::Matrix3d R_NE = NED_ENU_Q.normalized().toRotationMatrix();
+static const Eigen::Matrix3d R_AB = AIRCRAFT_BASELINK_Q.normalized().toRotationMatrix();
+
+/**
+ * @brief 3-D matrix, filled with zeros, that reprensent top-right and bottom-left of the rotation matrix
+ */
+static const auto zerom3d = Eigen::Matrix3d::Zero();
```
vooon: I think you leave that inside function scope.
Outside we perhaps better to keep prefixes, `NED_ENU_R`, `AIRCRAFT_BASELINK_R`.

Also capitalize zerom3d.

mavros_extras/src/plugins/odom.cpp
```diff
@@ -0,0 +1,156 @@
...
+		tf::poseMsgToEigen(odom->pose.pose, tr);
+		tf::vectorMsgToEigen(odom->twist.twist.linear, lin_vel_enu);
+		tf::vectorMsgToEigen(odom->twist.twist.angular, ang_vel_enu);
+
+		Eigen::Quaterniond enu_orientation;
+		tf::quaternionMsgToEigen(m_uas->get_attitude_orientation(), enu_orientation);
+
+		// body frame rotations must be aware of current attitude of the vehicle
+		auto q_ned_current = ftf::transform_orientation_enu_ned(
+					ftf::transform_orientation_baselink_aircraft(enu_orientation));
```
vooon: You sure? Looks a bit weird.

TSC21: After testing on simulation I'm mostly sure this is correct 'cause the transform is in body frame, which means it needs to follow the attitude of the body (aircraft) frame itself. With only applying a transform without making a rotation to the current attitude, I was having incorrect yaw rotations.
I'm also sure that this may be the src of some problems people are having with the mocap attitude having some weird rotations. When I make sure this works I'll probably assert of the transforms we have on the mocap plugin and check what are the variables that should be on world frame and what are the ones that need to be on body frame.

mavros/src/lib/ftf_frame_conversions.cpp
TSC21: > I think you leave that inside function scope.

It is used in all the later functions, so I don't see why not.

TSC21:
> Ok, now lgtm. But can you group ftf calls in one place?
Then automatically group `ftf::quaternion_...` and `ftf::covariance_...`

What do you mean with grouping? Put them all together in the code (line after line)?

TSC21:
I'm just waiting for @jgoppert testing and approval on the angular velocity transform so we can merge this.

vooon:
@TSC21 I prefer "rebase and merge" method (see green button drop-down list).

TSC21:
OK so I checked the msg definition from `LOCAL_POSITION_NED_COV` and seems like that the covariance upper right triangular values, besides being a 9x9 matrix (includes the covariance relative to the acceleration).
So I add to redo some code and add the a 9d covariance transform and a way to get the URT (upper-right triangle), which current implementation is not that crisp and looks more like an hack. So I accept some suggestions on how to make it shorter and efficient (I was playing with the math and got tired of trying to push it to a for function xD ).
@jgoppert, @vooon any suggestions?

TSC21:
> Also i'm now thinking about tft:: helpers that can directly handle geometric_msgs::Vector3 and return Eigen::Vector3d.
Because we have too much copy-pase code with this older tf-eigen-conversion lib.

Agreed. Probably we should add also that to 0.20 milestone.

jgoppert:
This works for me so I"m good to merge.

I would like to think about working toward using tf2 lookup as in odom2 branch. Right now we are assuming the child_id and frame_id which is really breaking the contract from the user who sent the message. Based on the child_id and frame_id, we should use tf2 to lookup the tranformations relative to the NED frame as I do here: https://github.com/mavlink/mavros/blob/odom2/mavros_extras/src/plugins/odom.cpp#L70

I believe many of the issues of users with frame confusion on mavros is that they don't know we are breaking the contract and assuming the frame_id and child_id. We should really extend this to cover all of mavros including set points etc.

Also, we should be publishing local_origin_ned as discussed here: 
http://www.ros.org/reps/rep-0103.html

For outdoor systems where it is desireable to work under the north east down [6] (NED) convention, define an appropriately transformed secondary frame with the "_ned" suffix:

X north
Y east
Z down

I realize there is a concern about overhead of these lookups. I have test flown odom2 branch and didn't notice any impact. I was using tf2_ros static publisher for most transforms which should require little overhead, they fixed the issue with tf 1 static_publisher where they actually had to republish them. The only real lookup was between base_link and local_origin_ned.

@vooon, @TSC21 , what do you think?

mavros/include/mavros/frame_tf.h
```diff
@@ -309,5 +336,28 @@ inline Eigen::Quaterniond mavlink_to_quaternion(const std::array<float, 4> &q)
...
+template<class T, std::size_t SIZE>
+inline void covariance9d_urt_to_mavlink(const T &cov, std::array<float, SIZE> &covmsg) {
+	std::vector<float> aux;
+	size_t n = 9;
+	for (size_t i=0; i < n; i++) {
+		for (size_t j=i; j < n; j++) {
+			aux.push_back(cov.data()[i + j * n]);
+		}
+	}
+	std::copy(aux.cbegin(), aux.cend(), covmsg.begin());
```
vooon: What? Looks weird. And why you ever need temporary **dynamic** array (vector)?

mavros_extras/src/plugins/odom.cpp
```diff
@@ -0,0 +1,176 @@
...
+		// get current ENU FCU orientation
+		Eigen::Quaterniond enu_orientation;
+		tf::quaternionMsgToEigen(m_uas->get_attitude_orientation(), enu_orientation);
+
+		// Build 9x9 covariance matrix to be transformed and sent
+		ftf::Covariance9d cov_full;
+		ftf::EigenMapCovariance9d cov_full_map(cov_full.data());
+		ftf::EigenMapConstCovariance6d cov_in(odom->pose.covariance.data());
+		cov_full_map << cov_in, Eigen::MatrixXd::Zero(6, 3),
+		Eigen::MatrixXd::Zero(3, 6), Eigen::Matrix3d::Identity();
```
vooon: 9d? I missed something?

jgoppert:
Yeah. We were just talking abut replacing it with a static declaration using the eigen template size.

mavros_extras/src/plugins/odom.cpp
TSC21: The mavlink msg has a URT of a 9d covariance matrix (position, orientation and acceleration)

mavros/include/mavros/frame_tf.h
TSC21: It was the most straighforward way found. Do you have another suggestion, so this becomes generic and not only for 9d matrices?

vooon: Since you already stated that it is `covariance9d` you don't need a generic solution.

Naive solution:

```cpp
EigenMapConstCovariance9d m(cov.data());

auto out = covmsg.begin();
// auto view = m.triangularView(Eigen::Upper); - TODO look docs

for (size_t x = 0; x < m.cols(); x++)
    for (size_t y = x; y < m.rows(); y++)
        *out++ = m(y, x);
```

mavros/include/mavros/frame_tf.h
```diff
@@ -27,13 +27,15 @@
...
-//! Type matching rosmsg for covariance 3x3
+//! Type matching rosmsg for 3x3 covariance matrix
 using Covariance3d = sensor_msgs::Imu::_angular_velocity_covariance_type;
 
-//! Type matching rosmsg for covariance 6x6
+//! Type matching rosmsg for 6x6 covariance matrix
 using Covariance6d = geometry_msgs::PoseWithCovariance::_covariance_type;
 
+//! Type matching rosmsg for 9x9 covariance matrix
+using Covariance9d = std::array<double, 81>;
```
vooon: Actually ROS uses `boost::array` for fixed size arrays, but ok.

mavros/src/lib/ftf_frame_conversions.cpp
```diff
@@ -85,21 +107,61 @@ Covariance3d transform_static_frame(const Covariance3d &cov, const StaticTF tran
...
 
-#if 0
 Covariance6d transform_static_frame(const Covariance6d &cov, const StaticTF transform)
 {
-	//! @todo implement me!!!
+	Covariance6d cov_out_;
+	EigenMapConstCovariance6d cov_in(cov.data());
+	EigenMapCovariance6d cov_out(cov_out_.data());
+
+	Eigen::MatrixXd affine6dTF(6, 6);
```
vooon: Why?! MatrixXd is **dynamic**! Since you know dimensions there no reason not to use **static** variation by passing size to Matrix template.

```cpp
using Matrix6d = Eigen::Matrix<double, 6, 6>;
using Matrix9d = Eigen::Matrix<double, 9, 9>;
```

mavros/src/lib/ftf_frame_conversions.cpp
```diff
@@ -85,21 +107,61 @@ Covariance3d transform_static_frame(const Covariance3d &cov, const StaticTF tran
...
 	}
+}
+
+Covariance9d transform_static_frame(const Covariance9d &cov, const StaticTF transform)
+{
+	Covariance9d cov_out_;
+	EigenMapConstCovariance9d cov_in(cov.data());
+	EigenMapCovariance9d cov_out(cov_out_.data());
+
+	Eigen::MatrixXd affine9dTF(9, 9);
```
vooon: Same as for 6d.

mavros_extras/src/plugins/odom.cpp
vooon: Also better to make (inside func):

```cpp
static const auto zero6x3d = Eigen::Matrix<double, 6, 3>::Zero();
static const auto zero3x6d = Eigen::Matrix<double, 3, 6>::Zero();
```

mavros/src/lib/ftf_frame_conversions.cpp
```diff
@@ -58,6 +58,12 @@ static const auto NED_ENU_R = NED_ENU_Q.normalized().toRotationMatrix();
 static const auto AIRCRAFT_BASELINK_R = AIRCRAFT_BASELINK_Q.normalized().toRotationMatrix();
 static const auto ZEROM3D = Eigen::Matrix3d::Zero();
 
+/**
+ * @brief Auxiliar matrices to Covariance transforms
+ */
+using Affine6dTF = Eigen::Matrix<double, 6, 6>;
+using Affine9dTF = Eigen::Matrix<double, 9, 9>;
```
TSC21: I probably have to, instead of creating a affine 3d transform, apply a Affine3d for each rotation and then build the covariance matrix with the diagonal blocks already rotated - the current output seems that it is just multiplying the matrices but not applying the rotation the main diagonal blocks.

TSC21: Here's an example:
```bash
[ INFO] [1494793156.340441773, 8.962000000]: Odometry: Cov URT: 
 4.97062e-10 -1.50062e-57            0            0            0            0            0            0            0
           0  4.97062e-10 -1.26289e-42            0            0            0            0            0            0
           0            0  4.97062e-10            0            0            0            0            0            0
           0            0            0       0.0002 -6.03797e-52            0            0            0            0
           0            0            0            0       0.0002 -5.08144e-37            0            0            0
           0            0            0            0            0       0.0002            0            0            0
           0            0            0            0            0            0            1 -3.01899e-48            0
           0            0            0            0            0            0            0            1 -2.54072e-33
           0            0            0            0            0            0            0            0            1
```

mhkabir:
Hey guys, I can help out with this now that I have some free time. It would be good to get a brief on the open stuff @vooon @TSC21. Also squashing the commits would be nice - the diff looks very clean already.

I did a plugin (not upstreamable) when I developed the API for PX4 (https://github.com/PX4/Firmware/pull/6074). You can find that here : https://github.com/ProjectArtemis/mavros/blob/f043fa951657aa74c0219da836329367a395208f/mavros_extras/src/plugins/estimator_bridge.cpp

It does a bunch of things, and the frame transforms are all verified, and it has several hours of GPS-denied flight on it. We have a stable version of that interface running on some of our vehicles out in the field in active service.

TSC21:
Thanks for the help @mhkabir. But taking a look at your code seems that you are hardcoding the covariance. The purpose right now is apply the transform over a covariance being received from an external estimator. I already know what to do, so tomorrow I will push the final change so to merge this. Cheers

TSC21:
The new URT of the rotated covariance after the new commit:
```bash
 9.86231e-26  4.44159e-10  5.43938e-26            0            0            0            0            0            0
           0 -9.86231e-26 -1.09494e-41            0            0            0            0            0            0
           0            0 -4.44159e-10            0            0            0            0            0            0
           0            0            0  9.86231e-26  4.44159e-10  5.43938e-26            0            0            0
           0            0            0            0 -9.86231e-26 -1.09494e-41            0            0            0
           0            0            0            0            0 -4.44159e-10            0            0            0
           0            0            0            0            0            0  9.86231e-26  4.44159e-10  5.43938e-26
           0            0            0            0            0            0            0 -9.86231e-26 -1.09494e-41
           0            0            0            0            0            0            0            0 -4.44159e-10
```
Seems an acceptable rotation now.
So it lgtm in my opinion.
@jgoppert, @vooon should we merge this then?

TSC21:
After talking with @jgoppert I came to the conclusion that the first implementation was right. So I remove the last changes and apply a definitive commit.
Here's the latest covariance result after rotation:
```bash
[ INFO] [1494943715.279968291, 11.882000000]: Odometry: Cov URT: 
 2.20784e-10 -6.66544e-58            0            0            0            0            0            0            0
           0  2.20784e-10  -5.6095e-43            0            0            0            0            0            0
           0            0  2.20784e-10            0            0            0            0            0            0
           0            0            0       0.0002 -6.03797e-52            0            0            0            0
           0            0            0            0       0.0002 -5.08144e-37            0            0            0
           0            0            0            0            0       0.0002            0            0            0
           0            0            0            0            0            0            1 -3.01899e-48            0
           0            0            0            0            0            0            0            1 -2.54072e-33
           0            0            0            0            0            0            0            0            1
```

TSC21:
In any case we should consider, in the future, rely on tf2 to apply the transforms.

@jgoppert, @vooon lgtm?

jgoppert:
Yeah, that diagonal looks good now.

mavros_extras/src/plugins/odom.cpp
```diff
@@ -0,0 +1,184 @@
+/**
+ * @brief Odometry plugin
+ * @file odom.cpp
+ * @author James Goppert <james.goppert8@gmail.com>
```
jgoppert: @TSC21 should probably add your name here, and also can you fix my email above, has an extra 8.

TSC21: Mail fixed. No need to add my email. It was your idea and your implementation. I just added some features and robustness ;)

vooon:
Oh, NOOOO! :) I mean in global_position, where huge matrix looks not great.
But in tf that small
```
r << m, 0, 0,
     0, m, 0,
     0, 0, m;
```
Way better than:
```
r.block(0,0) = m;
r.block(1,1) = m;
r.block(2,2) = m;
```

TSC21:
We have 6d and 9d matrices. What do you mean by small?

vooon:
I mean that `m` visually hide 9d assign to looks like 3d.
But yes, assigning of blocks should be faster. And in some cases `m.setZero()` is not needed.

TSC21:
@vooon lgtm?

mhkabir:
Nuno, lgtm is "Looks Good To _Me_" :D

TSC21:
> Nuno, lgtm is "Looks Good To Me" :D

Always learning xD well I was asking looks good to Merge eheh

TSC21:
Can we merge this then?

vooon:
Yes, LGTM. But i will manually rebase it to master.

vooon:
@jgoppert merge of @TSC21 changes broke rebase. @TSC21 are you ok if i squash all into single commit?

TSC21:
Sure no prob @vooon

vooon:
I will try to apply full patch and then edit history to split lib frame_tf changes and plugin. But it not succeed then i simply squash & commit.

vooon:
Rebased to master. @jgoppert, @TSC21 please test because i done some style changes that may result in different math results (but should not!).

PR #746, opened 2017-07-02
Extras: add ardupilot rangefinder plugin

khancyr:
This add ardupilot rangefinder message support. It will be deprecated with next version that support DISTANCE_SENSOR msg


TSC21:
I don't like the approach msg_type<->plugin. I think it's better purpose<->plugin, but anyway it is your call. Besides, this plugin will only allows receiving msg, no sending as `distance_sensor` allows. Anyways, this is a addition to a soon deprecated msg, so should be added to blacklist at that time.

khancyr:
@TSC21 I agree I could have put it in distance_sensor file.

PR #747, opened 2017-07-02
Distance sensors: uncrustify and and new sensor type

khancyr:
need new mavlink released for : https://github.com/mavlink/mavlink/commit/35a89a2af36f4c0a4024f7a7419b620626b0543f

### DIFF
--- a/mavros_extras/src/plugins/distance_sensor.cpp
+++ b/mavros_extras/src/plugins/distance_sensor.cpp
@@ -227,18 +227,21 @@ private:
 		range->max_range = dist_sen.max_distance * 1E-2;
 		range->field_of_view = sensor->field_of_view;
 
-		if (dist_sen.type == enum_value(MAV_DISTANCE_SENSOR::LASER)) {
-			range->radiation_type = sensor_msgs::Range::INFRARED;
-		}
-		else if (dist_sen.type == enum_value(MAV_DISTANCE_SENSOR::ULTRASOUND)) {
-			range->radiation_type = sensor_msgs::Range::ULTRASOUND;
-		}
-		else {
-			ROS_ERROR_NAMED("distance_sensor",
-					"DS: %s: Wrong/undefined type of sensor (type: %d). Droping!...",
-					sensor->topic_name.c_str(), dist_sen.type);
-			return;
-		}
+        switch (dist_sen.type) {
+            case enum_value(MAV_DISTANCE_SENSOR::LASER):
+            case enum_value(MAV_DISTANCE_SENSOR::RADAR):
+            case enum_value(MAV_DISTANCE_SENSOR::UNKNOWN):
+                range->radiation_type = sensor_msgs::Range::INFRARED;
+                break;
+            case enum_value(MAV_DISTANCE_SENSOR::ULTRASOUND):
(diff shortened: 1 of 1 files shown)


mavros_extras/src/plugins/distance_sensor.cpp
```diff
@@ -198,45 +198,48 @@ class DistanceSensorPlugin : public plugin::PluginBase {
...
-		range->min_range = dist_sen.min_distance * 1E-2; // in meters
+		range->min_range = dist_sen.min_distance * 1E-2;// in meters
 		range->max_range = dist_sen.max_distance * 1E-2;
 		range->field_of_view = sensor->field_of_view;
 
-		if (dist_sen.type == enum_value(MAV_DISTANCE_SENSOR::LASER)) {
+		switch (dist_sen.type) {
+		case enum_value(MAV_DISTANCE_SENSOR::LASER):
+		case enum_value(MAV_DISTANCE_SENSOR::RADAR):
+		case enum_value(MAV_DISTANCE_SENSOR::UNKNOWN):
```
vooon: Why unknown assumed as IR?

khancyr: the most common type on copter are 1-D Lidar that are laser type and IR on ROS. So that is why I chose this. Another posibillity could be to add another parameter on the config file. What do you think?

TSC21: > Another posibillity could be to add another parameter on the config file. 

What parameter would that be?

vooon:
Seems that Travis was broken:
```
executing command [sudo -H pip install -U -q future]

/usr/local/lib/python2.7/dist-packages/pip/_vendor/requests/packages/urllib3/util/ssl_.py:318: SNIMissingWarning: An HTTPS request has been made, but the SNI (Subject Name Indication) extension to TLS is not available on this platform. This may cause the server to present an incorrect TLS certificate, which can cause validation failures. You can upgrade to a newer version of Python to solve this. For more information, see https://urllib3.readthedocs.io/en/latest/security.html#snimissingwarning.

  SNIMissingWarning

/usr/local/lib/python2.7/dist-packages/pip/_vendor/requests/packages/urllib3/util/ssl_.py:122: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. You can upgrade to a newer version of Python to solve this. For more information, see https://urllib3.readthedocs.io/en/latest/security.html#insecureplatformwarning.

  InsecurePlatformWarning

/usr/local/lib/python2.7/dist-packages/pip/_vendor/requests/packages/urllib3/util/ssl_.py:122: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. You can upgrade to a newer version of Python to solve this. For more information, see https://urllib3.readthedocs.io/en/latest/security.html#insecureplatformwarning.

  InsecurePlatformWarning

ERROR: the following rosdeps failed to install

  pip: Failed to detect successful installation of [future]
```

TSC21:
Yeah. But this can get in now, as Mavlink is already on deb packages.

PR #756, opened 2017-07-17
Add reasonable local_position default covariance and use tf2 for odom

jgoppert:
This uses tf2 for the odom plugin and listens to the passed child_frame_id and frame_id fields. It also adds reasonable defaults for local_position default covariance.

### DIFF
--- a/mavros/src/plugins/local_position.cpp
+++ b/mavros/src/plugins/local_position.cpp
@@ -117,15 +117,19 @@ private:
 		odom->pose.pose = pose->pose;
 		for (int i=0; i< 3; i++) {
 			// linear velocity
-			odom->twist.covariance[i + 6*i] = 0.0001;
+			odom->twist.covariance[i + 6*i] = 1;
 			// angular velocity
-			odom->twist.covariance[(i + 3) + 6*(i + 3)] = 0.0001;
+			odom->twist.covariance[(i + 3) + 6*(i + 3)] = 1;
 			// position/ attitude
 			if (i==2) {
+				// z
 				odom->pose.covariance[i + 6*i] = 1000;
+				// yaw
 				odom->pose.covariance[(i + 3) + 6*(i + 3)] = 1000;
 			} else {
+				// x, y
 				odom->pose.covariance[i + 6*i] = 1000;
+				// roll, pitch
 				odom->pose.covariance[(i + 3) + 6*(i + 3)] = 1000;
 			}
 		}
--- a/mavros/src/plugins/local_position.cpp
+++ b/mavros/src/plugins/local_position.cpp
@@ -117,16 +117,16 @@ private:
 		odom->pose.pose = pose->pose;
 		for (int i=0; i< 3; i++) {
 			// linear velocity
-			odom->twist.covariance[i + 6*i] = 1e-4;
+			odom->twist.covariance[i + 6*i] = 1e-6;
 			// angular velocity
-			odom->twist.covariance[(i + 3) + 6*(i + 3)] = 1e-4;
+			odom->twist.covariance[(i + 3) + 6*(i + 3)] = 1e-6;
 			// position/ attitude
 			if (i==2) {
-				odom->pose.covariance[i + 6*i] = 1e-4;
-				odom->pose.covariance[(i + 3) + 6*(i + 3)] = 1e-4;
+				odom->pose.covariance[i + 6*i] = 1;
+				odom->pose.covariance[(i + 3) + 6*(i + 3)] = 1e-6;
 			} else {
-				odom->pose.covariance[i + 6*i] = 1e-4;
-				odom->pose.covariance[(i + 3) + 6*(i + 3)] = 1e-4;
+				odom->pose.covariance[i + 6*i] = 1e-6;
+				odom->pose.covariance[(i + 3) + 6*(i + 3)] = 1e-6;
 			}
 		}
 
--- a/mavros/src/plugins/local_position.cpp
+++ b/mavros/src/plugins/local_position.cpp
@@ -117,16 +117,16 @@ private:
 		odom->pose.pose = pose->pose;
 		for (int i=0; i< 3; i++) {
 			// linear velocity
-			odom->twist.covariance[i + 6*i] = 1e-6;
+			odom->twist.covariance[i + 6*i] = 1e-4;
 			// angular velocity
-			odom->twist.covariance[(i + 3) + 6*(i + 3)] = 1e-6;
+			odom->twist.covariance[(i + 3) + 6*(i + 3)] = 1e-4;
 			// position/ attitude
 			if (i==2) {
-				odom->pose.covariance[i + 6*i] = 1;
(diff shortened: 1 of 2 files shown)


mavros_extras/src/plugins/odom.cpp
```diff
@@ -66,90 +67,97 @@ class OdometryPlugin : public plugin::PluginBase {
...
-		/**
-		 * @brief Velocity 6-D Covariance matrix
-		 * WRT world frame (half upper right triangular)
-		 */
-		auto cov_vel = ftf::transform_frame_enu_ned(
-					ftf::transform_frame_baselink_aircraft(
-						ftf::transform_frame_ned_aircraft(odom->twist.covariance, q_ned_current)));
-
-		ftf::EigenMapConstCovariance6d cov_vel_map(cov_vel.data());
-		ROS_DEBUG_STREAM_NAMED("odom", "Odometry: velocity covariance matrix: " << std::endl << cov_vel_map);
```
TSC21: The transforms don't work right for you?

TSC21: I think for a question of sanity of the code, if we want to apply tf lookups to one plugin, we would have to generalize it to all plugins.

mavros_extras/src/plugins/odom.cpp
```diff
@@ -66,90 +67,97 @@ class OdometryPlugin : public plugin::PluginBase {
...
+		// lookup transforms
+		Eigen::Affine3d tfChild2NED;
+		Eigen::Affine3d tfParent2NED;
+		try {
+			tfChild2NED = tf2::transformToEigen(m_uas->tf2_buffer.lookupTransform(
```
TSC21: Again, this would have to be generalized into all the repository, and not in specific for this case.

mavros/src/plugins/local_position.cpp
```diff
@@ -113,7 +113,22 @@ class LocalPositionPlugin : public plugin::PluginBase {
...
+			odom->twist.covariance[(i + 3) + 6*(i + 3)] = 1e-6;
+			// position/ attitude
+			if (i==2) {
+				odom->pose.covariance[i + 6*i] = 1;
+				odom->pose.covariance[(i + 3) + 6*(i + 3)] = 1e-6;
+			} else {
+				odom->pose.covariance[i + 6*i] = 1e-6;
+				odom->pose.covariance[(i + 3) + 6*(i + 3)] = 1e-6;
+			}
+		}
```
TSC21: I think this is fine for now. But in the near future we should add the support for `LOCAL_POSITION_NED_COV`.

mavros_extras/src/plugins/odom.cpp
TSC21: So basically this would be a function that looks up to the transform between the parent and the child where a static tf is applied (ENU<->NED). That static tf would have to be added to the launch files I suppose.

mavros/src/plugins/local_position.cpp
vooon: Just guess, but i'm feeling that Eigen::Map allow to make more readable code.

mavros_extras/src/plugins/odom.cpp
vooon: I don't think that **this** PR should change other plugins, but generalized API would be nice to see.

mavros_extras/src/plugins/odom.cpp
```diff
@@ -66,90 +67,97 @@ class OdometryPlugin : public plugin::PluginBase {
...
+		Eigen::Vector3d angVelChild(
+			odom->twist.twist.angular.x,
+			odom->twist.twist.angular.y,
+			odom->twist.twist.angular.z);
+
+		Eigen::Quaterniond qParent2Child(
+			odom->pose.pose.orientation.w,
+			odom->pose.pose.orientation.x,
+			odom->pose.pose.orientation.y,
+			odom->pose.pose.orientation.z);
```
vooon: Hmm, i'm more and more like idea of Vector3 message to Eigen inlines.
```
auto posParent = util::to_eigen(odom->pose.pose.position);
```

mavros_extras/src/plugins/odom.cpp
```diff
@@ -66,90 +67,97 @@ class OdometryPlugin : public plugin::PluginBase {
...
+
+		Eigen::Vector3d linVelNED;
+		Eigen::Vector3d angVelNED;
+		Eigen::Vector3d posNED;
+		Eigen::Quaterniond qParent2NED;
+
+		Eigen::Matrix<double, 6, 6> poseCovParent;
+		Eigen::Matrix<double, 9, 9> poseCovNED;
+		Eigen::Matrix<double, 6, 6> twistCovChild;
+		Eigen::Matrix<double, 3, 3> attCovNED;
```
vooon: Better to use Eigen::Matrix3d, 6d, 9d.

mavros_extras/src/plugins/odom.cpp
```diff
@@ -66,90 +67,97 @@ class OdometryPlugin : public plugin::PluginBase {
...
-		lpos.az = zerov3f.z();
+		lpos.x = posNED.x();
+		lpos.y = posNED.y();
+		lpos.z = posNED.z();
+		lpos.vx = linVelNED.x();
+		lpos.vy = linVelNED.y();
+		lpos.vz = linVelNED.z();
+		lpos.ax = 0;
+		lpos.ay = 0;
+		lpos.az = 0;
```
vooon: Generated code should be fixed in it's generator!

```
pip install --user cogapp
```

mavros_extras/src/plugins/odom.cpp
```diff
@@ -159,16 +167,18 @@ class OdometryPlugin : public plugin::PluginBase {
...
-		//     cog.outl("att.{b} = ang_vel_ned.{a}();".format(**locals()))
+		//     cog.outl("att.{b} = angVelNED.{a}();".format(**locals()))
 		// ]]]
-		att.rollspeed = ang_vel_ned.x();
-		att.pitchspeed = ang_vel_ned.y();
-		att.yawspeed = ang_vel_ned.z();
+		att.rollspeed = angVelNED.x();
+		att.pitchspeed = angVelNED.y();
+		att.yawspeed = angVelNED.z();
 		// [[[end]]] (checksum: e100d5c18a64c243df616f342f712ca1)
```
vooon: Generated code.

TSC21:
@vooon I already talked with James and we are going to generalize an API that can replace the static transforms with tf lookups that implement the rotation between frames. That will basically allow an easier rotation of translation and orientation whenever need besides allowing generalize the rotations WRT MAV_FRAMEs being transformed.

TSC21:
@jgoppert can you please split this into two separate PR's (one with the changes to the covariance matrix and the other for the frame transforms)? The first one is good to merge. The other would be a more general API change, that would require more work, in which I would help as we discussed.

TSC21:
@jgoppert I'm closing this PR. Please when you can redo a PR with the covariance values only. I can work on a future PR for doing the transforms through `tf2`.

Issue #444, opened 2015-11-27
Issue with Eigen::Vector3d UAS::quaternion_to_rpy(const Eigen::Quaterniond &q)


jespestana:
I suggest an alternative implementation to the function
`Eigen::Vector3d UAS::quaternion_to_rpy(const Eigen::Quaterniond &q)`
From the source file (I provide the link to the current commit/implementation for future reference):
[mavros/mavros/src/lib/uas_quaternion_utils.cpp](https://github.com/mavlink/mavros/blob/8cfdd89aa67e21a4a6758a5225bc3a8c184608ed/mavros/src/lib/uas_quaternion_utils.cpp#L41).

As a small background: the calculated yaw, pitch and roll are intrinsic rotations in the ZYX axes: (First) rotation of yaw around the Z axis, (Second) rotation of pitch around the Y axis of the moving frame and (Third) rotation of roll around the X axis of the moving frame.

The problem is that the EulerAngles implementation from Eigen, used in the current implementation of the function `quaternion_to_rpy`, provides the following output range for the {yaw, pitch, roll} angles: yaw in [0,pi], pitch in [-pi,+pi] and roll in [-pi,+pi].

I think that the most desirable output would be in these ranges for the {yaw, pitch, roll} angles: yaw in [-pi,pi], pitch in [-pi/2,+pi/2] and roll in [-pi,+pi]. This is even more the case given that the YPR-ZYX euler angles present the gimbal-lock at `pitch=-pi/2` and `pitch=+pi/2`. The calculation suggested in this wikipedia article [Conversion between quaternions and Euler angles](https://en.wikipedia.org/wiki/Conversion_between_quaternions_and_Euler_angles#Conversion) gives an output on this desirable ranges. Its use was already suggested in the code for the function
`double UAS::quaternion_get_yaw(const Eigen::Quaterniond &q)`, but somehow not used for `quaternion_to_rpy`.

The resulting code, which I already tested and only fails at the gimbal-lock (unavoidable problem), would be the following:

``` cpp
Eigen::Vector3d UAS::quaternion_to_rpy(const Eigen::Quaterniond &q)
{
    // Following the example of Vladimir Ermakov, I use the equations from this wikipedia article
    // https://en.wikipedia.org/wiki/Conversion_between_quaternions_and_Euler_angles
    // Note that the Eigen implementation is not good for YPR, because of the output range, which would be:
[...]

PR #761, opened 2017-07-24
Proposing fix to Eigen::Vector3d UAS::quaternion_to_rpy(const Eigen::Quaterniond &q). New implementation works in the range: yaw in [-pi,pi], pitch in [-pi/2,+pi/2] and roll in [-pi,+pi].

jespestana:
Pull request related to this issue (see explanation in the issue):
github issue https://github.com/mavlink/mavros/issues/444#issuecomment-291604800

Main modification to the function:
```
mavros/mavros/src/lib/uas_quaternion_utils.cpp
  Eigen::Vector3d UAS::quaternion_to_rpy(const Eigen::Quaterniond &q)
```

I have tested the new implementation with the test cases (only with minor modifications):
```
mavros/mavros/test/test_quaternion_utils.cpp
  TEST(UAS, quaternion_to_rpy__123)
  TEST(UAS, quaternion_to_rpy__full_yaw_range)
  TEST(UAS, quaternion_to_rpy__pm_pi)
```

I have added the changes so far only to the the indigo-devel branch, and tested them with: [Ubuntu 14.04], [ROS Indigo] and [mavros indigo-devel branch].

### DIFF
--- a/mavros/src/lib/uas_quaternion_utils.cpp
+++ b/mavros/src/lib/uas_quaternion_utils.cpp
@@ -34,27 +34,27 @@ Eigen::Quaterniond UAS::quaternion_from_rpy(const Eigen::Vector3d &rpy)
 
 Eigen::Vector3d UAS::quaternion_to_rpy(const Eigen::Quaterniond &q)
 {
-    // Following the example of Vladimir Ermakov, I use the equations from this wikipedia article
-    // https://en.wikipedia.org/wiki/Conversion_between_quaternions_and_Euler_angles
-		// This function is injective for the range: yaw in [-pi,pi], pitch in [-pi/2,+pi/2] and roll in [-pi,+pi].
-    // Note that the Eigen implementation is not good for YPR, because of the output range, which would be:
-		//   	[code using Eigen EulerAngles] rpy = q.toRotationMatrix().eulerAngles(2, 1, 0).reverse();
-    //    [in docs    Eigen EulerAngles] valid range of EulerAngles: yaw, pitch, roll in the ranges [0:pi]x[-pi:pi]x[-pi:pi]
-		//    see: https://eigen.tuxfamily.org/dox/group__Geometry__Module.html#gac3d90b12b21e1aaa2a9de9b0e45e6b7c
+	// Following the example of Vladimir Ermakov, I use the equations from this wikipedia article
+	// https://en.wikipedia.org/wiki/Conversion_between_quaternions_and_Euler_angles
+	// This function is injective for the range: yaw in [-pi,pi], pitch in [-pi/2,+pi/2] and roll in [-pi,+pi].
+	// Note that the Eigen implementation is not good for YPR, because of the output range, which would be:
+	//      [code using Eigen EulerAngles] rpy = q.toRotationMatrix().eulerAngles(2, 1, 0).reverse();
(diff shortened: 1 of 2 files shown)


TSC21:
@jespestana we are not accepting PR against `indigo-devel` anymore, as it is considered developed as is. So please do this PR against master. Also, when I talked about testing, I was talking about flight testing (real or in simulation), not using the unittests. The flight tests are the ones that would give consistency to the the added code and validates it.

TSC21:
The main concern is in https://github.com/mavlink/mavros/blob/df9f38067d4d80aeda839cc3d71b0a7eed67327a/mavros_extras/src/plugins/vision_pose_estimate.cpp#L127, where the method is used. So a test in a vision/mocap system with and without this modification should be done.

jespestana:
> we are not accepting PR against indigo-devel anymore, as it is considered developed as is. So please do this PR against master.

@TSC21 Ok, I will perform the PR in the master branch. 

> Also, when I talked about testing, I was talking about flight testing (real or in simulation), not using the unittests. The flight tests are the ones that would give consistency to the the added code and validates it. 
> The main concern is in https://github.com/mavlink/mavros/blob/df9f38067d4d80aeda839cc3d71b0a7eed67327a/mavros_extras/src/plugins/vision_pose_estimate.cpp#L127, where the method is used. So a test in a vision/mocap system with and without this modification should be done.

Ok, this makes total sense. Thanks for explaining what is required. 

I will issue the PR after doing these tests. For the test: I think that flying a simple trajectory, but with a full rotation in yaw heading would do.

TSC21:
> I will issue the PR after doing these tests. For the test: I think that flying a simple trajectory, but with a full rotation in yaw heading would do.

Will you be flying with offboard control and a mocap/vision system that send position estimates through `vision_pose_estimate` plugin? It only makes sense like that.

jespestana:
> Will you be flying with offboard control and a mocap/vision system that send position estimates through vision_pose_estimate plugin? It only makes sense like that.

@TSC21 We have an optitrack system in our lab. We would feedback the position controller that is inside the autopilot using vision_pose_estimate, calculated from the optitrack pose estimates.

In order to use the `vision_pose_estimate` plugin, we just need to send the pose estimates through the `pose` topic, to which the `vision_pose_estimate` plugin is subscribed, right?
tf_listen parameter `false` and using this topic: https://github.com/mavlink/mavros/blob/master/mavros_extras/src/plugins/vision_pose_estimate.cpp#L60

TSC21:
> @TSC21 We have an optitrack system in our lab. We would feedback the position controller that is inside the autopilot using vision_pose_estimate, calculated from the optitrack pose estimates.
In order to use the vision_pose_estimate plugin, we just need to send the pose estimates through the pose topic, to which the vision_pose_estimate plugin is subscribed, right?
tf_listen parameter false and using this topic: https://github.com/mavlink/mavros/blob/master/mavros_extras/src/plugins/vision_pose_estimate.cpp#L60

That is correct. Make sure that you are feeding the data on ENU frame on the topic, so that MAVROS transforms it to NED. Please post some plots of the the orientation given by `VISION_POSITION_ESTIMATE` vs the attitude feedback you get in the QGC and make sure they are consistent. Would be good to make a comparison between with and without this modification of yours.

TSC21:
@jespestana did you make progress on this? We are still waiting for that PR :) Thanks!

TSC21:
@jespestana friendly ping :)

jgoppert:
@tsc21 I assume this never got merged? I just ran into it using Qualisys.

PR #776, opened 2017-08-19
build: make geoid datasets install mandatory

TSC21:
This obliges people running the dataset install script (`install_geographiclib_dataset.sh`) in order to avoid errors on height calculation, which was only launching an exception but no sigfault.


TSC21:
This CI system is maaadd...

mavros/CMakeLists.txt
```diff
@@ -39,6 +39,27 @@ endif()
...
+if(NOT GEOGRAPHICLIB_GRAVITY_PATH)
+  message(WARNING "No gravity field model datasets found. Please execute the script install_geographiclib_dataset.sh in /mavros/scripts")
+else()
+  message(STATUS "Gravity Field model datasets found in: " ${GEOGRAPHICLIB_GRAVITY_PATH}/share/GeographicLib/gravity)
+endif()
+if(NOT GEOGRAPHICLIB_MAGNETIC_PATH)
+  message(WARNING "No magnetic field model datasets found. Please execute the script install_geographiclib_dataset.sh in /mavros/scripts")
+else()
+  message(STATUS "Magnetic Field model datasets found in: " ${GEOGRAPHICLIB_MAGNETIC_PATH}/share/GeographicLib/magnetic)
+endif()
```
vooon: NONONO! All that block should not be here!
Should be a runtime error!

PR #778, opened 2017-08-20
geolib: datasets: warn when not installed; update install script; launch SIGINT when not installed

TSC21:


TSC21:
OK Travis now builds right for the inclusion of the datasets. Two things missing now:
- [ ] Update the  `install_geographiclib_datasets.sh` script in order to support Ubuntu 14.04 version of `geographiclib-tools`, which script for downloading the datasets (only geoid mode) is `geographiclib-datasets-download`
- [ ] Use a customized docker image for Jenkins to build when checking for the datasets.

TSC21:
I'm reconsidering this as just a warning, as putting as mandatory would probably be impossible on the rosbuild farm and on a release.

TSC21:
I'm going to leave the docker container update for another PR. This as is seems enough to fix travis and at the same time add the dataset load as mandatory. @vooon can you review please?

libmavconn/cmake/Modules/CheckGeographicLibDatasets.cmake
```diff
@@ -0,0 +1,27 @@
...
+else()
+  message(STATUS "Gravity Field model datasets found in: " ${GEOGRAPHICLIB_GRAVITY_PATH_}/gravity)
+  set(GEOGRAPHICLIB_GRAVITY_PATH ${GEOGRAPHICLIB_GRAVITY_PATH_}/gravity)
+endif()
+if(NOT GEOGRAPHICLIB_MAGNETIC_PATH_)
+  message(WARNING "No magnetic field model datasets found. Please execute the script install_geographiclib_dataset.sh in /mavros/scripts")
+else()
+  message(STATUS "Magnetic Field model datasets found in: " ${GEOGRAPHICLIB_MAGNETIC_PATH_}/magnetic)
+  set(GEOGRAPHICLIB_MAGNETIC_PATH ${GEOGRAPHICLIB_MAGNETIC_PATH_}/magnetic)
+endif()
```
vooon: NO! Wrong! You should check in runtime. UAS or so should try to load needed datasets and responds error log if it fails.

TSC21: Wrong? Wrong why? Check the all PR and you will see that's also done. This is a check during build time that warns people that they basically did not fully followed the installation instructions. A SIGINT is launched when that happens. What I can also do is verbose the SIGINT with a info msg so that people run the install script. Besides that, this is perfectly acceptable IMO and doesn't harm anyone.

.travis.yml
```diff
@@ -26,7 +31,7 @@ before_install:
   #- sudo sh -c "echo \"deb http://packages.ros.org/ros-shadow-fixed/ubuntu $ROS_CI_DESKTOP main\" > /etc/apt/sources.list.d/ros-latest.list"
   - wget http://packages.ros.org/ros.key -O - | sudo apt-key add -
   # Update python 2.7 version
-  - sudo apt-add-repository ppa:fkrull/deadsnakes-python2.7 -y
+  - sudo apt-add-repository ppa:fkrull/deadsnakes -y
```
khancyr: ppa deprecatead : use https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa
Why do you need it btw ?

TSC21: Problems in Travis. But thanks for the updated version

mavros/src/lib/uas_data.cpp
```diff
@@ -40,13 +47,15 @@ UAS::UAS() :
...
+		// Using smallest dataset with 5' grid,
 		// From default location,
 		// Use cubic interpolation, Thread safe
 		egm96_5 = std::make_shared<GeographicLib::Geoid>("egm96-5", "", true, true);
 	}
 	catch (const std::exception &e) {
+		// catch exception and shutdown node
 		ROS_ERROR_STREAM("UAS: GeographicLib exception: " << e.what());
+		signal(SIGINT, term);
 	}
```
vooon: No! Why you set signal callback here? Just call `ROS_ABORT()`.

TSC21: Well I wasn't sure what was the best approach. I know that SIGINT can be intercepted besised allowing proper debugging. I was thinking that we could added the info on the SIGINT stating that the problem is due to not having the dataset installed. Though I suppose that can also be handled on the ROS_ERROR_STREAM above.

vooon: But that it not what you wrote in code.
Code:
1. try to load egm96-5
2. on error:
    1. Log error
    2. Set SIGINT signal handler. `man signal`

PR #790, opened 2017-08-25
CMake: Fix link to atomic library in armhf and mips arch

TSC21:
Fixes https://github.com/bmwcarit/meta-ros/issues/525.
@bulwahn can you test it and check if this addresses and fixes the problem? Thanks

### DIFF
--- a/mavros/CMakeLists.txt
+++ b/mavros/CMakeLists.txt
@@ -45,6 +45,12 @@ include(CheckGeographicLibDatasets)
 include(EnableCXX11)
 include(MavrosMavlink)
 
+## Fix link to atomic library in armhf and mips arch
+set(TARGET_ARCH ${CMAKE_HOST_SYSTEM_PROCESSOR})
+if (${CMAKE_HOST_SYSTEM_PROCESSOR} MATCHES "armhf" OR ${CMAKE_HOST_SYSTEM_PROCESSOR} MATCHES "mips")
+    set(CMAKE_CXX_LINK_FLAGS "${CMAKE_CXX_LINK_FLAGS} -latomic")
+endif()
+
 # detect if sensor_msgs has BatteryState.msg
 # http://answers.ros.org/question/223769/how-to-check-that-message-exists-with-catkin-for-conditional-compilation-sensor_msgsbatterystate/
 list(FIND sensor_msgs_MESSAGE_FILES "msg/BatteryState.msg" BATTERY_STATE_MSG_IDX)


TSC21:
@bulwahn friendly ping

bulwahn:
@TSC21 I proposed an alternative solution in #797, which I came up with in parallel to your solution on last Friday. Maybe you can have a look, consider which you would find better and test #797 as well, esp. if it breaks your default build setups. I will test today if this pull request also resolves the issue in meta-ros.

bulwahn:
@TSC21 I tested your commit applied in meta-ros, and it remains to fail with:
```
| devel/lib/libmavros.so: undefined reference to `__atomic_load_8'
| devel/lib/libmavros.so: undefined reference to `__atomic_store_8'
| collect2: error: ld returned 1 exit status
```
for the arm architecture.
I am not sure, but I believe checking the HOST architecture might be ineffective in the cross-compiliation where you would actually want to check the TARGET architecture instead to set the flags accordingly. Rather than looking deeper into your solution now, I would be interested in your opinion of #797.
For the record, this is the build configuration I tested the commit with:
```
Build Configuration:
BB_VERSION        = "1.35.0"
BUILD_SYS         = "x86_64-linux"
NATIVELSBSTRING   = "ubuntu-14.04"
TARGET_SYS        = "arm-oe-linux-gnueabi"
MACHINE           = "qemuarm"
DISTRO            = "nodistro"
DISTRO_VERSION    = "nodistro.0"
TUNE_FEATURES     = "arm armv5 thumb dsp"
TARGET_FPU        = "soft"
meta              = "HEAD:82ea78bd2f03b6ba4d720595d3a3fbd96e0232f9"
meta-oe           
meta-multimedia   
meta-python       = "master:ac2a6d2b5d69937577effcb8bb2149651bac0176"
meta-ros          = "mavros-solution-TSC21:8dcc2ba0c5f050dfd41c10dc8a5a9ee639f9d58c"
```

TSC21:
I see your point. But I would like to this being added to the CMakeList and only be added the flag when required for the target architecture. Maybe changing from the HOST to the TARGET arch would work? I will add that change, and if it works we can merge this. If not, we can accept your PR.

TSC21:
@bulwahn I think I found a solution to check for the TARGET arch. Will replace the code of this PR soon.

bulwahn:
@TSC21 Once a new commit is available, just ping me and run tests in the cross-compile tool chain to check if it works. By the way, the issue occurs with arm, mips and ppc, not just arm and mips.

TSC21:
https://github.com/axr/solar-cmake/blob/master/TargetArch.cmake seems to aid on the solution. The trouble is: there are so many archs that can be linked to the `-march` compile variable, so I suppose a more general explicit link would be enough, as adding this wouldn't be cost-beneficial. I'm closing this in favor of #797.

Issue #789, opened 2017-08-25
Consistency problem in the TF frames (in px4_config.yaml)


AlexisTM:
### Issue details

The default TF frames are not consistent in the configuration file.

Within the file, either `local_origin` and `map` are used for the same frame, except for local_position and global_position.

Also the `global_position/frame_id` and `local_position/frame_id` shouldn't point to fcu.

https://github.com/mavlink/mavros/blob/d283fb0a3f094caa4282922351fb965f69f9c898/mavros/launch/px4_config.yaml#L45-L55

* `global_position/frame_id` should be `earth` (**bug**)
* `global_position/tf/frame_id` should be `map` (**consistency**)

https://github.com/mavlink/mavros/blob/d283fb0a3f094caa4282922351fb965f69f9c898/mavros/launch/px4_config.yaml#L65-L72

* `local_position/frame_id` should be `map` (**bug**)
* `local_position/tf/frame_id` should be `map` (**consistency**)

TSC21:
@AlexisTM thanks for reviewing this out. This was something to be done some time ago but wasn't done yet.

> Within the file, either `local_origin` and `map` are used for the same frame, except for local_position and global_position.

Should all be replaced by `map` as `local_origin` is not a frame of reference of MAVROS.

> * `global_position/frame_id` should be `earth` (**bug**)

Should be `map` also. `earth` is for the global frame, though the pose published by the plugin is relative to the `map` frame.

TSC21:
Would you mind applying a PR with the above corrections? Thanks

TSC21:
Fixed with #791

PR #791, opened 2017-08-25
Solving default frame consistency in config files

AlexisTM:
Solving default frame consistency in config files #789 

* Changing the `local_origin` frame to `map` for consistency.
* Correction of the `global_position/frame_id` and `local_position/frame_id` to map frame

### DIFF
--- a/mavros/launch/apm_config.yaml
+++ b/mavros/launch/apm_config.yaml
@@ -19,7 +19,7 @@ sys:
 
 # sys_time
 time:
-  time_ref_source: "base_link"  # time_reference source
+  time_ref_source: "fcu"  # time_reference source
   timesync_mode: NONE
   timesync_avg_alpha: 0.6 # timesync averaging factor
 
@@ -50,6 +50,7 @@ global_position:
   tf:
     send: false               # send TF?
     frame_id: "map"  # TF frame_id
+    global_frame_id: "earth"  # TF earth frame_id
     child_frame_id: "base_link" # TF child_frame_id
 
 # imu_pub
--- a/mavros/launch/apm_config.yaml
+++ b/mavros/launch/apm_config.yaml
@@ -69,6 +69,7 @@ local_position:
     send: false
     frame_id: "map"
     child_frame_id: "base_link"
+    send_fcu: false
 
 # param
 # None, used for FCU params
--- a/mavros/launch/apm_config.yaml
+++ b/mavros/launch/apm_config.yaml
@@ -44,12 +44,12 @@ cmd:
 
 # global_position
 global_position:
-  frame_id: "fcu"           # pose and fix frame_id
+  frame_id: "map"           # pose and fix frame_id
   rot_covariance: 99999.0   # covariance for attitude?
   use_relative_alt: true    # use relative altitude for local coordinates
   tf:
     send: false               # send TF?
-    frame_id: "local_origin"  # TF frame_id
+    frame_id: "map"  # TF frame_id
     child_frame_id: "fcu_utm" # TF child_frame_id
 
 # imu_pub
@@ -63,10 +63,10 @@ imu:
 
 # local_position
 local_position:
-  frame_id: "fcu"
+  frame_id: "map"
   tf:
     send: false
-    frame_id: "local_origin"
+    frame_id: "map"
     child_frame_id: "fcu"
 
 # param
--- a/mavros/launch/apm_config.yaml
+++ b/mavros/launch/apm_config.yaml
@@ -169,7 +169,7 @@ fake_gps:
 # px4flow
 px4flow:
   frame_id: "px4flow"
-  ranger_fov: !degrees 0.0  # XXX TODO
+  ranger_fov: !degrees 6.8  # XXX TODO
   ranger_min_range: 0.3     # meters
   ranger_max_range: 5.0     # meters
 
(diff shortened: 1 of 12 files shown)


AlexisTM:
I updated the APM config as well. @TSC21

TSC21:
Where you see `fcu_utm` please update to `fcu`

AlexisTM:
Should check for the time reference. 

Is this related to tf?

```
time:
  time_ref_source: "fcu"  # time_reference source
```

TSC21:
> Should check for the time reference. 

Keep it as `fcu`. That has nothing to do with frames of reference.

mavros/launch/px4_config.yaml
```diff
@@ -44,18 +44,18 @@ cmd:
...
 global_position:
-  frame_id: "fcu"           # pose and fix frame_id
+  frame_id: "map"           # pose and fix frame_id
   rot_covariance: 99999.0   # covariance for attitude?
   use_relative_alt: true    # use relative altitude for local coordinates
   tf:
     send: false               # send TF?
-    frame_id: "local_origin"  # TF frame_id
+    frame_id: "map"  # TF frame_id
     global_frame_id: "earth"  # TF earth frame_id
```
TSC21: Can you add `global_frame_id` also to apm_config please?

TSC21:
@AlexisTM looking good but can you please run a sanity check on the source code also and match the variables, that represent the parameters, to the config files? Basically the same you did above but for the cpp files. Thanks

AlexisTM:
The only parameters not synced are in `sys_status.cpp`

https://github.com/mavlink/mavros/blob/d283fb0a3f094caa4282922351fb965f69f9c898/mavros/src/plugins/sys_status.cpp#L431-L433

* `global_position/tf/send` default to `false`
* `imu, checked
* `local_position/tf/send` default to `false`
* `local_position/tf/send_fcu` default to `false`
* `mission/pull_after_gcs` default to `true`

TSC21:
> The only parameters not synced are in `sys_status.cpp`

Can you sync with the values that are on the `.cpp` file, please?

> Note that the builds are failing because of the MAVLink signing development.

Yeah we need that the release comes out on the deb pakages.

AlexisTM:
* Added sys_status, and mavros_extras plugins.
* Added the FOV for the Px4flow (6.8 degrees at 5 meters)
* Adding the copter_visualization configuration in the yaml file

TSC21:
@AlexisTM `copter_visualization` is a separate node that can be run in parallel. So the parameters are  not loaded from the config files of the `mavros` node. The parameter config should got: or to a `launch` file that launches the `copter_visualization` node, or to a `yaml` file on the launch folder and getting the parameters on the cpp file through `XmlRpc::XmlRpcValue` (check how it is done on `servo_state_publisher.cpp`).
The rest of the changes though seem consistent. Thanks

TSC21:
Also, can you please change the `vibration` `frame_id` on both plugin and yaml to `base_link`? Thanks

AlexisTM:
I think it is all fine with the way the parameters are read for the `copter_visualization`. Indeed, there are only a finite number of parameters, I would propose to avoid the use of `XmlRpc::XmlRpcValue`. 

Setting the parameters in a yaml file or the launch file is sufficient.

mavros/launch/apm_config.yaml
```diff
@@ -167,7 +169,7 @@ fake_gps:
 # px4flow
 px4flow:
   frame_id: "px4flow"
-  ranger_fov: !degrees 0.0  # XXX TODO
+  ranger_fov: !degrees 6.8  # XXX TODO
```
TSC21: You can remove the `TODO` comment now

TSC21: `6.8 degreens at 5 meters, 31 degrees at 1 meter`

mavros_extras/launch/px4flow_config.yaml
```diff
@@ -25,7 +25,7 @@ image:
 
 px4flow:
   frame_id: "px4flow"
-  ranger_fov: !degrees 0.0  # XXX TODO
+  ranger_fov: !degrees 6.8  # XXX TODO
```
TSC21: `6.8 degreens at 5 meters, 31 degrees at 1 meter`

mavros_extras/src/plugins/px4flow.cpp
```diff
@@ -45,7 +45,9 @@ class PX4FlowPlugin : public plugin::PluginBase {
 		flow_nh.param<std::string>("frame_id", frame_id, "px4flow");
 
 		/** Default rangefinder is Maxbotix HRLV-EZ4 */
-		flow_nh.param("ranger_fov", ranger_fov, 0.0);	/** @todo Check MAxbotix HRLV-EZ4 Field-of-View */
+		flow_nh.param("ranger_fov", ranger_fov, 0.119428926);	/** @todo Check MAxbotix HRLV-EZ4 Field-of-View
```
TSC21: Replace with
```c++
/**
 * @note This is a narrow beam (60cm wide at 5 meters,
 * but also at 1 meter) 6.8 degrees at 5 meters, 31 degrees at 1 meter
 */
flow_nh.param("ranger_fov", ranger_fov, 0.119428926);
```

AlexisTM:
You are welcome. ;)

PR #798, opened 2017-08-30
Partial waypoint uploading support and bug fix in waypoint uploading

JMare:
This PR adds support for partial waypoint uploading with a new service including support in mavwp cli script.

Also fixed a bug we uncovered when uploading missions to APM.

When state is TXWP and mission item N is received by FCU, FCU sends request for N+1.
If that request packet gets dropped, mavros will resend mission_item N.
At that point FCU will be expecting N+1 and mavros will be sending N.
FCU responds by sending a mission_ack of type "invalid_sequence".
This causes mavros to abort the transfer and go_idle() even though the FCU is still requesting N+1.

This PR makes mavros stay in TXWP state when that particular mission_ack is received, so the transfer will complete when the autopilot resends the request for N+1.

### DIFF
--- a/mavros/scripts/mavwp
+++ b/mavros/scripts/mavwp
@@ -137,8 +137,11 @@ def do_load_partial(args):
     with args.file:
         wps = [w for w in wps_file.read(args.file)]
 
+    if args.start == args.end:
+        args.end += 1
+
     try:
-        ret = M.push_partial(waypoints=wps[args.start:args.end+1], start_index=args.start)
+        ret = M.push_partial(waypoints=wps[args.start:args.end], start_index=args.start)
     except rospy.ServiceException as ex:
         fault(ex)
 
--- a/mavros/scripts/mavwp
+++ b/mavros/scripts/mavwp
@@ -137,11 +137,11 @@ def do_load_partial(args):
     with args.file:
         wps = [w for w in wps_file.read(args.file)]
 
-    if args.start == args.end:
-        args.end += 1
+    if args.start_index == args.end_index:
+        args.end_index += 1
 
     try:
-        ret = M.push_partial(waypoints=wps[args.start:args.end], start_index=args.start)
+        ret = M.push_partial(waypoints=wps[args.start_index:args.end_index], start_index=args.start_index)
     except rospy.ServiceException as ex:
         fault(ex)
 
@@ -209,9 +209,9 @@ def main():
 
     load_partial_args = subarg.add_parser('load_partial', help="partially load waypoints from file")
     load_partial_args.set_defaults(func=do_load_partial)
-    load_partial_args.add_argument('start', type=int, help="waypoint start id")
-    load_partial_args.add_argument('end', type=int, help="waypoint end id")
     load_partial_args.add_argument('file', type=argparse.FileType('rb'), help="input file (QGC/MP format)")
+    load_partial_args.add_argument('start_index', type=int, help="waypoint start index")
+    load_partial_args.add_argument('end_index', type=int, help="waypoint end index")
 
     pull_args = subarg.add_parser('pull', help="pull waypoints from FCU")
     pull_args.set_defaults(func=do_pull)
--- a/mavros/scripts/mavwp
+++ b/mavros/scripts/mavwp
@@ -131,6 +131,23 @@ def do_load(args):
             _load_call(wps)
 
 
+def do_load_partial(args):
+    wps = []
+    wps_file = get_wp_file_io(args)
+    with args.file:
+        wps = [w for w in wps_file.read(args.file)]
+
+    try:
+        ret = M.push_partial(waypoints=wps[args.start:args.end+1], start_index=args.start)
+    except rospy.ServiceException as ex:
+        fault(ex)
+
+    if not ret.success:
+        fault("Request failed. Check mavros logs")
+
+    print_if(args.verbose, "Waypoints transfered.")
+
+
 def do_dump(args):
     done_evt = threading.Event()
     def _write_file(topic):
@@ -186,6 +203,12 @@ def main():
     load_args.set_defaults(func=do_load)
     load_args.add_argument('-p', '--preserve-home', action='store_true', help="Preserve home location (WP 0, APM only)")
     load_args.add_argument('file', type=argparse.FileType('rb'), help="input file (QGC/MP format)")
+
+    load_partial_args = subarg.add_parser('load_partial', help="partially load waypoints from file")
+    load_partial_args.set_defaults(func=do_load_partial)
+    load_partial_args.add_argument('start', type=int, help="waypoint start id")
+    load_partial_args.add_argument('end', type=int, help="waypoint end id")
+    load_partial_args.add_argument('file', type=argparse.FileType('rb'), help="input file (QGC/MP format)")
 
     pull_args = subarg.add_parser('pull', help="pull waypoints from FCU")
     pull_args.set_defaults(func=do_pull)
--- a/mavros/scripts/mavwp
+++ b/mavros/scripts/mavwp
@@ -148,7 +148,7 @@ def do_load_partial(args):
     if not ret.success:
         fault("Request failed. Check mavros logs")
 
-    print_if(args.verbose, "Waypoints transfered.")
+    print_if(args.verbose, "Waypoints transfered:", ret.wp_transfered)
 
 
 def do_dump(args):
--- a/mavros/scripts/mavwp
+++ b/mavros/scripts/mavwp
@@ -25,7 +25,6 @@ except ImportError:
     no_prettytable = True
 
 
-
 def get_wp_file_io(args):
     return M.QGroundControlWP()
 
@@ -96,9 +95,9 @@ def do_load(args):
     with args.file:
         wps = [w for w in wps_file.read(args.file)]
 
-    def _load_call(waypoint_list):
+    def _load_call(start, waypoint_list):
         try:
-            ret = M.push(waypoints=waypoint_list)
+            ret = M.push(start_index=start, waypoints=waypoint_list)
         except rospy.ServiceException as ex:
             fault(ex)
 
@@ -118,37 +117,28 @@ def do_load(args):
 
         done_evt.set()
 
-    if not args.preserve_home:
-        _load_call(wps)
-    else:
-        # Note: _load_call() emit publish on this topic, so callback only changes
-        # waypoint 0, and signal done event.
-        sub = M.subscribe_waypoints(_fix_wp0)
-        if not done_evt.wait(30.0):
-            fault("Something went wrong. Topic timed out.")
+    if args.start_index > 0:
+        # Push partial
+        if args.start_index == args.end_index:
+            args.end_index += 1
+
+        if args.end_index == 0:
+            _load_call(args.start_index, wps[args.start_index:])
         else:
-            sub.unregister()
-            _load_call(wps)
-
-
-def do_load_partial(args):
-    wps = []
-    wps_file = get_wp_file_io(args)
-    with args.file:
-        wps = [w for w in wps_file.read(args.file)]
-
-    if args.start_index == args.end_index:
-        args.end_index += 1
-
-    try:
-        ret = M.push_partial(waypoints=wps[args.start_index:args.end_index], start_index=args.start_index)
-    except rospy.ServiceException as ex:
-        fault(ex)
-
-    if not ret.success:
-        fault("Request failed. Check mavros logs")
-
-    print_if(args.verbose, "Waypoints transfered:", ret.wp_transfered)
+            _load_call(args.start_index, wps[args.start_index:args.end_index])
+    else:
+        # Full push
+        if not args.preserve_home:
+            _load_call(0, wps)
+        else:
+            # Note: _load_call() emit publish on this topic, so callback only changes
+            # waypoint 0, and signal done event.
+            sub = M.subscribe_waypoints(_fix_wp0)
+            if not done_evt.wait(30.0):
+                fault("Something went wrong. Topic timed out.")
+            else:
+                sub.unregister()
+                _load_call(0, wps)
 
 
 def do_dump(args):
@@ -205,13 +195,9 @@ def main():
     load_args = subarg.add_parser('load', help="load waypoints from file")
     load_args.set_defaults(func=do_load)
     load_args.add_argument('-p', '--preserve-home', action='store_true', help="Preserve home location (WP 0, APM only)")
+    load_args.add_argument('-s', '--start-index', type=int, default=0, help="waypoint start index for partial updating")
+    load_args.add_argument('-e', '--end-index', type=int, default=0, help="waypoint end index for partial updating")
     load_args.add_argument('file', type=argparse.FileType('rb'), help="input file (QGC/MP format)")
-
-    load_partial_args = subarg.add_parser('load_partial', help="partially load waypoints from file")
-    load_partial_args.set_defaults(func=do_load_partial)
-    load_partial_args.add_argument('file', type=argparse.FileType('rb'), help="input file (QGC/MP format)")
-    load_partial_args.add_argument('start_index', type=int, help="waypoint start index")
-    load_partial_args.add_argument('end_index', type=int, help="waypoint end index")
 
     pull_args = subarg.add_parser('pull', help="pull waypoints from FCU")
     pull_args.set_defaults(func=do_pull)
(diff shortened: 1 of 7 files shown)


JMare:
@TSC21 Sure I can make a video tomorrow of partial waypoint uploading to APM SITL.

One of the CI checks failed, do you know what tests it runs? I ran catkin run_tests and it seemed ok.

mavros/scripts/mavwp
```diff
@@ -187,6 +207,12 @@ def main():
     load_args.add_argument('-p', '--preserve-home', action='store_true', help="Preserve home location (WP 0, APM only)")
     load_args.add_argument('file', type=argparse.FileType('rb'), help="input file (QGC/MP format)")
 
+    load_partial_args = subarg.add_parser('load_partial', help="partially load waypoints from file")
+    load_partial_args.set_defaults(func=do_load_partial)
+    load_partial_args.add_argument('file', type=argparse.FileType('rb'), help="input file (QGC/MP format)")
+    load_partial_args.add_argument('start_index', type=int, help="waypoint start index")
+    load_partial_args.add_argument('end_index', type=int, help="waypoint end index")
```
vooon: I think that indexes better set via options, not positional args. Start index should be `required=True`; End - by default use last entry.

mavros/scripts/mavwp
```diff
@@ -187,6 +207,12 @@ def main():
     load_args.add_argument('-p', '--preserve-home', action='store_true', help="Preserve home location (WP 0, APM only)")
     load_args.add_argument('file', type=argparse.FileType('rb'), help="input file (QGC/MP format)")
 
+    load_partial_args = subarg.add_parser('load_partial', help="partially load waypoints from file")
```
vooon: Name it `load-partial`.

TSC21:
> One of the CI checks failed, do you know what tests it runs? I ran catkin run_tests and it seemed ok.

Nothing to worry about the CI. This is a Jenkins related problem ;)

TSC21:
@JMare any updates?

TSC21:
Jenkins test this please.

JMare:
@TSC21 We implemented the changes @vooon suggested.
In doing so we realised that our implementation was leaving send_waypoints in a bad state so we are trying to fix that.

Our latest work is in https://github.com/MonashUAS/mavros/tree/partial_tmp
Should have it done in a couple of days.

TSC21:
@JMare perfect thank you! Do you have a way of testing the same with an autopilot using PX4 Firmware?

JMare:
@TSC21 I have never tried but PX4 SITL looks easy enough since I am familiar with gazebo.
I'll give it a go tommorow.

TSC21:
> @TSC21 I have never tried but PX4 SITL looks easy enough since I am familiar with gazebo.
I'll give it a go tommorow.

Great thank you!

JMare:
@JamesStewy Managed to iron out the bugs.

Having a hard time with PX4 Sitl, going to just flash a Pixhawk to test over the next few days.

TSC21:
What problems are you facing?

JMare:
@TSC21 I had build errors on the gazebo plugins.
I am using a non standard build of gazebo though.

Today I tested with a Pixhawk running PX4 Firmware and partial way point updating does not work.

Doing some searching and looking through https://github.com/PX4/Firmware/blob/master/src/modules/mavlink/mavlink_mission.cpp

It does not look like PX4 Firmware implements partial waypoint updating at all. 
This feature may have to be APM specific.

TSC21:
Ok just be sure it does not break the normal waypoiny feed to the PX4 Firmware then. Please also document it does not currently work for the PX4 Firmware for lack of support on the Firmware side. Thanks

JMare:
Yes I checked and was still able to waypoint load/clear/pull as normal so it doesn't break that.

Should we make the feature enabled only for APM FCU as specified in launch files?

TSC21:
Yes please. If you may, please open an issue on the PX4 firmware as a feature request, so to be tracked.

mavros/src/plugins/waypoint.cpp
```diff
@@ -147,6 +147,7 @@ class WaypointPlugin : public plugin::PluginBase {
 		wp_state = WP::IDLE;
 
 		wp_nh.param("pull_after_gcs", do_pull_after_gcs, false);
+        wp_nh.param("enable_partial_push",enable_partial_push,false);
```
TSC21: use tab for indentation

mavros/src/plugins/waypoint.cpp
```diff
@@ -215,6 +216,8 @@ class WaypointPlugin : public plugin::PluginBase {
 	ros::Timer wp_timer;
 	ros::Timer shedule_timer;
 	bool do_pull_after_gcs;
+    bool enable_partial_push;
```
TSC21: use tab for indentation

mavros/src/plugins/waypoint.cpp
```diff
@@ -736,6 +739,14 @@ class WaypointPlugin : public plugin::PluginBase {
...
 		{
 			//Partial Waypoint update
+
+            if(!enable_partial_push)
+            {
+                ROS_WARN_NAMED("wp","Partial Push not enabled. (Only supported on APM)");
+                res.success = false;
+                res.wp_transfered = 0;
+                return true;
+            }
```
TSC21: don't use spaces. Use tab for indentation (equivalent to 8 spaces).

TSC21: Or just run the uncrustify tool on the file (`uncrustify -c ${ROS_WORKSPACE}/mavros/mavros/tools/uncrustify-cpp.cfg --replace --no-backup <path/to/file.ext>`)

TSC21:
Two last things:
1. Solve the conflict on the .cpp file;
2. Apply the Doxygen-like commenting for documentation on the full file, as in https://github.com/mavlink/mavros/blob/master/mavros/src/plugins/imu.cpp, if you may. (It will help me out with https://github.com/mavlink/mavros/issues/802). Thanks

vooon:
Hmm, unfortunately i do not see partial mission in `MAV_PROTOCOL_CAPABILITY`.
But you may assume that if FCU is APM then that feature is supported.

mavros/scripts/mavwp
```diff
@@ -118,17 +117,28 @@ def do_load(args):
...
-        _load_call(wps)
+    if args.start_index > 0:
+        # Push partial
+        if args.start_index == args.end_index:
+            args.end_index += 1
+
+        if args.end_index == 0:
+            _load_call(args.start_index, wps[args.start_index:])
+        else:
+            _load_call(args.start_index, wps[args.start_index:args.end_index])
```
vooon:
```python
end_index = args.end_index or len(wps)
_load_call(args.start_index, wps[args.start_index:end_index])
```

mavros/scripts/mavwp
```diff
@@ -185,6 +195,8 @@ def main():
     load_args = subarg.add_parser('load', help="load waypoints from file")
     load_args.set_defaults(func=do_load)
     load_args.add_argument('-p', '--preserve-home', action='store_true', help="Preserve home location (WP 0, APM only)")
+    load_args.add_argument('-s', '--start-index', type=int, default=0, help="waypoint start index for partial updating")
+    load_args.add_argument('-e', '--end-index', type=int, default=0, help="waypoint end index for partial updating")
```
vooon: Add to help "(default: last element in waypoint list)"

mavros/src/plugins/waypoint.cpp
```diff
@@ -147,6 +147,7 @@ class WaypointPlugin : public plugin::PluginBase {
 		wp_state = WP::IDLE;
 
 		wp_nh.param("pull_after_gcs", do_pull_after_gcs, false);
+		wp_nh.param("enable_partial_push",enable_partial_push,false);
```
vooon: Better:

```
//! @note currently (2017) only APM support partial mission update
wp_nh.param("enable_partial_push", enable_partial_push, m_uas->is_ardupilotmega());
```

TSC21:
> But you may assume that if FCU is APM then that feature is supported.

Yeah I suppose that instead of a parameter we can check for the FCU type by checking `UAS::MAV_AUTOPILOT`.

mavros/src/plugins/waypoint.cpp
```diff
@@ -649,18 +663,30 @@ class WaypointPlugin : public plugin::PluginBase {
...
+	void mission_write_partial_list(uint16_t start_index, uint16_t end_index)
+	{
+		ROS_DEBUG_NAMED("wp", "WP:m: write partial list %u - %u", start_index, end_index);
+
+		mavlink::common::msg::MISSION_WRITE_PARTIAL_LIST mwpl {};
+		mwpl.start_index = start_index;
+		mwpl.end_index = end_index;
+
+		m_uas->msg_set_target(mwpl);
+		UAS_FCU(m_uas)->send_message_ignore_drop(mwpl);
```
vooon: `msg_set_target()` should be in msg fill group of lines.

mavros/src/plugins/waypoint.cpp
```diff
@@ -702,39 +728,86 @@ class WaypointPlugin : public plugin::PluginBase {
...
 	{
 		unique_lock lock(mutex);
 
 		if (wp_state != WP::IDLE)
 			// Wrong initial state, other operation in progress?
 			return false;
 
-		wp_state = WP::TXLIST;
+		if (req.start_index)
+		{
```
vooon: Style:
```
if () {
}
else if () {
}
else {
}
```

Unfortunately uncrustify do not fix that.

mavros/src/plugins/waypoint.cpp
```diff
@@ -702,39 +728,86 @@ class WaypointPlugin : public plugin::PluginBase {
...
+				return true;
+			}
+
+			if (waypoints.size() < req.start_index + req.waypoints.size())
+			{
+				ROS_WARN_NAMED("wp","WP: Partial push out of range rejected.");
+				res.success = false;
+				res.wp_transfered = 0;
+				return true;
+			}
```
vooon: Partial update does not support extension of existing mission?

mavros_msgs/srv/WaypointPush.srv
```diff
@@ -1,7 +1,8 @@
 # Send waypoints to device
-#
+# Start index will define a partial waypoint update
 # Returns success status and transfered count
```
vooon: Better edit that style:
```
# Send waypoints to device
#
#  :start_index: will define a partial waypoint update. Set to 0 for full update.
#
# Returns success status and transferred count
```

mavros/src/plugins/waypoint.cpp
vooon: Oh, no, that would not work.

here that might be defaulted to `false` but in `connection_cb()` need to check:

```
if (!wp_nh.getParam(...)) {
// no override param - detect by FCU
enable = m_uas->is_ardupilotmega();
}
```

vooon:
I think that parameter is fine, it might be used as override.

mavros/src/plugins/waypoint.cpp
JMare: Yes this is better. Then i will remove it from the APM config yaml.

mavros/src/plugins/waypoint.cpp
JMare: No, it doesn't the waypoint count cant change

JMare:
@vooon I will make the changes you suggested

@TSC21 I will fix the merge errors and do the documentation when I can.

Thanks

@JamesStewy can you please make the recommended changes to the pythons scripts

JMare:
@vooon We have applied all your suggestions!

@TSC21 I have rebased against master to fix merge conflict.

One other thought we had while we are modifying this plugin is to make the current waypoint number more accessible. It seems the only way to access the current waypoint is to iterate through the waypoints array from /mavros/mission/waypoints checking is_current. This is quite clumsy.
What do you think of adding the current seq number outside of the waypoint array in the published msg as we did here?

https://github.com/MonashUAS/mavros/blob/e57ceca1a8ac86501f4a1a31877609229814ee6c/mavros_msgs/msg/WaypointList.msg

TSC21:
I agree with that idea. Can you also please also update the Doxygen documentation in the plugin, similar to what is done in `imu.cpp`?

JMare:
@TSC21 Please take a look at the last commit and see if that is what you are looking for.

I have not used doxygen before so not sure if it is too redundant.

mavros/src/plugins/waypoint.cpp
```diff
@@ -232,7 +232,12 @@ class WaypointPlugin : public plugin::PluginBase {
 
 	/* -*- rx handlers -*- */
 
-	// WaypointItem is MISSION_ITEM child
+	/**
+	 * @brief handle MISSION_ITEM mavlink msg
+	 * handles and stores mission items when pulling waypoints
+	 * @param msg       Received Mavlink msg
```
TSC21: Please use tab instead of whitespaces. (use the uncrustify tool)

TSC21:
@JMare it is good (despite the ident style). Can you do the same for the `low-level send functions` and the `mid-level helpers`? Thanks

mavros/src/plugins/waypoint.cpp
JMare: I did use uncrustify but i guess it doesn't fix tabs that aren't at the start of a line.
I've fixed this

mavros/src/plugins/waypoint.cpp
```diff
@@ -513,10 +509,16 @@ class WaypointPlugin : public plugin::PluginBase {
...
 	{
 		lock_guard lock(mutex);
-		if (connected)
+		if (connected) {
 			shedule_pull(BOOTUP_TIME_DT);
-		else
+
+			if (!wp_nh.getParam("enable_partial_push",enable_partial_push)) {
+				enable_partial_push = m_uas->is_ardupilotmega();
+			}
```
vooon: I just checked NodeHandle, and better to use `wp_nh.hasParam()`: http://docs.ros.org/jade/api/roscpp/html/classros_1_1NodeHandle.html#aadd6a7c41de08f54a913e97f26a3af5e

mavros/src/plugins/waypoint.cpp
```diff
@@ -541,6 +545,7 @@ class WaypointPlugin : public plugin::PluginBase {
 		}
 	}
 
+	// @brief Act on first heartbeat from FCU
 	void connection_cb(bool connected) override
```
vooon: Override functions do not need special documentation (exception if it do some not obvious things).

mavros/src/plugins/waypoint.cpp
```diff
@@ -556,6 +561,7 @@ class WaypointPlugin : public plugin::PluginBase {
 		}
 	}
 
+	// @brief Callback for scheduled waypoint pull
```
TSC21: The comment should on the following format to be accepted on Doxygen:
`//! @brief ....`
or
```
  /** @brief ......
   */
```
The first is better for one line comment though.

mavros/src/plugins/waypoint.cpp
```diff
@@ -620,6 +628,8 @@ class WaypointPlugin : public plugin::PluginBase {
 		}
 	}
 
+	// @brief wait until a waypoint pull is complete
```
TSC21: /**
 * @brief wait until a waypoint pull is complete.
 * Pull happens asynchronously, this function blocks until it is done.
 */

mavros/src/plugins/waypoint.cpp
```diff
@@ -628,6 +638,8 @@ class WaypointPlugin : public plugin::PluginBase {
 		       && !is_timedout;
 	}
 
+	// @brief wait until a waypoint push is complete
```
TSC21:
```c++
/**
 * @brief wait until a waypoint pull is complete.
 * Push happens asynchronously. This function blocks until it is done.
 */
```

mavros/src/plugins/waypoint.cpp
```diff
@@ -573,6 +579,7 @@ class WaypointPlugin : public plugin::PluginBase {
 		mission_request_list();
 	}
 
+	// @brief Send ACK back to FCU after pull
```
TSC21: `//! @brief Send ACK back to FCU after pull`

mavros/src/plugins/waypoint.cpp
TSC21: @JMare just like this:
`// Act on first heartbeat from FCU` (no `@brief` and no `//!`, which results in it not being processed on Doxygen)

JMare:
Thanks @TSC21 and @vooon I will implement your suggestions but I am going away this weekend and wont be able to do this for a couple of days.

JMare:
Thanks @TSC21 :)

PR #804, opened 2017-09-02
IMU and attitude: general clean-up

TSC21:
This is a general cleanup of the IMU (and attitude) plugin, and includes the following:
1. Rename `imu_pub` to `imu` so to generalize it more (it pubs and stores, and also has attitude data)
2. Store data in NED frame, which allows it to use whenever we want and in specific cases where we want to track body frame orientation - example: https://github.com/mavlink/mavros/blob/master/mavros_extras/src/plugins/odom.cpp#L100-L103
3. General clean-up on comments to better documentation and Doxygen presentation
4. Cleans the HIL plugin body frame tracking (it shouldn't be required).


mavros/src/plugins/imu.cpp
```diff
@@ -0,0 +1,501 @@
...
+	 * @param[in,out] gyro_ned	Angular velocity/rate in the aircraft NED frame
+	 */
+	void publish_imu_data(uint32_t time_boot_ms, Eigen::Quaterniond &orientation_enu,
+				Eigen::Quaterniond &orientation_ned, Eigen::Vector3d &gyro_enu, Eigen::Vector3d &gyro_ned)
+	{
+		auto imu_ned_msg = boost::make_shared<sensor_msgs::Imu>();
+		auto imu_enu_msg = boost::make_shared<sensor_msgs::Imu>();
+
+		/** Fill message header
+		 */
```
vooon: That should not be a doxygen comment.

mavros/src/plugins/imu.cpp
```diff
@@ -0,0 +1,501 @@
...
+ * millG to m/s**2 coeff
+ */
+static constexpr double MILLIG_TO_MS2 = 9.80665 / 1000.0;
+/**
+ * millBar to Pascal coeff
+ */
+static constexpr double MILLIBAR_TO_PASCAL = 1.0e2;
+/**
+ * Radians to degrees
+ */
```
vooon: Why? Why that large format for single line comment? `//!` is also interpreted by doxygen!

TSC21:
> I can't see what exactly you changed in imu because you also renamed file.

Yeah sorry about that. I basically added the storage of the NED orientation, which is useful when we want to track body frame rotations.

mavros/src/plugins/imu.cpp
TSC21: Yeah you are right, I just wanted to keep it consistent. But makes more sense the shorter format.

mavros/src/plugins/imu.cpp
```diff
@@ -0,0 +1,530 @@
...
+	{
+		if (has_att_quat)
+			return;
+
+		/** Orientation on the NED-aicraft frame:
+		 *  @snippet src/plugins/imu.cpp ned_aircraft_orient1
+		 */
+		// [ned_aircraft_orient1]
+		auto ned_aircraft_orientation = ftf::quaternion_from_rpy(att.roll, att.pitch, att.yaw);
+		// [ned_aircraft_orient1]
```
vooon: What is that? Also i don't think that we need function body comments in doxygen.

mavros/src/plugins/imu.cpp
```diff
@@ -0,0 +1,530 @@
...
+		// [pub_enu]
+	}
+
+	/**
+	 * @brief fill and publish IMU data_raw message; store linear acceleration for IMU data
+	 * @param[in] header		Message frame_id and timestamp
+	 * @param[in,out] gyro		Orientation in the base_link ENU frame
+	 * @param[in,out] accel_enu	Linear acceleration in the base_link ENU frame
+	 * @param[in,out] accel_ned	Linear acceleration in the aircraft NED frame
+	 */
```
vooon: I do not see any assigns to any parameters. Why `in,out`?

mavros/src/plugins/imu.cpp
```diff
@@ -0,0 +1,530 @@
...
+		ftf::EigenMapCovariance3d c(cov.data());
+
+		c.setZero();
+		if (stdev) {
+		    double sr = stdev * stdev;
+		    c.diagonal() << sr, sr, sr;
+		}
+		else {
+		   c(0,0) = -1.0;
+		}
```
vooon: Spaces instead of tabs. Check uncrustify.

mavros/src/plugins/imu.cpp
```diff
@@ -0,0 +1,530 @@
...
+		magn_pub.publish(magn_msg);
+	}
+
+	/* -*- message handlers -*- */
+
+	/**
+	 * @brief Handle ATTITUDE MAVlink message.
+	 * Message specification: http://mavlink.org/messages/common/#ATTITUDE
+	 * @param[in] msg	Received Mavlink msg
+	 * @param[in,out] att	ATTITUDE msg
```
vooon: Why `in,out` again? Also in most cases there no need to specify `@param[in]`. In most places `@param` is enough (assume `in` by default).
And really what is needed is `@param[out]`.

mavros/src/plugins/imu.cpp
```diff
@@ -0,0 +1,530 @@
...
+		atmp_msg->header = header;
+		atmp_msg->fluid_pressure = press.press_abs * 100.0;
+		press_pub.publish(atmp_msg);
+	}
+
+	/**
+	 * @brief Checks for connection and overrides variable values
+	 * @param[in] connected		Is connected?
+	 */
+	void connection_cb(bool connected) override
```
vooon: That might be documented once in base class.

mavros/src/plugins/imu.cpp
TSC21: `[in,out]` is not explicit on the Doxygen docs. But I assume it is a parameter that is fed on the function, it is modified, and presents itself as an output (unless the `out` represents the output of `return` - is that it?).

mavros/src/plugins/imu.cpp
TSC21: sure

mavros/src/plugins/imu.cpp
TSC21: what do you mean?

mavros/src/plugins/imu.cpp
TSC21: This is how you include a code snippet on Doxygen docs. Without it, the comments above the code lines would appear loose on the documentation (just the lines, and not the code associated to the comment). That way you guarantee that the code you are referring to on the comment is also documented on the output of Doxygen.

mavros/src/plugins/imu.cpp
TSC21: Does `@param[out]` represent the output of the return function? If that is the case, then no `@param[out]` should be inserted, as these functions return `void`.

mavros/src/plugins/imu.cpp
TSC21: Oh ok this is an override. So should I remove the associated doc tags?

mavros/src/plugins/imu.cpp
vooon: No, return has it's own tag - `@returns`. `@param[out]` is when you modify argument.

Example:
```c++
/**
 * Copy memory area.
 * @note Memory areas should not overlap.
 * @param[out] *dest destination pointer.
 * @param *src source pointer
 * @param n area size in bytes
 * @returns *dest pointer
 */
void *memcpy(void *dest, const void *src, size_t n);
```

mavros/src/plugins/imu.cpp
vooon: https://github.com/mavlink/mavros/blob/master/mavros/include/mavros/mavros_plugin.h you may full-fill comment in base class.
All overrides should inherit that explanation. Else we will copy-paste same explanations in each plugin.

mavros/src/plugins/imu.cpp
TSC21: @vooon thanks for clarifying that

Issue #807, opened 2017-09-11
Unable to switch to AUTO.MISSION. But QGC does it. 


shakthi-prashanth-m:
This is only bug and feature tracker, please use it
to report bugs or request features.

----

**Issue details**

I am unable to enter to MISSION mode.

```
 // Set to Mission mode
  ROS_INFO("Now, Setting to Mission mode...");
  mavros_msgs::SetMode mission_set_mode;
  mission_set_mode.request.custom_mode = "AUTO.MISSION";
  ros::ServiceClient set_mode_client =
      nh.serviceClient<mavros_msgs::SetMode>("mavros/set_mode");

  if (set_mode_client.call(mission_set_mode) &&
      mission_set_mode.response.success) {
    ROS_INFO("In Mission mode now");  /* **************** => I got this printed **********/
    // ARM
    mavros_msgs::CommandBool arm_cmd;
    arm_cmd.request.value = true;
    ros::ServiceClient arming_client =
        nh.serviceClient<mavros_msgs::CommandBool>("mavros/cmd/arming");

    // Lets not overload FCU
    sleep(5.0);

    if (!current_state.armed) {
      if (arming_client.call(arm_cmd) && arm_cmd.response.success) {
        ROS_INFO("Vehicle armed");
      }
    }
  } else {
    ROS_ERROR("Failed to set Mission mode!");
    ros::shutdown();
}
```
I am sending Waypoint missions to PX4 & setting to Mission mode. Though `mission_set_mode.response.succes` is true, from the state callback (`/mavros/state`) I came to know that mode switch didn't happen. How could this happen ? If setting to Mission mode fails, it should be reported in `mission_set_mode.response.succes` as `false` isn't it ?
But the same operation works in QGC. What QGC does differently ? Am I missing something?

**MAVROS version and platform**

Mavros: 0.19.0
ROS: Kinetic
Ubuntu: 16.04

**Autopilot type and version**

PX4
Version: 4.4.0

Kindly help to resolve this.
Thanks in advance.

--Shakthi

TSC21:
> I am sending Waypoint missions to PX4 & setting to Mission mode. Though `mission_set_mode.response.succes` is true, from the state callback (`/mavros/state`) I came to know that mode switch didn't happen. How could this happen ? If setting to Mission mode fails, it should be reported in `mission_set_mode.response.succes` as `false` isn't it ?

[...]

PR #810, opened 2017-09-13
waypoint: Publish current waypoint seq

JMare:
This PR makes the current waypoint sequence number more accessible.

Currently finding the sequence number requires iterating through `/mavros/mission/waypoints` and checking `is_current`.

This PR adds seq field to the waypointlist message outside of the waypoint array.

As discussed in https://github.com/mavlink/mavros/pull/798#issuecomment-327745029.

PR #811, opened 2017-09-16
Set flight mode: fix and enhancements

TSC21:
This PR:
1. Fixes `mavsys mode` call - `base_mode` should be set to 1 (`MAV_MODE_FLAG_CUSTOM_MODE_ENABLED`) in order to set custom modes (`AUTO` submodes for example in PX4 Pro) - confirmed the requirement for both PX4 and APM;
2. Also on `mavsys` - it removes the exclusivity of using `-c` or `-b` - it is verified that some custom modes require the `base_mode` to be set to a number different that 0 (or 1);
3. `SetMode.srv`: adds flag references (useful for setting the `base_mode` through the command line)
4. Updates enums on `enum_to_string`

What is left to do for another PR: Add as alternative the command call to `MAV_CMD_DO_SET_MODE`, as the `SET_MODE` Mavlink message is [deprecated](http://mavlink.org/messages/common#SET_MODE).

This was tested by me and by @shakthi-prashanth-m.


TSC21:
Fixes #807.

TSC21:
Just to leave this clear: to set the flight mode to `AUTO_MISSION` in PX4 Pro, you should first disable the datalink failsafe by setting `NAV_DLL_ACT` parameter to 0. Already updated in http://wiki.ros.org/mavros/CustomModes

vaasudEva:
Tested on Gazebo SITL, Working.  Thanks @tsc21.

mavros/scripts/mavsys
```diff
@@ -20,7 +20,11 @@ from mavros_msgs.srv import SetMode, StreamRate, StreamRateRequest
 
 
 def do_mode(args):
-    base_mode = 0
+    # Setting to MAV_MODE_FLAG_CUSTOM_MODE_ENABLED is mandatory, so to change to
+    # a custom mode/submode.
+    # NOTE: In order to change to AUTO.MISSION on PX4 Pro, you need to disable
+    # datalink failsafe, which can be done by setting NAV_DLL_ACT to 0.
+    base_mode = 1 # 1 = MAV_MODE_FLAG_CUSTOM_MODE_ENABLED
```
vooon: Please read the code *before* edit!

[sys_status.cpp 931-945](https://github.com/mavlink/mavros/blob/master/mavros/src/plugins/sys_status.cpp#L931-L945)
```cpp
		if (req.custom_mode != "") {
			if (!m_uas->cmode_from_str(req.custom_mode, custom_mode)) {
				res.mode_sent = false;
				return true;
			}

			/**
			 * @note That call may trigger unexpected arming change because
			 *       base_mode arming flag state based on previous HEARTBEAT
			 *       message value.
			 */
			base_mode |= (m_uas->get_armed()) ? enum_value(MAV_MODE_FLAG::SAFETY_ARMED) : 0;
			base_mode |= (m_uas->get_hil_state()) ? enum_value(MAV_MODE_FLAG::HIL_ENABLED) : 0;
			base_mode |= enum_value(MAV_MODE_FLAG::CUSTOM_MODE_ENABLED);
		}
```

mavros/scripts/mavsys
```diff
@@ -88,9 +92,8 @@ def main():
 
     mode_args = subarg.add_parser('mode', help="Set mode")
     mode_args.set_defaults(func=do_mode)
-    mode_group = mode_args.add_mutually_exclusive_group(required=True)
-    mode_group.add_argument('-b', '--base-mode', type=int, help="Base mode code")
-    mode_group.add_argument('-c', '--custom-mode', type=str, help="Custom mode string (same as in ~/state topic)")
+    mode_args.add_argument('-b', '--base-mode', type=int, help="Base mode code")
+    mode_args.add_argument('-c', '--custom-mode', type=str, help="Custom mode string (same as in ~/state topic)")
```
vooon: Document where you need to specify base mode with custom mode.

mavros/src/lib/enum_to_string.cpp
```diff
@@ -158,7 +158,6 @@ static const std::array<const std::string, 29> mav_type_strings{{
 /* 25 */ "VTOL reserved 5",               // VTOL reserved 5
 /* 26 */ "Onboard gimbal",                // Onboard gimbal
 /* 27 */ "Onboard ADSB peripheral",       // Onboard ADSB peripheral
-/* 28 */ "Dodecarotor",                   // Dodecarotor
 }};
```
vooon: `pip install -U pymavlink`

TSC21: I have the latest pymvalink and this was thr resulting enun

mavros/scripts/mavsys
TSC21: https://github.com/mavlink/mavros/pull/810#issuecomment-329759400

mavros/scripts/mavsys
TSC21: So you have to explain me why without setting the base_mode to 1 in mavsys it does not change to flight submodes. From my evaluation, what you have above is not doing what is supposed.

shakthi-prashanth-m: The original issue that I faced was https://github.com/mavlink/mavros/issues/807. I was unable to switch to MISSION mode after uploading missions. But whereas I was able to do so in QGC.
So, as me & @TSC21  went on experimenting  the necessary changes, we had set `base_mode=1` & unset of `NAV_DLL_ACT`. In reality only later change is required, not `base_mode=1`.  
Switching to LOITER mode too wasn't happening. This was also resolved by doing unset of `NAV_RCL_ACT`. 
So, the conclusion is the mode switch to AUTO.MISSION & AUTO.LOITER will happen by setting parameters `NAV_DLL_ACT` & `NAV_RCL_ACT` to `0`. It is verified by using both C++ and `mavsys` commands.
```
$ rosrun mavros mavsys mode -c AUTO.MISSION
Mode changed
```
Thanks @TSC21 for suggesting param set commands which resolved my issue.
--Shakthi

mavros/scripts/mavsys
vooon: Base mode 3 is weird. CUSTOM_MODE_ENABLED + TEST_ENABLED.
More likely to see CUSTOM_MODE + SAFETY_ARMING + AUTO_ENABLED.
But last time i checked APM it do not rely on base_mode at all (except custom mode flag).

If PX4 now require AUTO_ENABLED just add quirk to handler. Nothing more.

vooon: And `GCS_Common.cpp` still operates same way:
```cpp
1289 /*
1290   handle a SET_MODE MAVLink message
1291  */
1292 void GCS_MAVLINK::handle_set_mode(mavlink_message_t* msg)
1293 {
1294     uint8_t result = MAV_RESULT_FAILED;
1295     mavlink_set_mode_t packet;
1296     mavlink_msg_set_mode_decode(msg, &packet);
1297 
1298     // only accept custom modes because there is no easy mapping from Mavlink flight modes to AC flight modes
1299     if (packet.base_mode & MAV_MODE_FLAG_CUSTOM_MODE_ENABLED) {
1300         if (set_mode(packet.custom_mode)) {
1301             result = MAV_RESULT_ACCEPTED;
1302         }
1303     } else if (packet.base_mode == MAV_MODE_FLAG_DECODE_POSITION_SAFETY) {
1304         // set the safety switch position. Must be in a command by itself
1305         if (packet.custom_mode == 0) {
1306             // turn safety off (pwm outputs flow to the motors)
1307             hal.rcout->force_safety_off();
1308             result = MAV_RESULT_ACCEPTED;
1309         } else if (packet.custom_mode == 1) {
1310             // turn safety on (no pwm outputs to the motors)
1311             if (hal.rcout->force_safety_on()) {
1312                 result = MAV_RESULT_ACCEPTED;
1313             }
1314         }
1315     }
1316 
1317     // send ACK or NAK
1318     mavlink_msg_command_ack_send_buf(msg, chan, MAVLINK_MSG_ID_SET_MODE, result);
1319 }
```

TSC21: Ok so if one wants to set different flags, where should it be set? In https://github.com/mavlink/mavros/blob/master/mavros/src/plugins/sys_status.cpp#L931-L945 ?

vooon: https://www.youtube.com/watch?v=8zVYru974iQ

(Something went wrong with gtk-recordmydesktop, video goes too fast...)

TSC21:
It says the video is not available @vooon

vooon:
I done test with px4 sitl - no changes needed (at least when qgc on).
https://youtu.be/99QWMSk58Dc

vooon:
Hmm without QGC it don't switch to auto.mission. So i'm suppose that this may be related to arming auth.

vooon:
Yeah, that is all about arming auth
```
[DEBUG] [1505736362.506400911]: udp0: recv: v1.0   OK Message-Id: 76 [33 bytes] IDs: 1.1 Seq: 12
```

Solution: place intel's mavlink-router between FCU and mavros.

TSC21:
Why did you close this PR? The solution is not that! The solution, as stated above, is to close disable the failsafes.

TSC21:
> Hmm without QGC it don't switch to auto.mission. So i'm suppose that this may be related to arming auth.

Wrong - if you read @shakthi-prashanth-m comment you will see it is because of the failsafe.

> Solution: place intel's mavlink-router between FCU and mavros.

No it is not the solution.

shakthi-prashanth-m:
@voon ,  I think mavlink-router won't come in the picture as I used px4 SITL (gazebo) in my case.  Its only after parameter datalink failsafe the issue was resolved.

PR #814, opened 2017-09-18
mavsys: mode: add solutions for setting AUTO.MISSION and AUTO.LOITER modes

TSC21:
Also the enums are updated - can push this to another PR if needed.

### DIFF
--- a/mavros/scripts/mavsys
+++ b/mavros/scripts/mavsys
@@ -20,7 +20,15 @@ from mavros_msgs.srv import SetMode, StreamRate, StreamRateRequest
 
 
 def do_mode(args):
+<<<<<<< HEAD
     base_mode = 0
+=======
+    # Setting to MAV_MODE_FLAG_CUSTOM_MODE_ENABLED is mandatory, so to change to
+    # a custom mode/submode.
+    # NOTE: In order to change to AUTO.MISSION on PX4 Pro, you need to disable
+    # datalink failsafe, which can be done by setting NAV_DLL_ACT to 0.
+    base_mode = 1 # 1 = MAV_MODE_FLAG_CUSTOM_MODE_ENABLED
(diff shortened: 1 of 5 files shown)


mavros/scripts/mavsys
```diff
@@ -20,7 +20,15 @@ from mavros_msgs.srv import SetMode, StreamRate, StreamRateRequest
 
 
 def do_mode(args):
+<<<<<<< HEAD
     base_mode = 0
+=======
```
khancyr: looks like rebase error ^^

TSC21: That's fixed @khancyr. thanks

TSC21:
Looks like we don't need to unset the datalink failsafe. For PX4, the heartbeat type published by `sys_status`, for now, needs to be of type `MAV_TYPE_GCS` so to be accepted and published on the `telemetry_status` uORB msg and that way set the datalink stream. IOT to it be accepted also with `MAV_TYPE_ONBOARD_CONTROLLER`, I will purpose a patch to the Firmware.
And that is not all: https://github.com/PX4/Firmware/blob/master/src/modules/mavlink/mavlink_receiver.cpp#L1718 - the 1st condition is that `msg->sysid != mavlink_system.sysid`, so we cannot have `system_id == tgt_system_id`. And that's why I'm adding that option by default to the launcher filers.

TSC21:
@shakthi-prashanth-m for your interest.

TSC21:
@khancyr in Ardupilot is there any special case like this? Do you treat the hearbeat differently according to the `MAV_TYPE`?

vooon:
Mavros sysid may be set via url, addition to launch is not required.
But for me that check looks weird, because OBC is part of same Unmanned System as FCU.

So FCU should understand when (hb.sysid == my.sysid && hb.type == OBC) - we have link with our OBC.

PR #816, opened 2017-09-19
Use tf2 for odom plugin and set reasoable defaults for local pos cov.

jgoppert:
These are my outstanding changes to get visual odometry working well for my setup. This uses tf2 so it actually pays attention to the frame of the odometry message as specified in the message, instead of assuming it is ENU.

### DIFF
--- a/mavros_extras/src/plugins/odom.cpp
+++ b/mavros_extras/src/plugins/odom.cpp
@@ -16,6 +16,7 @@
 
 #include <mavros/mavros_plugin.h>
 #include <eigen_conversions/eigen_msg.h>
+#include <tf2_eigen/tf2_eigen.h>
 
 #include <nav_msgs/Odometry.h>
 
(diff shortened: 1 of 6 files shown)


TSC21:
@jgoppert can you please take a look on what me and @vooon proposed on https://github.com/mavlink/mavros/pull/756?

jgoppert:
This requires this frame definition to be added, as mentioned in the ros frame REP.

http://www.ros.org/reps/rep-0103.html

For outdoor systems where it is desireable to work under the north east down [6] (NED) convention, define an appropriately transformed secondary frame with the "_ned" suffix:

mavros_extras/src/plugins/odom.cpp
```diff
@@ -66,87 +67,97 @@ class OdometryPlugin : public plugin::PluginBase {
...
-		ftf::EigenMapConstCovariance6d cov_vel_map(cov_vel.data());
-		ROS_DEBUG_STREAM_NAMED("odom", "Odometry: velocity covariance matrix: " << std::endl << cov_vel_map);
+		// lookup transforms
+		Eigen::Affine3d tfChild2NED;
+		Eigen::Affine3d tfParent2NED;
+		try {
+			tfChild2NED = tf2::transformToEigen(m_uas->tf2_buffer.lookupTransform(
+				odom->child_frame_id, "local_origin_ned", ros::Time(0)));
+			tfParent2NED = tf2::transformToEigen(m_uas->tf2_buffer.lookupTransform(
+				odom->header.frame_id, "local_origin_ned", ros::Time(0)));
```
TSC21: If we are to add this, we should also add the static transform publisher for it right?

jgoppert: Yeah, sorry, just added that.

TSC21:
Also, the way the body rotations are handled by default on the plugin does not work on your setup (meaning, without `tf2`)?

mavros_extras/src/plugins/odom.cpp
```diff
@@ -66,87 +67,97 @@ class OdometryPlugin : public plugin::PluginBase {
...
+		poseCovNED = RPose*poseCovParent*RPose.transpose();
 		lpos.estimator_type = utils::enum_value(estimator_type);
-		ftf::covariance9d_urt_to_mavlink(cov_full_tf, lpos.covariance);
+		for (int i=0; i<6; i++) for (int j=0; j<6; j++) {
+			lpos.covariance[i*6 + j] = poseCovNED(i, j);
+		}
 
 		// [[[cog:
-		// for a, b in (('', 'pos_ned'), ('v', 'lin_vel_ned'), ('a', 'zerov3f')):
+		// for a, b in (('', 'posNED'), ('v', 'linVelNED')):
```
TSC21: is there a reason you changed from `pos_ned` to `posNED`? Throughout MAVROS we use the `name_name` convention instead of `nameName`. Would be good to keep that consistency here if you don't mind.

jgoppert:
> Also, the way the body rotations are handled by default on the plugin does not work on your setup (meaning, without tf2)?

tf2 is already used in mavros, I'm actually using the tf2 listener/ broadcaster from m_uas. So you can't run mavros without tf2. If you mean from an end-user perspective, they were specifying the odometry message before but maybe not paying attention to what frame_id/child_frame_id they specified. Now we will start listening to these.

TSC21:
> tf2 is already used in mavros, I'm actually using the tf2 listener/ broadcaster from m_uas. So you can't run mavros without tf2. If you mean from an end-user perspective, they were specifying the odometry message before but maybe not paying attention to what frame_id/child_frame_id they specified. Now we will start listening to these.

I agree with you and I believe we need to start somewhere and then expand it in a general fashion so the API is changed the correct way. But what I wanted to understand is if it really works or not without `tf2`, meaning the current implementation? I'm not saying to keep it, I just want to know if it works for your setup. I remember that time that for me, on sim, it was working.

jgoppert:
@TSC21 Currently I redirect this odometry message from another package that does not publish it in the local_origin  (ENU) frame, so it breaks. I would have to subscribe, fix the odometry message to be the mavlink assumed frame and republish.

TSC21:
> Currently I redirect this odometry message from another package that does not publish it in the local_origin (ENU) frame,

I assume that with a static publisher you could change the frame.

jgoppert:
No, the ros node also publishes to tf2, so you can't override it. Granted you could do some trickery, but I think it is much better to fix mavros. I setup the static transform publisher, let me know what you think.

jgoppert:
I just flew this in SITL and it looks good now. I rebased it on master.

mavros_extras/src/plugins/odom.cpp
```diff
@@ -66,88 +67,92 @@ class OdometryPlugin : public plugin::PluginBase {
...
-		lpos.vx = lin_vel_ned.x();
-		lpos.vy = lin_vel_ned.y();
-		lpos.vz = lin_vel_ned.z();
-		lpos.ax = zerov3f.x();
-		lpos.ay = zerov3f.y();
-		lpos.az = zerov3f.z();
-		// [[[end]]] (checksum: 4620cfc77f3e5d768d7c23da5f56ddc6)
+		for (int i=0; i<6; i++) for (int j=0; j<6; j++) {
+			lpos.covariance[i*6 + j] = pose_cov_ned(i, j);
+		}
```
vooon: According to [mavlink docs][1] covariance is 9x9 not 6x6, and only upper triangular part.

[1]: http://mavlink.org/messages/common#LOCAL_POSITION_NED_COV

mavros_extras/src/plugins/odom.cpp
```diff
@@ -66,88 +67,92 @@ class OdometryPlugin : public plugin::PluginBase {
...
+
+		lpos.x = pos_NED.x();
+		lpos.y = pos_NED.y();
+		lpos.z = pos_NED.z();
+		lpos.vx = lin_vel_NED.x();
+		lpos.vy = lin_vel_NED.y();
+		lpos.vz = lin_vel_NED.z();
+		lpos.ax = 0;
+		lpos.ay = 0;
+		lpos.az = 0;
```
vooon: Plase keep generators. Usage is simple:
```
pip install --user cogapp
cog.py -cr <file>
```

mavros_extras/src/plugins/odom.cpp
```diff
@@ -156,17 +161,15 @@ class OdometryPlugin : public plugin::PluginBase {
 		mavlink::common::msg::ATTITUDE_QUATERNION_COV att {};
 
 		att.time_usec = stamp_usec;
-		ftf::quaternion_to_mavlink(q_ned, att.q);
-		ftf::covariance_to_mavlink(cov_vel, att.covariance);
+		ftf::quaternion_to_mavlink(q_parent2ned, att.q);
+		att_cov_ned = tf_child2ned.linear()*twist_cov_child.block<3, 3>(0, 0)*tf_child2ned.linear().transpose();
+		for (int i=0; i<3; i++) for (int j=0; j<3; j++) {
+			att.covariance[i*3 + j] = att_cov_ned(i, j);
+		}
```
vooon: Maybe `Matrix::array()` is what you need to pass to `ftf::covariance_to_mavlink()`?

mavros_extras/src/plugins/odom.cpp
```diff
@@ -66,88 +67,92 @@ class OdometryPlugin : public plugin::PluginBase {
...
+		Eigen::Vector3d ang_vel_child(
+			odom->twist.twist.angular.x,
+			odom->twist.twist.angular.y,
+			odom->twist.twist.angular.z);
+
+		Eigen::Quaterniond q_parent2child(
+			odom->pose.pose.orientation.w,
+			odom->pose.pose.orientation.x,
+			odom->pose.pose.orientation.y,
+			odom->pose.pose.orientation.z);
```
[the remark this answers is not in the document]
vooon: In master, `ftf::to_eigen()` for Vector3, Point and Quaternion.

TSC21:
Replaced by #824

TSC21:
@jgoppert can we delete the `odom` branch from mavros repo, since this is already on upstream master?

Issue #793, opened 2017-08-29
setpoint_attitude subscriber missing


andresR8:
### Issue details

setpoint_attitude is not working in the latest release. 

### MAVROS version and platform

Mavros: 0.20.1-0xenial-20170828-082737-0800
ROS: Kinetic
Ubuntu: 16.04

### Autopilot type and version

[  ] ArduPilot
[ X] PX4

Version: 1.6.5dev

### Node logs
Apparently mavros is not subscribing to the topic
```
rostopic info /mavros/setpoint_attitude/attitude
Type: geometry_msgs/PoseStamped

Publishers: 
 * /rover_teleop (http://kiwi:35311/)

Subscribers: None

```

### Diagnostics

```
header: 
  seq: 411
  stamp: 
    secs: 1503984484
    nsecs: 805333322
  frame_id: ''
status: 
  - 
    level: 0
    name: mavros: FCU connection
    message: connected
    hardware_id: /dev/ttyACM0:57600
    values: 
      - 
        key: Received packets:
        value: 32207
      - 
        key: Dropped packets:
        value: 0
      - 
        key: Buffer overruns:
        value: 0
      - 
        key: Parse errors:
        value: 0
      - 
        key: Rx sequence number:
        value: 182
      - 
        key: Tx sequence number:
        value: 207
      - 
        key: Rx total bytes:
        value: 6863362
      - 
        key: Tx total bytes:
        value: 693265
      - 
        key: Rx speed:
        value: 19258.000000
      - 
        key: Tx speed:
        value: 1927.000000
  - 
    level: 2
    name: mavros: GPS
    message: No satellites
    hardware_id: /dev/ttyACM0:57600
    values: 
      - 
        key: Satellites visible
        value: 0
      - 
        key: Fix type
        value: 0
      - 
        key: EPH (m)
        value: Unknown
      - 
        key: EPV (m)
        value: Unknown
  - 
    level: 0
    name: mavros: Heartbeat
    message: Normal
    hardware_id: /dev/ttyACM0:57600
    values: 
      - 
        key: Heartbeats since startup
        value: 533
      - 
        key: Frequency (Hz)
        value: 1.039960
      - 
        key: Vehicle type
        value: Ground rover
      - 
        key: Autopilot type
        value: PX4 Autopilot
      - 
        key: Mode
        value: OFFBOARD
      - 
        key: System status
        value: Standby
  - 
    level: 0
    name: mavros: System
    message: Normal
[...]

PR #821, opened 2017-09-24
fix: attitude callback trigger

andresR8:
After the correct subscription of the topics (#793). When another node publish to the topics (target_attitude & thrust)  the corresponding callback function  `&SetpointAttitudePlugin::attitude_pose_cb` is not triggered.

Apparently the sync objects should be defined as a private property of the plugin and not only locally in the init method.

### DIFF
--- a/mavros/src/plugins/setpoint_attitude.cpp
+++ b/mavros/src/plugins/setpoint_attitude.cpp
@@ -27,11 +27,11 @@
 #include <geometry_msgs/TwistStamped.h>
 #include <mavros_msgs/Thrust.h>
 
+#include <boost/thread.hpp>
+
 namespace mavros {
 namespace std_plugins {
 
-using SyncPoseThrust = message_filters::sync_policies::ApproximateTime<geometry_msgs::PoseStamped, mavros_msgs::Thrust>;
-using SyncTwistThrust = message_filters::sync_policies::ApproximateTime<geometry_msgs::TwistStamped, mavros_msgs::Thrust>;
 
 /**
  * @brief Setpoint attitude plugin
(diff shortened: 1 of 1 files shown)


mavros/src/plugins/setpoint_attitude.cpp
```diff
@@ -107,6 +107,14 @@ class SetpointAttitudePlugin : public plugin::PluginBase,
 	message_filters::Subscriber<geometry_msgs::PoseStamped> pose_sub;
 	message_filters::Subscriber<geometry_msgs::TwistStamped> twist_sub;
 
+	typedef  message_filters::sync_policies::ApproximateTime<geometry_msgs::PoseStamped, mavros_msgs::Thrust> SyncPoseThrustPolicy;
```
TSC21: Why not `using` instead of `typedef`?

mavros/src/plugins/setpoint_attitude.cpp
```diff
@@ -107,6 +107,14 @@ class SetpointAttitudePlugin : public plugin::PluginBase,
...
+	typedef  message_filters::Synchronizer<SyncPoseThrustPolicy> SyncPoseThrust;
+	boost::shared_ptr<SyncPoseThrust> sync_thrust;
+
+	typedef  message_filters::sync_policies::ApproximateTime<geometry_msgs::TwistStamped, mavros_msgs::Thrust> SyncTwistThrustPolicy;
+	typedef  message_filters::Synchronizer<SyncTwistThrustPolicy> SyncTwistThrust;
+	boost::shared_ptr<SyncTwistThrust> sync_twist;
```
TSC21: Organize by types first and then by subject.

mavros/src/plugins/setpoint_attitude.cpp
```diff
@@ -27,11 +27,11 @@
 #include <geometry_msgs/TwistStamped.h>
 #include <mavros_msgs/Thrust.h>
 
+#include <boost/thread.hpp>
+
 namespace mavros {
 namespace std_plugins {
 
-using SyncPoseThrust = message_filters::sync_policies::ApproximateTime<geometry_msgs::PoseStamped, mavros_msgs::Thrust>;
-using SyncTwistThrust = message_filters::sync_policies::ApproximateTime<geometry_msgs::TwistStamped, mavros_msgs::Thrust>;
```
vooon: Useless change. `using` is fine.

mavros/src/plugins/setpoint_attitude.cpp
```diff
@@ -27,11 +27,11 @@
 #include <geometry_msgs/TwistStamped.h>
 #include <mavros_msgs/Thrust.h>
 
+#include <boost/thread.hpp>
+
```
vooon: Not needed change.

mavros/src/plugins/setpoint_attitude.cpp
vooon: Just add `Policy` suffix, and then add
```cpp
using SyncPoseThrust = message_filters::Synchronizer<SyncPoseThrustPplicy>;
using SyncTwistThrust = ...;
```

mavros/src/plugins/setpoint_attitude.cpp
```diff
@@ -107,6 +107,14 @@ class SetpointAttitudePlugin : public plugin::PluginBase,
...
 	message_filters::Subscriber<geometry_msgs::TwistStamped> twist_sub;
 
+	typedef  message_filters::sync_policies::ApproximateTime<geometry_msgs::PoseStamped, mavros_msgs::Thrust> SyncPoseThrustPolicy;
+	typedef  message_filters::Synchronizer<SyncPoseThrustPolicy> SyncPoseThrust;
+	boost::shared_ptr<SyncPoseThrust> sync_thrust;
+
+	typedef  message_filters::sync_policies::ApproximateTime<geometry_msgs::TwistStamped, mavros_msgs::Thrust> SyncTwistThrustPolicy;
+	typedef  message_filters::Synchronizer<SyncTwistThrustPolicy> SyncTwistThrust;
+	boost::shared_ptr<SyncTwistThrust> sync_twist;
+
```
vooon: Here you need only this two lines:
```cpp
std::unique_ptr<SyncPoseThrust> sync_pose;
std::unique_ptr<SyncTwistThrust> sync_twist;
```

vooon:
@andresR8 thank you. Patch applied manually.

andresR8:
@vooon I've tested your commit (mavros 0.21.2) and I cannot receive the corresponding callback `attitude_pose_cb`

PR #823, opened 2017-09-24
Odom general fixes

TSC21:


jgoppert:
I think you need to keep the RPose part. This is the part that actually rotates the covariance into the frame stated by the user in the message. Also the odometry message doesn't have accel, so we can keep it to 6d.

mavros/launch/apm_config.yaml
```diff
@@ -193,6 +193,8 @@ vibration:
 # None
 
 odometry:
+  frame_tf:
+    desired_frame: "ned"
```
jgoppert: why not just make this local_origin_ned

TSC21: it is. The idea is in the future handle other frame rotations (with offset for example). This is in a more general fashion

mavros_extras/src/plugins/odom.cpp
```diff
@@ -63,96 +66,103 @@ class OdometryPlugin : public plugin::PluginBase {
...
+		Eigen::Vector3d ang_vel_ned(tf_child2ned.linear() * ftf::to_eigen(odom->twist.twist.angular));
+		Eigen::Quaterniond q_parent2ned(tf_parent2ned.linear() * ftf::to_eigen(odom->pose.pose.orientation));
+
+		/** Apply covariance transforms */
+		/** Pose+Accel 9-D Covariance matrix
+		 *  WRT world frame (half upper right triangular)
+		 */
+		Eigen::Matrix<double, 9, 9> r_9d;
+		r_9d.block<3, 3>(0, 0) = r_9d.block<3, 3>(3, 3) = tf_parent2ned.linear();
+		r_9d.block<3, 3>(6, 6) = Eigen::Matrix3d::Identity();
```
jgoppert: This is only rotating the rotation, but you need to also rotate the velocity as stated in the message, the velocity is given in the child frame.

Also, I think we can use the 9x6 matrix that I setup since we don't use accels.

TSC21:
> I think you need to keep the RPose part. This is the part that actually rotates the covariance into the frame stated by the user in the message. Also the odometry message doesn't have accel, so we can keep it to 6d.

If you take a look on how the code is, you will see that every aspect of your code was kept but using Eigen::Map. It is also true that there is no accel on it, that's why when the rotation is applied, the rotation matrix has the identity on the <6,6> block. At an higher level, I didn't change anything, I just tighten things.

mavros_extras/src/plugins/odom.cpp
```diff
@@ -63,96 +66,103 @@ class OdometryPlugin : public plugin::PluginBase {
...
 		Eigen::Affine3d tf_parent2ned;
 		try {
+			// transform lookup WRT world frame
 			tf_child2ned = tf2::transformToEigen(m_uas->tf2_buffer.lookupTransform(
-				odom->child_frame_id, "local_origin_ned", ros::Time(0)));
+							odom->child_frame_id, "local_origin" + desired_frame, ros::Time(0)));
+			// transform lookup WRT body frame
 			tf_parent2ned = tf2::transformToEigen(m_uas->tf2_buffer.lookupTransform(
-				odom->header.frame_id, "local_origin_ned", ros::Time(0)));
+							odom->header.frame_id, "local_origin" + desired_frame, ros::Time(0)));
```
jgoppert: this is currently resolved to "local_originned", so I think it would be better to just leave the full frame name in the config file and not assume the user wants local_origin something.

mavros_extras/src/plugins/odom.cpp
TSC21: > This is only rotating the rotation, but you need to also rotate the velocity as stated in the message, the velocity is given in the child frame.

Check the code bellow: https://github.com/mavlink/mavros/pull/823/files#diff-750c83d1bf274515183c544f142bac77R132

mavros_extras/src/plugins/odom.cpp
TSC21: oh sorry it's a typo.

This assumption will be a way of generalize it for all the API.

TSC21:
@jgoppert can we continue the review on #823? There the code is rebased.

PR #824, opened 2017-09-24
Odom fix

TSC21:
Rebased version of #816. Added some fixes.


TSC21:
Awesome @jgoppert! @vooon anything else you want to add? As said, the idea would be generalize this on the API so to replace the static transforms we have on frame_tf. But for now we can have odom plugin as the proof of concept.

vooon:
LGTM, except `inline` on not inlined method.

PR #826, opened 2017-09-27
Adding an optional service call on MISSION_ITEM_REACHED

gui2dev:
modified:   mavros/src/plugins/waypoint.cpp
	modified:   mavros_msgs/CMakeLists.txt
	new file:   mavros_msgs/srv/WaypointReached.srv

I added the option to call a service when a waypoint is reached.
This helped me to trigger some specific device and save some log.

Basic implementation is to advertise a WaypointReached service and specify it's topic to mavros by setting `/mavros/mission/trig_after_reached` to `true` and specify `/mavros/mission/trig_srv_name`


vooon:
Why not to use regular topic? And without any config, you always can rewrite topic path in launch.

gui2dev:
I totally agree, it did not cross my mind that you could remap that kind of topic.

TSC21:
@gui2dev is this PR still applicable? Or do you plan to do it differently?

gui2dev:
Well yes, still open. That's my first PR, so I'm not sure what I should do.
For now I have to remove `/mavros/mission/trig_srv_name` parameter and give the service a regular name but I'm not sure where I should do the modifications.

gui2dev:
I guess that it has. I guess I just have to remove the

gui2dev:
@vooon, I just updated the service topic for a regular path.

vooon:
Why you need service? What you need is simple publisher, that will publish on reached message rx.
Then your node just subscribe and wait for reached be published, that will work almost same as service, but without return path.
But i don't see any good usage of that `success` field.
Note that services have higher overhead than regular topics.

gui2dev:
@vooon I see your point there. Gonna change asap.

mavros/src/plugins/waypoint.cpp
```diff
@@ -154,6 +155,10 @@ class WaypointPlugin : public plugin::PluginBase {
 
 		wp_nh.param("pull_after_gcs", do_pull_after_gcs, true);
 
+		wp_nh.param("trig_after_reached", do_signal_reached, false);
```
vooon: That parameter not needed, since you use topics.

mavros/src/plugins/waypoint.cpp
```diff
@@ -154,6 +155,10 @@ class WaypointPlugin : public plugin::PluginBase {
...
+	
+		wp_reached_pub = wp_nh.advertise<std_msgs::UInt16>("reached",1,true);
```
vooon: Unfortunately std_msgs do not have stamped uint16 msg, so custom WaypointReached.msg is fine.

mavros/src/plugins/waypoint.cpp
```diff
@@ -222,11 +229,14 @@ class WaypointPlugin : public plugin::PluginBase {
 	ros::Timer shedule_timer;
 	bool do_pull_after_gcs;
 	bool enable_partial_push;
+	bool do_signal_reached;
+
+	std::string trig_srv_name;
```
vooon: Unused.

mavros/src/plugins/waypoint.cpp
```diff
@@ -222,11 +229,14 @@ class WaypointPlugin : public plugin::PluginBase {
...
 
 	bool reshedule_pull;
 
 	static constexpr int BOOTUP_TIME_MS = 15000;	//! system startup delay before start pull
-	static constexpr int LIST_TIMEOUT_MS = 30000;	//! Timeout for pull/push operations
+	static constexpr int LIST_TIMEOUT_MS = 60000;	//! Timeout for pull/push operations
```
vooon: Should not be in that PR.

mavros/src/plugins/waypoint.cpp
```diff
@@ -403,6 +413,11 @@ class WaypointPlugin : public plugin::PluginBase {
 	{
 		/* in QGC used as informational message */
 		ROS_INFO_NAMED("wp", "WP: reached #%d", mitr.seq);
+		if (do_signal_reached) {
```
vooon: Condition not needed.

vooon:
I think i do requested changes myself. @gui2dev thanks!

gui2dev:
My pleasure !

PR #860, opened 2017-11-13
GPS fix's frame_id changed to body-fixed.

pavloblindnology:
According to [ROS](http://docs.ros.org/jade/api/sensor_msgs/html/msg/NavSatFix.html), frame_id of NavSatFix message is 
> This is a Euclidean frame relative to the vehicle, not a reference  ellipsoid.

So, usually, this should be `base_link`, not `map`.
Some ROS packages (like [robot_localization](http://wiki.ros.org/robot_localization)) rely on it to be specified exactly this way. 

This pull request adds child_frame_id parameter for global_position plugin and uses it for 

- global_position/global/header/frame_id
- global_position/raw/fix/header/frame_id
- global_position/local/child_frame_id (to make it complete)

### DIFF
--- a/mavros/src/plugins/global_position.cpp
+++ b/mavros/src/plugins/global_position.cpp
@@ -116,8 +116,8 @@ private:
 	ros::Subscriber gp_set_global_origin_sub;
 	ros::Subscriber hp_sub;
 
-	std::string frame_id;       //!< origin frame for topic headers
-	std::string child_frame_id; //!< body-fixed frame for topic headers
+	std::string frame_id;				//!< origin frame for topic headers
+	std::string child_frame_id;	//!< body-fixed frame for topic headers
 	std::string tf_frame_id;	//!< origin for TF
 	std::string tf_global_frame_id;	//!< global origin for TF
 	std::string tf_child_frame_id;	//!< frame for TF and Pose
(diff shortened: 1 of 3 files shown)


mavros/src/plugins/global_position.cpp
```diff
@@ -115,7 +116,8 @@ class GlobalPositionPlugin : public plugin::PluginBase {
 	ros::Subscriber gp_set_global_origin_sub;
 	ros::Subscriber hp_sub;
 
-	std::string frame_id;		//!< frame for topic headers
+	std::string frame_id;       //!< origin frame for topic headers
```
TSC21: Can you fix this indentation please?

mavros/src/plugins/global_position.cpp
```diff
@@ -116,8 +116,8 @@ class GlobalPositionPlugin : public plugin::PluginBase {
 	ros::Subscriber gp_set_global_origin_sub;
 	ros::Subscriber hp_sub;
 
-	std::string frame_id;       //!< origin frame for topic headers
-	std::string child_frame_id; //!< body-fixed frame for topic headers
+	std::string frame_id;				//!< origin frame for topic headers
```
TSC21: Two less tabs and it will be good :)

pavloblindnology: What is the convention for Tab - 8 spaces?

TSC21: It's the current one being used. Also accepted by the `uncrustify` config.

pavloblindnology: Done.

TSC21:
Thanks!

TSC21:
@vooon Travis CI is not triggering at all...

Issue #226, opened 2015-02-26
Setpoint plugins documentation (0.11)


vooon:
Need check that all setpoint plugins documented.

- [x] accel
- [x] attitude
- [x] position
- [x] velocity
- [x] actuator_control

PR #863, opened 2017-11-14
RPM plugin added for reporting Ardupilot RPM #226 message

pavloblindnology:
RPM plugin added for reporting Ardupilot RPM # 226 message
http://mavlink.org/messages/ardupilotmega#RPM

### DIFF
--- a/mavros_extras/mavros_plugins.xml
+++ b/mavros_extras/mavros_plugins.xml
@@ -26,6 +26,9 @@
 	<class name="rangefinder" type="mavros::extra_plugins::RangefinderPlugin" base_class_type="mavros::plugin::PluginBase">
 		<description>Publish RANGEFINDER message data from FCU sensors in companion computer.</description>
 	</class>
+	<class name="rpm" type="mavros::extra_plugins::RPMPlugin" base_class_type="mavros::plugin::PluginBase">
+		<description>Publish RPM message data from FCU sensors in companion computer.</description>
+	</class>
 	<class name="vibration" type="mavros::extra_plugins::VibrationPlugin" base_class_type="mavros::plugin::PluginBase">
 		<description>Publish VIBRATION message data from FCU.</description>
 	</class>
--- a/mavros_extras/mavros_plugins.xml
+++ b/mavros_extras/mavros_plugins.xml
@@ -27,7 +27,7 @@
 		<description>Publish RANGEFINDER message data from FCU sensors in companion computer.</description>
 	</class>
 	<class name="rpm" type="mavros::extra_plugins::RPMPlugin" base_class_type="mavros::plugin::PluginBase">
-		<description>Publish RPM message data from FCU sensors in companion computer.</description>
+		<description>Publish RPM message data from FCU sensors in companion computer. (Ardupilot only).</description>
 	</class>
 	<class name="vibration" type="mavros::extra_plugins::VibrationPlugin" base_class_type="mavros::plugin::PluginBase">
 		<description>Publish VIBRATION message data from FCU.</description>
--- a/mavros_extras/src/plugins/rpm.cpp
+++ b/mavros_extras/src/plugins/rpm.cpp
@@ -1,0 +1,67 @@
+/**
+ * @brief RPM plugin
+ * @file rpm.cpp
+ * @author Pavlo Kolomiiets <pkolomiets@gmail.com>
+ *
+ * @addtogroup plugin
+ * @{
+ */
+/*
+ * Copyright 2017 Pavlo Kolomiiets.
+ *
+ * This file is part of the mavros package and subject to the license terms
+ * in the top-level LICENSE file of the mavros repository.
+ * https://github.com/mavlink/mavros/tree/master/LICENSE.md
+ */
+
+#include <mavros/mavros_plugin.h>
+#include <mavros_msgs/RPM.h>
+
+namespace mavros {
+namespace extra_plugins {
+/**
+ * @brief Ardupilot rpm plugin.
+ *
+ * This plugin allows publishing RPM sensor data from Ardupilot FCU to ROS.
+ *
+ */
+class RPMPlugin : public plugin::PluginBase {
+public:
+	RPMPlugin() : PluginBase(),
+		nh("~")
+	{ }
+
+	void initialize(UAS &uas_)
+	{
+		PluginBase::initialize(uas_);
+
+		rpm_pub = nh.advertise<mavros_msgs::RPM>("rpm", 10);
+	}
+
+	Subscriptions get_subscriptions()
+	{
+		return {
+			make_handler(&RPMPlugin::handle_rpm)
+		};
+	}
+
+private:
+	ros::NodeHandle nh;
+
+	ros::Publisher rpm_pub;
+
+	void handle_rpm(const mavlink::mavlink_message_t *msg, mavlink::ardupilotmega::msg::RPM &rpm) {
+		auto rpm_msg = boost::make_shared<mavros_msgs::RPM>();
+		rpm_msg->header.stamp = ros::Time::now();
+		rpm_msg->header.frame_id = "";
+		rpm_msg->rpm.push_back(rpm.rpm1);
+		rpm_msg->rpm.push_back(rpm.rpm2);
+
+		rpm_pub.publish(rpm_msg);
+	}
+};
+}	// namespace extra_plugins
+}	// namespace mavros
+
+#include <pluginlib/class_list_macros.h>
+PLUGINLIB_EXPORT_CLASS(mavros::extra_plugins::RPMPlugin, mavros::plugin::PluginBase)
--- a/mavros_extras/src/plugins/rpm.cpp
+++ b/mavros_extras/src/plugins/rpm.cpp
@@ -48,14 +48,20 @@ public:
 private:
 	ros::NodeHandle nh;
 
+	std::vector<float> rpms;
+
 	ros::Publisher rpm_pub;
 
 	void handle_rpm(const mavlink::mavlink_message_t *msg, mavlink::ardupilotmega::msg::RPM &rpm) {
+
+		rpms.resize(2);
+		rpms[0] = rpm.rpm1;
+		rpms[1] = rpm.rpm2;
+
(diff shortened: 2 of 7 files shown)


mavros_extras/src/plugins/rpm.cpp
```diff
@@ -0,0 +1,67 @@
...
+		return {
+			make_handler(&RPMPlugin::handle_rpm)
+		};
+	}
+
+private:
+	ros::NodeHandle nh;
+
+	ros::Publisher rpm_pub;
+
```
TSC21: Can you please add `Doxygen` like docs this as in `imu.cpp`? Ex: https://github.com/mavlink/mavros/blob/master/mavros/src/plugins/imu.cpp#L254-L259.
Thanks

mavros_extras/src/plugins/rpm.cpp
```diff
@@ -0,0 +1,67 @@
...
+	void handle_rpm(const mavlink::mavlink_message_t *msg, mavlink::ardupilotmega::msg::RPM &rpm) {
+		auto rpm_msg = boost::make_shared<mavros_msgs::RPM>();
+		rpm_msg->header.stamp = ros::Time::now();
+		rpm_msg->header.frame_id = "";
+		rpm_msg->rpm.push_back(rpm.rpm1);
```
TSC21: You have to first resize the array. Example: https://github.com/mavlink/mavros/blob/master/mavros/src/plugins/rc_io.cpp#L162

mavros_extras/mavros_plugins.xml
```diff
@@ -26,6 +26,9 @@
 	<class name="rangefinder" type="mavros::extra_plugins::RangefinderPlugin" base_class_type="mavros::plugin::PluginBase">
 		<description>Publish RANGEFINDER message data from FCU sensors in companion computer.</description>
 	</class>
+	<class name="rpm" type="mavros::extra_plugins::RPMPlugin" base_class_type="mavros::plugin::PluginBase">
+		<description>Publish RPM message data from FCU sensors in companion computer.</description>
```
TSC21: Please put an annotation referring this is a Ardupilot specific msg: Example: `Publish RPM message data from FCU sensors in companion computer. (Ardupilot only)`

pavloblindnology:
The intent is to compute wheel odometry, as currently this feature is very raw in Ardupilot beta firmware.

mavros_msgs/msg/RPM.msg
```diff
@@ -0,0 +1,4 @@
+# RPM message
+
+std_msgs/Header header
+float32[] rpm
```
TSC21: RPM msg just has two fields for RPM. So why not a fixed size array?

TSC21:
Also I'm thinking that this plugin would be useful if it could be written as a plugin to write wheel odometry directly. So it could be called `wheel_odometry` and would publish Odmetry msgs and TF from it based on the RPM msgs received. Something like this: https://answers.ros.org/question/241602/get-odometry-from-wheels-encoders/ ; https://answers.ros.org/question/231657/using-rpm-as-a-form-of-wheel-odometry/.
I can merge this plugin but I think it would be more useful for all the community with land robots using Ardupilot to directly extract the Odometry from MAVROS and the more complex stuff do it on their won. What's your opinion @pavloblindnology?

mavros_extras/src/plugins/rpm.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+	/**
+	 * @brief Handle RPM MAVlink (Ardupilot) message.
+	 * Message specification: http://mavlink.org/messages/ardupilotmega#RPM
+	 * @param msg	Received Mavlink msg
+	 * @param rpm	RPM msg
+	 */
+	void handle_rpm(const mavlink::mavlink_message_t *msg, mavlink::ardupilotmega::msg::RPM &rpm) {
+
+		rpms.resize(2);
```
TSC21: If `rpm_msg->rpm` is a `float32[2]`, you don't need to create a vector and do this operation.

mavros_msgs/msg/RPM.msg
TSC21: I would also advise you to add a timestamp field to it on the Mavlink msg (proposing a PR for it). It is much more accurate to have the time that data was generated than having the time that you receive the msg propagated. Though for ground robots it's not really relevant, at least we could have timesync between the onboard companion and the autopilot.

mavros_extras/src/plugins/rpm.cpp
pavloblindnology: Maybe in the future Ardupilot will report more rpm-s (4 wheel encoders, for example), so vector was chosen for extensibility.

mavros_msgs/msg/RPM.msg
pavloblindnology: Timestamping is a good idea - I just didn't know the messages are allowed to be modified for compatibility reasons. If yes, then of course I will add it.

pavloblindnology:
Regarding wheel_odometry. I was thinking about that. Don't know if it's really useful. Wheel odometry is going to be computed inside Ardupilot (and with more precision, because they are using encoder's ticks directly, and not converting them to RPMs, which, I suppose, reduce the accuracy). So, this whole thing was conceived as a temporary solution to add wheel odometry to the robot while it is still raw in Ardupilot firmware, and while all the stuff is connected through Pixhawk. So, finally, for odometry\positioning people will use either pure Pixhawk with everything done internally, or use ROS with encoders and other stuff connected directly to the main computer (without Pixhawk at all). In both cases there is no need in wheel odometry implemented as a mavros plugin. In 1st case - bcz it's done internally, in 2nd case - bcz you don't use mavros and you need separate node for wheel odometry.
What do you think @TSC21?

pavloblindnology:
And if still computing wheel odometry inside mavros plugin, then I think better idea is to report it as geometry_msgs/TwistWithCovarianceStamped - which is exactly information wheel encoders produce, and which can be digested to smth like robot_localization. nav_msgs/Odometry is too redundant I think - if fused with some other odometry message - you'll have two different poses with say small covariances - which will result in jumping from one pose to another (this behaviour is described in robot_localization package).

mavros_extras/src/plugins/rpm.cpp
vooon: I think variable length is ok, but you do not need `rpms` member. Just use local variable, or do:
```cpp
rpm_msg->rpm.resize(2);
rpm_msg->rpm[0] = rpm.rpm1;
rpm_msg->rpm[1] = rpm.rpm2;
```

mavros_msgs/msg/RPM.msg
TSC21: > Timestamping is a good idea - I just didn't know the messages are allowed to be modified for compatibility reasons. If yes, then of course I will add it.

Just create a PR against Mavlink repo with the new field.

TSC21:
> Regarding wheel_odometry. I was thinking about that. Don't know if it's really useful. Wheel odometry is going to be computed inside Ardupilot (and with more precision, because they are using encoder's ticks directly, and not converting them to RPMs, which, I suppose, reduce the accuracy).

My question is: so why don't create a msg to use ticks instead of RPM?

> So, this whole thing was conceived as a temporary solution to add wheel odometry to the robot while it is still raw in Ardupilot firmware, and while all the stuff is connected through Pixhawk.

What would be the output from it after?

> So, finally, for odometry\positioning people will use either pure Pixhawk with everything done internally, or use ROS with encoders and other stuff connected directly to the main computer (without Pixhawk at all).

You are right. But in this case, we are creating a solution for something for the current existent structure. Meaning that we are creating a plugin that uses encoders data coming on Mavlink msg to process the wheel odometry. Which IMO is what is useful for this case. Cause as you say, people can use encoders directly connected to the OBC or they can just use the data coming from Ardupilot to compute it directly on the plugin and then use that data for robot localization.

I think that while it is not done internally, we could had a much more complete plugin which provides useful wheel odometry. I don't think it makes sense to just add a plugin as a RPM parser to ROS when we can extend it to do something more useful.

> which will result in jumping from one pose to another (this behaviour is described in robot_localization package).

If you have a position estimator filter (like a KF) that behavior is avoided. What I suggest is that we add the option to: 1. publish `nav_msgs/Odometry`, which gives an estimation of the pos/velocity already; 2. publish `geometry_msgs/TwistWithCovarianceStamped` as you suggest, to be consumed by a localization package.

> Also, nav_msgs/Odometry and corresponding TFs need to know initial position and orientation connected with local_position\global_position origin, to fuse them with other sources. And this will need cooperating with those plugins.

That seems reasonable IMO. We already do that on other plugins.

So do you think you can tackle the approaches we discussed above? I think it is much more reasonable and will come up to be much more useful to everyone. Also, I would suggest to have a Ardupilot msg that provides the encoder ticks instead of RPM.

mavros_extras/src/plugins/rpm.cpp
pavloblindnology: @vooon Will update.

mavros_msgs/msg/RPM.msg
TSC21: Btw I see you created a PR for Mavlink to add the timestamp: why not extend it to already support an array instead of having `rpm1` and `rpm2` fields?

pavloblindnology:
@TSC21 Ok. Let's do it as you suggest. The only concern is about new message for ticks, as it adds burden to throughput.

pavloblindnology:
And maybe it's better to report not ticks but angle rotated. This will require less modifications to Ardupilot and will save us from specifying CPR in mavros plugin.

TSC21:
> And maybe it's better to report not ticks but angle rotated. This will require less modifications to Ardupilot and will save us from specifying CPR in mavros plugin.

I leave it to your decision based on what you are implementing and what can be useful to everyone.

pavloblindnology:
Even more, not angle rotated but distance travelled - saves us specifying the radius.

pavloblindnology:
Is it better to create new PR or extend this one?

TSC21:
I think it is better to create a new one.

pavloblindnology:
Then I'm closing this PR.

Issue #867, opened 2017-11-20
install_geographiclib_datasets.sh fails on empty directories


fnoop:
### Issue details

If download of the geographiclib datasets fails for any reason (like it hangs downloading or can't reach), it leaves empty subdirectories of /usr/share/GeographicLib.  The bash script only checks for existence of these directories, not of the datasets within so any subsequent runs of the script do nothing even if the datasets are not present.

Instead, it should check for the actual files.  Change:
```
	if [[ -d "/usr/share/GeographicLib/$dir" or -d "/usr/local/share/GeographicLib/$dir" ]]; then
		return
	fi
```
 to (eg.)
```
	files=$(shopt -s nullglob dotglob; echo /usr/share/GeographicLib/$dir/$model* /usr/local/share/GeographicLib/$dir/$model*)
	if (( ${#files} )); then
		return
	fi
```

### MAVROS version and platform

Mavros: master
ROS: Lunar
Ubuntu: Debian 9 / Ubuntu 16.04

TSC21:
Thanks for pointing that @fnoop. Would you mind making a PR with this change? Thanks

fnoop:
Sure, PR #868 raised

TSC21:
Closed by #868

PR #868, opened 2017-11-20
Improve install_geographiclib_datasets.sh script

fnoop:
PR to support issue #867 

 - Ensure dataset files actually exist, not just directories which are empty if a download attempt fails
 - Improve output, turn off debug
 - Use builtin hash instead of which, more reliable

### DIFF
--- a/mavros/scripts/install_geographiclib_datasets.sh
+++ b/mavros/scripts/install_geographiclib_datasets.sh
@@ -15,7 +15,8 @@ run_get() {
 	local tool="$2"
 	local model="$3"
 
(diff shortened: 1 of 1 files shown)


khancyr:
Looks good.
But I don't understand the hash change, I can find pointer about that on google, could you briefly explain what was the problem or the improvement with hash ?

fnoop:
@khancyr https://stackoverflow.com/a/677212
Also the which command as it was used output the path if found to stdout, which was messy output wise

khancyr:
nice ! thanks !

PR #870, opened 2017-11-21
mavros.ftp error correction in python

urpylka:
This patch solves two problems with using mavros.ftp on python.

First - close ftp-resource on Ctrl+C interrupt
Second - reset connection when an mavftp error (Busy, File already opened) occurs

### DIFF
--- a/mavros/src/mavros/ftp.py
+++ b/mavros/src/mavros/ftp.py
@@ -76,6 +76,7 @@ class FTPFile(object):
         try:
             ret = open_(file_path=path, mode=m)
         except rospy.ServiceException as ex:
+            reset_server()
             raise IOError(str(ex))
 
         _check_raise_errno(ret)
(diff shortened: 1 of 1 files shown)


TSC21:
@urpylka can you please describe the patch? (intent, changes?)

mavros/src/mavros/ftp.py
```diff
@@ -53,6 +53,9 @@ def __init__(self, name, mode):
     def __del__(self):
         self.close()
 
+    def __exit__(self):
+        self.close()
+
```
vooon: Class already have `__exit__()` method, look bottom.

mavros/src/mavros/ftp.py
```diff
@@ -73,6 +76,7 @@ def open(self, path, mode):
         try:
             ret = open_(file_path=path, mode=m)
         except rospy.ServiceException as ex:
+            reset_server()
```
vooon: I'm not sure that reset here is good idea. Higher level code should do that.

urpylka: Ok, I agree with you

mavros/src/mavros/ftp.py
urpylka: It's true. I was not careful

PR #875, opened 2017-11-27
Add a service to modify setpoint pos/vel mav frame

khancyr:


TSC21:
@khancyr what's the benefit of using a service for this instead of a simple parameter set? Also I'm adding the `MAV_FRAME` `to_string` and `from_str` on #311 (https://github.com/mavlink/mavros/pull/311/files#diff-16d305e0bc99d7473027f67b442960d3R139). Maybe something like that is enough, where we set the parameter on the launch file and consider it that way. Dynamic reconfigure can also be used. Unless you want to set the frame online using another node triggering the service. Is that the purpose?

mavros/src/plugins/setpoint_position.cpp
```diff
@@ -93,12 +100,25 @@ class SetpointPositionPlugin : public plugin::PluginBase,
 		 */
 		const uint16_t ignore_all_except_xyz_y = (1 << 11) | (7 << 6) | (7 << 3);
 
-		auto p = ftf::transform_frame_enu_ned(Eigen::Vector3d(tr.translation()));
-		auto q = ftf::transform_orientation_enu_ned(
-					ftf::transform_orientation_baselink_aircraft(Eigen::Quaterniond(tr.rotation())));
+		auto p = [&] {
+			if (static_cast<MAV_FRAME>(mav_frame) == MAV_FRAME::BODY_NED || static_cast<MAV_FRAME>(mav_frame) == MAV_FRAME::BODY_OFFSET_NED) {
```
TSC21: For `BODY_OFFSET_NED` we should have a way of setting the offset. That would of course require adding some new functions to `frame_tf.h`. But guess we can spare that work since the idea in the near future is to change everything from static transforms to using `tf2`.

TSC21:
Looks like Travis has awake :+1:

khancyr:
yep, in first approach I use param, but then I realized that it will be more usefull to use a service to switch on frame

TSC21:
> yep, in first approach I use param, but then I realized that it will be more useful to use a service to switch on frame

Can we have both options, where the service overwrites the param defined value if used?

khancyr:
ok I will do that tomorrow!

mavros/src/plugins/setpoint_position.cpp
```diff
@@ -93,12 +100,25 @@ class SetpointPositionPlugin : public plugin::PluginBase,
...
+				return ftf::transform_frame_baselink_aircraft(Eigen::Vector3d(tr.translation()));
+			} else {
+				return ftf::transform_frame_enu_ned(Eigen::Vector3d(tr.translation()));
+			}
+		} ();
+
+		auto q = [&] {
```
vooon: A bit confuses me, use full `[&] () { ... } ();`

mavros/src/plugins/setpoint_position.cpp
```diff
@@ -68,13 +72,16 @@ class SetpointPositionPlugin : public plugin::PluginBase,
...
 	ros::Subscriber setpoint_sub;
+	ros::ServiceServer mav_frame_srv;
 
 	std::string tf_frame_id;
 	std::string tf_child_frame_id;
 
 	bool tf_listen;
 	double tf_rate;
 
+	uint8_t mav_frame = utils::enum_value(MAV_FRAME::LOCAL_NED);
```
vooon: Better keep type `MAV_FRAME`.

mavros/src/plugins/setpoint_position.cpp
```diff
@@ -93,12 +100,25 @@ class SetpointPositionPlugin : public plugin::PluginBase,
...
+		auto p = [&] {
+			if (static_cast<MAV_FRAME>(mav_frame) == MAV_FRAME::BODY_NED || static_cast<MAV_FRAME>(mav_frame) == MAV_FRAME::BODY_OFFSET_NED) {
+				return ftf::transform_frame_baselink_aircraft(Eigen::Vector3d(tr.translation()));
+			} else {
+				return ftf::transform_frame_enu_ned(Eigen::Vector3d(tr.translation()));
+			}
+		} ();
+
+		auto q = [&] {
+			if (static_cast<MAV_FRAME>(mav_frame) == MAV_FRAME::BODY_NED || static_cast<MAV_FRAME>(mav_frame) == MAV_FRAME::BODY_OFFSET_NED) {
```
vooon: With correct type only what you need: `mav_frame == MAV_FRAME::BODY_NED || ...`

mavros/src/plugins/setpoint_position.cpp
```diff
@@ -93,12 +100,25 @@ class SetpointPositionPlugin : public plugin::PluginBase,
...
+				return ftf::transform_orientation_baselink_aircraft(Eigen::Quaterniond(tr.rotation()));
+			} else {
+				return ftf::transform_orientation_enu_ned(
+							ftf::transform_orientation_baselink_aircraft(Eigen::Quaterniond(tr.rotation())));
+			}
+		} ();
 
 		set_position_target_local_ned(stamp.toNSec() / 1000000,
-					utils::enum_value(MAV_FRAME::LOCAL_NED),
+					mav_frame,
```
vooon: `utils::enum_value(mav_frame)`

mavros_msgs/srv/SetMavFrame.srv
```diff
@@ -0,0 +1,19 @@
...
+uint8 MISSION = 2
+uint8 GLOBAL_RELATIVE_ALT = 3
+uint8 LOCAL_ENU = 4
+uint8 GLOBAL_INT = 5
+uint8 GLOBAL_RELATIVE_ALT_INT = 6
+uint8 LOCAL_OFFSET_NED = 7
+uint8 BODY_NED = 8
+uint8 BODY_OFFSET_NED = 9
+uint8 GLOBAL_TERRAIN_ALT = 10
+uint8 GLOBAL_TERRAIN_ALT_INT = 11
```
vooon: Better if that part of message will be generated from pymavlink using cog.py.

mavros/src/plugins/setpoint_position.cpp
khancyr: done

mavros/src/plugins/setpoint_position.cpp
khancyr: done

mavros/src/plugins/setpoint_position.cpp
khancyr: done

mavros/src/plugins/setpoint_position.cpp
khancyr: done

mavros_msgs/srv/SetMavFrame.srv
khancyr: I use cog to generate those on utils, but I don't see how to do it for the .srv , any clues ?

TSC21: example: https://github.com/mavlink/mavros/blob/master/mavros_msgs/msg/ADSBVehicle.msg

khancyr: thanks ! Done !

mavros/src/plugins/setpoint_velocity.cpp
```diff
@@ -62,7 +69,8 @@ class SetpointVelocityPlugin : public plugin::PluginBase,
 	ros::Subscriber vel_unstamped_sub;
 	ros::ServiceServer mav_frame_srv;
 
-	uint8_t mav_frame = utils::enum_value(MAV_FRAME::LOCAL_NED);
+	std::string mav_frame_str;
```
vooon: For what purpose store string repr in class? I think that it may be local variable.

mavros_msgs/srv/SetMavFrame.srv
```diff
@@ -1,18 +1,36 @@
...
+uint8 BODY_OFFSET_NED = 9                # Offset in body NED frame. This makes sense if adding setpoints to the current flight path, to avoid an obstacle - e.g. useful to command 2 m/s^2 acceleration to the east.
+uint8 GLOBAL_TERRAIN_ALT = 10            # Global coordinate frame with above terrain level altitude. WGS84 coordinate system, relative altitude over terrain with respect to the waypoint coordinate. First value / x: latitude in degrees, second value / y: longitude in degrees, third value / z: positive altitude in meters with 0 being at ground level in terrain model.
+uint8 GLOBAL_TERRAIN_ALT_INT = 11        # Global coordinate frame with above terrain level altitude. WGS84 coordinate system, relative altitude over terrain with respect to the waypoint coordinate. First value / x: latitude in degrees*10e-7, second value / y: longitude in degrees*10e-7, third value / z: positive altitude in meters with 0 being at ground level in terrain model.
+# [[[end]]] (checksum: 4c32e7225de1a74ff7c284b9935270e3)
```
vooon: In other messages we have prefix `FRAME_`.
Also possible good idea to make MavFrame message, like CommandCode.msg.

TSC21: > In other messages we have prefix `FRAME_`.

@khancyr can you address this please? Thanks!

mavros/src/plugins/setpoint_velocity.cpp
khancyr: ok done !

mavros_msgs/srv/SetMavFrame.srv
khancyr: Ok I have change that

PR #876, opened 2017-11-28
add some help for cog on readme

khancyr:


mavros/README.md
```diff
@@ -255,6 +255,17 @@ Contributing
  - with `catkin tools`, issue `catkin run_tests`;
 11. If everything goes as planned, push the changes (`git push -u <remote_repo> <feature_branch>`) and issue a pull request.
 
+NOTE: cog.py can be use to generate/update enum from mavlink.
+To install it :
+
+    sudo pip install --user cogapp
```
vooon: You don't need sudo when use `--user` (module will be installed into `$HOME/.local/lib/python<ver>/...`).

vooon:
Perhaps we need to separate contribution guide from readme.
And there we may be more specific about tools used.

mavros/README.md
khancyr: yep right!

khancyr:
I think that having a separated contrib.md could be an excellent idea!

PR #883, opened 2017-12-09
move member variable earth initialization

ShingoMatsuura:
I think it is better the earth variable is initialized at constructor.


TSC21:
@ShingoMatsuura sounds good. Would you mind doing the same for `setpoint_position` and `global_position` plugins? Thanks.

ShingoMatsuura:
In setpoint_position and global_position, they use earth as local variable and initialize before use it. In fake_gps, earth's forward and reverse function didn't work normally so I initialized it explicitly.

ShingoMatsuura:
Thank you for merge!

PR #889, opened 2017-12-13
Fix vision odom.

jgoppert:
These fixes are necessary to get vision based heading working correctly for me.

### DIFF
--- a/mavros/src/lib/uas_data.cpp
+++ b/mavros/src/lib/uas_data.cpp
@@ -53,6 +53,7 @@ UAS::UAS() :
 
 	// send static transform from "local_origin" (ENU) to "local_origin_ned" (NED)
 	publish_static_transform("local_origin", "local_origin_ned", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, M_PI_2)));
+	publish_static_transform("fcu", "fcu_frd", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, 0)));
 }
 
 /* -*- heartbeat handlers -*- */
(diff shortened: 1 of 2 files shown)


TSC21:
what is `fcu_frd`?

jgoppert:
ROS body frame is forward-left-up (flu), px4 body frame is  forward-right-down (frd).

mavros/src/lib/uas_data.cpp
```diff
@@ -53,6 +53,7 @@ UAS::UAS() :
 
 	// send static transform from "local_origin" (ENU) to "local_origin_ned" (NED)
 	publish_static_transform("local_origin", "local_origin_ned", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, M_PI_2)));
+	publish_static_transform("fcu", "fcu_frd", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, 0)));
```
TSC21: we actually considered this on https://github.com/mavlink/mavros/blob/master/mavros/src/lib/ftf_frame_conversions.cpp#L32-L37. So I was feeling strange that you were not considering this here also, since for vision plugin, that is also considered: https://github.com/mavlink/mavros/blob/master/mavros_extras/src/plugins/vision_pose_estimate.cpp#L127-L129

TSC21: can we actually call them `body_nwu` and `body_ned` instead?

jgoppert: Yeah, that's fine with me.

TSC21: @jgoppert please update the PR with the naming convention above

TSC21: We will keep the convention as is.

ChristophTobler:
What is actually the difference between the plugins `vision_pose_estimate` and `odom`?

TSC21:
It's the published Mavlink msgs

PR #891, opened 2017-12-14
convert accel/gyro to enu not forward-left-up

ChristophTobler:
The variables even say ENU but are actually converted to forward-left-up. I guess `mag_field` would have to be ENU as well?

### DIFF
--- a/mavros/src/plugins/imu.cpp
+++ b/mavros/src/plugins/imu.cpp
@@ -112,8 +112,8 @@ private:
 	bool has_raw_imu;
 	bool has_scaled_imu;
 	bool has_att_quat;
-	Eigen::Vector3d linear_accel_vec_enu;
-	Eigen::Vector3d linear_accel_vec_ned;
+	Eigen::Vector3d linear_accel_vec_flu;
+	Eigen::Vector3d linear_accel_vec_frd;
 	ftf::Covariance3d linear_acceleration_cov;
 	ftf::Covariance3d angular_velocity_cov;
 	ftf::Covariance3d orientation_cov;
@@ -144,14 +144,14 @@ private:
 
 	/**
 	 * @brief Fill and publish IMU data message.
-	 * @param time_boot_ms		Message timestamp (not syncronized)
-	 * @param orientation_enu	Orientation in the base_link ENU frame
-	 * @param orientation_ned	Orientation in the aircraft NED frame
-	 * @param gyro_enu		Angular velocity/rate in the base_link ENU frame
-	 * @param gyro_ned		Angular velocity/rate in the aircraft NED frame
+	 * @param time_boot_ms     Message timestamp (not syncronized)
+	 * @param orientation_enu  Orientation in the base_link ENU frame
+	 * @param orientation_ned  Orientation in the aircraft NED frame
+	 * @param gyro_flu         Angular velocity/rate in the base_link Forward-Left-Up frame
+	 * @param gyro_frd         Angular velocity/rate in the aircraft Forward-Right-Down frame
 	 */
 	void publish_imu_data(uint32_t time_boot_ms, Eigen::Quaterniond &orientation_enu,
-				Eigen::Quaterniond &orientation_ned, Eigen::Vector3d &gyro_enu, Eigen::Vector3d &gyro_ned)
+				Eigen::Quaterniond &orientation_ned, Eigen::Vector3d &gyro_flu, Eigen::Vector3d &gyro_frd)
 	{
 		auto imu_ned_msg = boost::make_shared<sensor_msgs::Imu>();
 		auto imu_enu_msg = boost::make_shared<sensor_msgs::Imu>();
@@ -165,12 +165,12 @@ private:
 		tf::quaternionEigenToMsg(orientation_ned, imu_ned_msg->orientation);
 
 		// Convert from Eigen::Vector3d to geometry_msgs::Vector3
-		tf::vectorEigenToMsg(gyro_enu, imu_enu_msg->angular_velocity);
-		tf::vectorEigenToMsg(gyro_ned, imu_ned_msg->angular_velocity);
+		tf::vectorEigenToMsg(gyro_flu, imu_enu_msg->angular_velocity);
+		tf::vectorEigenToMsg(gyro_frd, imu_ned_msg->angular_velocity);
 
 		// Eigen::Vector3d from HIGHRES_IMU or RAW_IMU, to geometry_msgs::Vector3
-		tf::vectorEigenToMsg(linear_accel_vec_enu, imu_enu_msg->linear_acceleration);
-		tf::vectorEigenToMsg(linear_accel_vec_ned, imu_ned_msg->linear_acceleration);
+		tf::vectorEigenToMsg(linear_accel_vec_flu, imu_enu_msg->linear_acceleration);
+		tf::vectorEigenToMsg(linear_accel_vec_frd, imu_ned_msg->linear_acceleration);
 
 		// Pass ENU msg covariances
 		imu_enu_msg->orientation_covariance = orientation_cov;
@@ -206,25 +206,25 @@ private:
 
 	/**
 	 * @brief Fill and publish IMU data_raw message; store linear acceleration for IMU data
-	 * @param header	Message frame_id and timestamp
-	 * @param gyro		Orientation in the base_link ENU frame
-	 * @param accel_enu	Linear acceleration in the base_link ENU frame
-	 * @param accel_ned	Linear acceleration in the aircraft NED frame
+	 * @param header      Message frame_id and timestamp
+	 * @param gyro_flu    Orientation in the base_link Forward-Left-Up frame
+	 * @param accel_flu   Linear acceleration in the base_link Forward-Left-Up frame
+	 * @param accel_frd   Linear acceleration in the aircraft Forward-Right-Down frame
 	 */
-	void publish_imu_data_raw(std_msgs::Header &header, Eigen::Vector3d &gyro,
-				Eigen::Vector3d &accel_enu, Eigen::Vector3d &accel_ned)
+	void publish_imu_data_raw(std_msgs::Header &header, Eigen::Vector3d &gyro_flu,
+				Eigen::Vector3d &accel_flu, Eigen::Vector3d &accel_frd)
 	{
 		auto imu_msg = boost::make_shared<sensor_msgs::Imu>();
 
 		// Fill message header
 		imu_msg->header = header;
 
-		tf::vectorEigenToMsg(gyro, imu_msg->angular_velocity);
-		tf::vectorEigenToMsg(accel_enu, imu_msg->linear_acceleration);
+		tf::vectorEigenToMsg(gyro_flu, imu_msg->angular_velocity);
+		tf::vectorEigenToMsg(accel_flu, imu_msg->linear_acceleration);
 
 		// Save readings
-		linear_accel_vec_enu = accel_enu;
-		linear_accel_vec_ned = accel_ned;
+		linear_accel_vec_flu = accel_flu;
+		linear_accel_vec_frd = accel_frd;
 
 		imu_msg->orientation_covariance = unk_orientation_cov;
 		imu_msg->angular_velocity_covariance = angular_velocity_cov;
@@ -277,7 +277,7 @@ private:
 		 *  @snippet src/plugins/imu.cpp ned_ang_vel1
 		 */
 		// [ned_ang_vel1]
-		auto gyro_ned = Eigen::Vector3d(att.rollspeed, att.pitchspeed, att.yawspeed);
+		auto gyro_frd = Eigen::Vector3d(att.rollspeed, att.pitchspeed, att.yawspeed);
 		// [ned_ang_vel1]
 
 		/** The RPY describes the rotation: aircraft->NED.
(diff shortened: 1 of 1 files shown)


TSC21:
@ChristophTobler this is actually not correct. You want the IMU data to be presented in the ROS body frame, which is as you say FLU, or more correct, NWU. What actually may be considered wrong is the variable name suffixes, which should be `_nwu` and not `_enu`. Would you mind updating this PR with the changes on the naming instead?

ChristophTobler:
@TSC21 Ah, ok. Sure!

TSC21:
@ChristophTobler can you update this please?

ChristophTobler:
@TSC21 :+1: I'll wait for https://github.com/mavlink/mavros/pull/893 to go in as it will create a conflict

ChristophTobler:
@TSC21 This needed a bit more changes than I expected... Could you please go through them? Thanks!

mavros/src/plugins/imu.cpp
```diff
@@ -328,18 +328,18 @@ class IMUPlugin : public plugin::PluginBase {
 		/** MAVLink quaternion exactly matches Eigen convention.
 		 *  The RPY describes the rotation: aircraft->NED.
 		 *  It is required to change this to aircraft->base_link:
-		 *  @snippet src/plugins/imu.cpp ned->baselink->enu
+		 *  @snippet src/plugins/imu.cpp ned->baselink->nwu
 		 */
-		auto enu_baselink_orientation = ftf::transform_orientation_aircraft_baselink(
+		auto nwu_baselink_orientation = ftf::transform_orientation_aircraft_baselink(
```
TSC21: actually this is `enu_baselink_orientation` - check what is applied upfront. The orientation should be on ENU frame. Raw IMU data is in the NWU frame (ROS body frame).

mavros/src/plugins/imu.cpp
```diff
@@ -328,18 +328,18 @@ class IMUPlugin : public plugin::PluginBase {
 		/** MAVLink quaternion exactly matches Eigen convention.
 		 *  The RPY describes the rotation: aircraft->NED.
 		 *  It is required to change this to aircraft->base_link:
-		 *  @snippet src/plugins/imu.cpp ned->baselink->enu
+		 *  @snippet src/plugins/imu.cpp ned->baselink->nwu
```
TSC21: `enu` and not `nwu`

mavros/src/plugins/imu.cpp
```diff
@@ -282,22 +282,22 @@ class IMUPlugin : public plugin::PluginBase {
 
 		/** The RPY describes the rotation: aircraft->NED.
 		 *  It is required to change this to aircraft->base_link:
-		 *  @snippet src/plugins/imu.cpp ned->baselink->enu
+		 *  @snippet src/plugins/imu.cpp ned->baselink->nwu
 		 */
-		// [ned->baselink->enu]
-		auto enu_baselink_orientation = ftf::transform_orientation_aircraft_baselink(
+		// [ned->baselink->nwu]
```
TSC21: `enu`

mavros/src/plugins/imu.cpp
```diff
@@ -282,22 +282,22 @@ class IMUPlugin : public plugin::PluginBase {
...
+		auto nwu_baselink_orientation = ftf::transform_orientation_aircraft_baselink(
```
TSC21: `enu_baselink_orientation`

mavros/src/plugins/imu.cpp
```diff
@@ -282,22 +282,22 @@ class IMUPlugin : public plugin::PluginBase {
 
 		/** The RPY describes the rotation: aircraft->NED.
 		 *  It is required to change this to aircraft->base_link:
-		 *  @snippet src/plugins/imu.cpp ned->baselink->enu
+		 *  @snippet src/plugins/imu.cpp ned->baselink->nwu
```
TSC21: `enu`

mavros/src/plugins/imu.cpp
```diff
@@ -145,85 +145,85 @@ class IMUPlugin : public plugin::PluginBase {
...
+		auto imu_nwu_msg = boost::make_shared<sensor_msgs::Imu>();
 
 		// Fill message header
-		imu_enu_msg->header = m_uas->synchronized_header(frame_id, time_boot_ms);
+		imu_nwu_msg->header = m_uas->synchronized_header(frame_id, time_boot_ms);
 		imu_ned_msg->header = m_uas->synchronized_header("aircraft", time_boot_ms);
 
 		// Convert from Eigen::Quaternond to geometry_msgs::Quaternion
-		tf::quaternionEigenToMsg(orientation_enu, imu_enu_msg->orientation);
+		tf::quaternionEigenToMsg(orientation_nwu, imu_nwu_msg->orientation);
```
TSC21: orientation is in `enu`

mavros/src/plugins/imu.cpp
```diff
@@ -328,18 +328,18 @@ class IMUPlugin : public plugin::PluginBase {
...
-		auto enu_baselink_orientation = ftf::transform_orientation_aircraft_baselink(
+		auto nwu_baselink_orientation = ftf::transform_orientation_aircraft_baselink(
 					ftf::transform_orientation_ned_enu(ned_aircraft_orientation));
 
 		/** The angular velocity expressed in the aircraft frame.
 		 *  It is required to apply the static rotation to get it into the base_link frame:
 		 *  @snippet src/plugins/imu.cpp rotate_gyro
 		 */
-		auto gyro_enu = ftf::transform_frame_aircraft_baselink(gyro_ned);
+		auto gyro_nwu = ftf::transform_frame_aircraft_baselink(gyro_ned);
```
TSC21: this is right!

mavros/src/plugins/imu.cpp
```diff
@@ -145,85 +145,85 @@ class IMUPlugin : public plugin::PluginBase {
...
-	 * @param orientation_enu	Orientation in the base_link ENU frame
+	 * @param orientation_nwu	Orientation in the base_link NWU frame
 	 * @param orientation_ned	Orientation in the aircraft NED frame
-	 * @param gyro_enu		Angular velocity/rate in the base_link ENU frame
+	 * @param gyro_nwu		Angular velocity/rate in the base_link NWU frame
 	 * @param gyro_ned		Angular velocity/rate in the aircraft NED frame
 	 */
-	void publish_imu_data(uint32_t time_boot_ms, Eigen::Quaterniond &orientation_enu,
-				Eigen::Quaterniond &orientation_ned, Eigen::Vector3d &gyro_enu, Eigen::Vector3d &gyro_ned)
+	void publish_imu_data(uint32_t time_boot_ms, Eigen::Quaterniond &orientation_nwu,
```
TSC21: everything in `publish_imu_data` is on ENU frame. Everything on `publish_imu_data_raw` is on NWU frame.

mavros/src/plugins/imu.cpp
```diff
@@ -145,85 +145,85 @@ class IMUPlugin : public plugin::PluginBase {
 	/**
 	 * @brief Fill and publish IMU data message.
 	 * @param time_boot_ms		Message timestamp (not syncronized)
-	 * @param orientation_enu	Orientation in the base_link ENU frame
+	 * @param orientation_nwu	Orientation in the base_link NWU frame
```
TSC21: `enu`

mavros/include/mavros/mavros_uas.h
```diff
@@ -174,7 +174,7 @@ class UAS {
 	/**
 	 * @brief Store IMU data [ENU]
 	 */
-	void update_attitude_imu_enu(sensor_msgs::Imu::Ptr &imu);
+	void update_attitude_imu_nwu(sensor_msgs::Imu::Ptr &imu);
```
TSC21: all the changes on `mavros_uas` and `uas_data` are not correct cause the orientation frame of reference is local ENU, and not body NWU.

mavros/src/plugins/local_position.cpp
```diff
@@ -87,12 +87,12 @@ class LocalPositionPlugin : public plugin::PluginBase {
 		auto enu_velocity = ftf::transform_frame_ned_enu(Eigen::Vector3d(pos_ned.vx, pos_ned.vy, pos_ned.vz));
 
 		//--------------- Get Odom Information ---------------//
-		// Note this orientation describes baselink->ENU transform
-		auto enu_orientation_msg = m_uas->get_attitude_orientation_enu();
-		auto baselink_angular_msg = m_uas->get_attitude_angular_velocity_enu();
-		Eigen::Quaterniond enu_orientation;
-		tf::quaternionMsgToEigen(enu_orientation_msg, enu_orientation);
-		auto baselink_linear = ftf::transform_frame_enu_baselink(enu_velocity, enu_orientation.inverse());
+		// Note this orientation describes baselink->NWU transform
```
TSC21: This is all ENU frame (relative to local ENU frame, and not to NWU body frame).

ChristophTobler:
I don't fully agree. This should be NWU https://github.com/mavlink/mavros/blob/master/mavros/src/plugins/imu.cpp#L226, right? If so, then https://github.com/mavlink/mavros/blob/master/mavros/src/plugins/imu.cpp#L172 is not enu but nwu as well. Otherwise the msg would be all mixed...

TSC21:
Yes you are right - https://github.com/mavlink/mavros/blob/master/mavros/src/plugins/imu.cpp#L226-L227 should be fixed to: 
```
		linear_accel_vec_enu = ftf::transform_frame_ned_enu(accel_nwu);
		linear_accel_vec_ned = accel_ned;
```

ChristophTobler:
Ok:+1:
I guess you mean
```
		linear_accel_vec_enu = ftf::transform_frame_ned_enu(accel_ned);
		linear_accel_vec_ned = accel_ned;
```
I'll make the changes

TSC21:
>Ok:+1:
>I guess you mean
>```
>		linear_accel_vec_enu = ftf::transform_frame_ned_enu(accel_ned);
>		linear_accel_vec_ned = accel_ned;
>```
>I'll make the changes

No I really mean `linear_accel_vec_enu = ftf::transform_frame_ned_enu(accel_nwu);`, cause you already have the transform from `ned` to `nwu` when you pass the data on the argument.

ChristophTobler:
That doesn't make any sense... If you rotate nwu with z: 90 x:180 you will end up with wnd

TSC21:
@ChristophTobler OK forget what I written above and let's focus on what is stated [here](https://github.com/mavlink/mavros/blob/master/mavros/src/plugins/imu.cpp#L63-L66). So what's actually happening is that the data we are publishing in the IMU msg is a mix of frames - all the data from the IMU (being it from the raw msgs or not) is transformed from the aircraft frame (PX4 body frame - FRD or NED) to the base_link frame (ROS body frame - FLU or ENU). So the sensor data is all on the body frame. The exception is, and as you can read above, is regarding the orientation, which then is expressed WRT local ENU frame (that's why you first apply the aircraft->base_link transform and then the NED->ENU, so the orientation is WRT to ENU local frame - https://github.com/mavlink/mavros/blob/master/mavros/src/plugins/imu.cpp#L288-L289).
So what you will have in the end published by the [`imu_data_raw()`](https://github.com/mavlink/mavros/blob/master/mavros/src/plugins/imu.cpp#L214) is sensor data in the ROS body frame, i.e. the NWU frame, and by [`imu_data()`](https://github.com/mavlink/mavros/blob/master/mavros/src/plugins/imu.cpp#L153) is the sensor data also in the NWU frame, and the orientation WRT [ENU frame](https://github.com/mavlink/mavros/blob/master/mavros/src/plugins/imu.cpp#L189).
So the respective changes should reflect what I said above. I actually forgot the above detail and was thinking that the msg would have to be all on the same frame, but we actually warn people of the above. So please, if you may, just have the orientation related data with `_enu` sufix, while the rest is in `_nwu`. Leave some more comments there if you think they are required to let people know what's happening, but we actually have a warning on the beginning.

ChristophTobler:
@TSC21 I think I get it with the orientation as it is a non-fixed orientation and dependent on the attitude. I don't quite get the `base_link` though...

But in general I'm not happy with how the suffixes are used and mixed. Because, except for the orientation, it has nothing to do with NED, ENU, NWU etc. as there is no actual North, East etc. I my opinion, this should all be in local/body coordinates -> forward-left-up etc. -> e.g. `accel_flu`

TSC21:
http://www.springer.com/cda/content/document/cda_downloaddocument/9780857296344-c2.pdf?SGWID=0-0-45-1143141-p174116371
OK so together with the ROS REP, we should have the following notation then:
1. Everything related to local/geodetic frame, should be expressed on the ENU frame - for example, the orientation on the `imu` plugin, the position in the `global_position` and `local_position` plugins;
2. Everything related to body frame, WRT ROS convention, should have the suffix `_flu`, if WRT PX4 convention, suffix `_frd` - for example, the gyro, mag and accel data in the `imu` plugin, the velocities being fed to the `odom` plugin (handled differently there, but serve as example);
3. And based on the above source, `it is reasonable to assume that the directions of the vehicle-carried and local NED coordinate systems constantly coincide with each other`.
Sounds good to you @ChristophTobler?

ChristophTobler:
@TSC21 Yes, thanks! I'll update the changes:+1:

ChristophTobler:
@TSC21 Here's another go at it :)

ChristophTobler:
@TSC21 I think I have done that. Or which ones do you have in mind?

TSC21:
For example: https://github.com/mavlink/mavros/pull/891/files#diff-cf299de9fb904a063d0530de8036ed6bR321

TSC21:
Thanks @ChristophTobler!

ChristophTobler:
@TSC21 Thanks!

PR #893, opened 2017-12-15
IMU plugin: add raw IMU conversion for PX4

ChristophTobler:
This will be needed once https://github.com/ChristophTobler/Firmware/commits/pr-raw_imu_snappy goes into PX4. We need it to run ViSLAM on the Snapdragon Flight with PX4.


mavros/src/plugins/imu.cpp
```diff
@@ -433,6 +440,10 @@ class IMUPlugin : public plugin::PluginBase {
 		if (m_uas->is_ardupilotmega()) {
 			accel_ned *= MILLIG_TO_MS2;
 			accel_enu *= MILLIG_TO_MS2;
+
```
TSC21: can you delete this new line please @ChristophTobler?

ChristophTobler: done

TSC21:
Thanks @ChristophTobler

PR #897, opened 2017-12-19
travis-ci: fix

TSC21:


TSC21:
Seems like it's having trouble with the dependency on GeographicLib. Will close this for further investigation

PR #908, opened 2017-12-27
mavros_extras: Command_ack plugin added.

pavloblindnology:
Plugin for listening to `COMMAND_ACK` messages from FCU and publishing them to ROS.


vooon:
I think is is very specific plugin. Not sure should it be in master or not. And it should be blacklisted by default.

TSC21:
I think this should be integrated on the `command` plugin instead.

vooon:
Definitely nope. Catching ACK needed only in very specific situations.
While general `command` always used.

TSC21:
> Definitely nope. Catching ACK needed only in very specific situations.
While general `command` always used.

And you think it makes sense to have a single plugin to catch ACK?

vooon:
Why not? What i'm not sure about, it is needed to be in upstream, or Pavlo just use it in their environment.
Anyway it is not necessary to place plugin into main or extras package.

TSC21:
> Anyway it is not necessary to place plugin into main or extras package.

Agreed. Though this can be here as a reference.

pavloblindnology:
It is blacklisted by default.
So, if not in main and extras package then where should it be?

TSC21:
> It is blacklisted by default.
So, if not in main and extras package then where should it be?
Should I delete this PR?

Again my consideration is that this should be included on the `command` plugin, as this is directly related to a command response and nothing more. If one does not want to parse the acks, we can just add a condition parameter on the `command` plugin to get the msgs but publish them only bases if that condition is set or not. Having a new plugin just for this is an overkill.

pavloblindnology:
@TSC21 I totally agree. `ACK` is caught by the `command` anyway. The only matter is publish it to ROS or not.

vooon:
I'm disagree. It is very specific, and should not be passed by default.
`command` do request, wait ack, and then return ack value in response. That's it.

If anyone want some special converter, they can just write special plugin.
Plugins may be placed in **any** ros package, so user can just do specific package for plugins needed for their robot.

@pavloblindnology you may just place this plugin in repo with bringup & others robot specific packages.
Then test how it works. Maybe later you may find that ACK's is not needed.

pavloblindnology:
@vooon Ok, I got you. Should I remove this PR?

pavloblindnology:
@vooon Can you please direct me to some example of mavros plugin implemented outside of mavros pakage? Thank you.

vooon:
`mavros_extras` package may be used as an example.

You need:
- `CmakeLists.txt` - you only need part that makes shared library
- `package.xml` - need `<mavros />` in `<export>`
- `mavros_plugins.xml`
- `src/plugins/plugin_name.cpp`

Also read that: http://wiki.ros.org/ROS/Tutorials/catkin/CreatingPackage

pavloblindnology:
Thank you @vooon. Seems I've already figured out it myself.
I'm closing this PR.

PR #910, opened 2017-12-27
extras: obstacle_distance plugin

TSC21:


TSC21:
@mrivi FYI. Please give a test and let me know how it goes.

mrivi:
@TSC21 I am still on holidays, I will test it as soon as I am back.

TSC21:
@mrivi what's the current testing state of this?

TSC21:
@mrivi can you please report here your difficulties on getting the msg published to the FCU? Did you check if the plugin is loaded at the beginning?

mrivi:
The problem was that mavros was sending a mavlink v1 packet. When the packet got to the parser in the mavlink_receiver it was dropped because the obstacle_distance message is present only in v2.

This PR #920 fixes the issue.

TSC21:
OK good to know it is working now. Merging!

PR #911, opened 2017-12-28
mavros_extras fake_gps:add vision topic param to use other pose msg.

ShingoMatsuura:
I am using fake_gps plugin with hector_SLAM. I had to change subscribe topic name for vision source so I added new vision param of subscribing topic name.


TSC21:
http://wiki.ros.org/roslaunch/XML/remap

ShingoMatsuura:
thanks!

PR #919, opened 2018-01-13
manual_control: add topic for sending MANUAL_CONTROL messages

okalachev:
PX4 supports "overriding" manual control inputs via MANUAL_CONTROL message: https://github.com/PX4/Firmware/blob/master/src/modules/mavlink/mavlink_receiver.cpp#L1559 (did not check for APM).

I added `~manual_control/override` topic (not sure about the name, this is the analogy with `rc/override`), allowing user to control the copter with a "virtual" RC with normalised values. For example:

```
rostopic pub -r 5 /mavros/manual_control/override mavros_msgs/ManualControl "header:
  seq: 0
  stamp: {secs: 0, nsecs: 0}
  frame_id: ''
x: 0.0
y: 0.0
z: 1000.0
r: 200.0
buttons: 0" 
```

in `POSCTL` mode makes the copter fly up and rotate (tested in SITL).

Also, I think `1` is the correct value for `queue_size` for such topics (`rc/override`, `manual_control/override`, etc), because we want to send relevant messages, we don't want to queue them.

### DIFF
--- a/mavros/src/plugins/manual_control.cpp
+++ b/mavros/src/plugins/manual_control.cpp
@@ -34,6 +34,7 @@ public:
 		PluginBase::initialize(uas_);
 
 		control_pub = manual_control_nh.advertise<mavros_msgs::ManualControl>("control", 10);
+		send_sub = manual_control_nh.subscribe("send", 1, &ManualControlPlugin::send_cb, this);
 	}
 
 	Subscriptions get_subscriptions() {
@@ -46,6 +47,7 @@ private:
 	ros::NodeHandle manual_control_nh;
 
 	ros::Publisher control_pub;
+	ros::Subscriber send_sub;
 
 	/* -*- rx handlers -*- */
 
@@ -61,6 +63,22 @@ private:
 		manual_control_msg->buttons = manual_control.buttons;
 
 		control_pub.publish(manual_control_msg);
+	}
+
+	/* -*- callbacks -*- */
+
+	void send_cb(const mavros_msgs::ManualControl::ConstPtr req)
+	{
+		mavlink::common::msg::MANUAL_CONTROL msg;
+		msg.target = m_uas->get_tgt_system();
+
+		msg.x = req->x;
+		msg.y = req->y;
+		msg.z = req->z;
+		msg.r = req->r;
+		msg.buttons = req->buttons;
+
+		UAS_FCU(m_uas)->send_message_ignore_drop(msg);
 	}
 };
 }	// namespace std_plugins


mavros/src/plugins/manual_control.cpp
```diff
@@ -34,6 +34,7 @@ class ManualControlPlugin : public plugin::PluginBase {
 		PluginBase::initialize(uas_);
 
 		control_pub = manual_control_nh.advertise<mavros_msgs::ManualControl>("control", 10);
+		control_sub = manual_control_nh.subscribe("override", 1, &ManualControlPlugin::override_cb, this);
```
vooon: Should be called `override_sub` in case topic name `override`. Maybe name it `send`?

okalachev: Yes, `send` is better. But, in such case, there'll be two topics, `send` and `control`, it's non-obvious, that these are send and receive, actually.

okalachev:
I renamed `override` to `send`. I think, the best naming is something like `send` and `receive`, but this will break up the compatibility, which is not good.

mchen0037:
@okalachev Could you give a brief explanation of how this works? When I use the example you give 
```
rostopic pub -r 5 /mavros/manual_control/override mavros_msgs/ManualControl "header:
  seq: 0
  stamp: {secs: 0, nsecs: 0}
  frame_id: ''
x: 0.0
y: 0.0
z: 1000.0
r: 200.0
buttons: 0" 
```
My copter flies straight up while spinning about yaw , but if I pass 0 in for z, it just sits there.

Is there a way to pass in normalized RC values? 

Thanks!

okalachev:
First, the topic was renamed to `send` (`/mavros/manual_control/send`). So it should not work at all :)

mchen0037:
Sorry--I just copy-pasted. I'm publishing it to /mavros/manual_control/send

okalachev:
In PX4, `x` and `y` are the pitch and roll sticks, normalized from -1000 to 1000.
`z` is a throttle stick, normilized from 0 to 1000 (unlike in MAVLink documentation).
`r` is yaw stick, normilized from -1000 to 1000.

Don't you have it working this way?

mchen0037:
Okay, thanks! I was sending in yaw values into z and throttle into r.

romaster93:
@okalachev 
Is there any further explanation possible?

I modified the command to 'send'. It is output normally in rostopic ehco manual_control / send. However, the FCU is not responding.

okalachev:
Sorry, I don't get your question. If publish values to this topic than the vehicle should behave as it was a real RC.

romaster93:
ok. 
I will attach a picture

romaster93:
@okalachev 

my env : 
px4 firmware
gazebo 9
ubuntu 18.04
ros melodic

i tyed this :

roslaunch mavros px4.launch

make px4_sitl_default gazebo

./QGroundControl.AppImage

rosrun test mavkbteleop.py ( https://github.com/romaster93/px4_test/blob/master/mavkbteleop.py )

and All work.


I can takeoff using QGroundControl. 

and i checked ' rostopic echo /mavros/rc/out ' :

header :

         seq: 225

         stamp:

             secs: 1587478017

             nsecs: 334894783

         frame_id : ' '

channels : [900, 900, 900, 900, 900, 900, 0 ,0, 0 ,0 ,0 ,0 ,0 ,0 ,0]

 

check 'rostopic echo /mavros/rc/in : 

header: 
  seq: 194
  stamp: 
    secs: 1587478017
    nsecs: 291170812
  frame_id: ''
rssi: 100
channels: [1500, 1500, 1500, 1500, 1000, 1000, 1000, 1000, 61566, 21897, 0, 46240, 43647, 32532, 0, 15184, 37888, 32532]

check 'rostopic echo /mavros/rc/override' :

channels: [1500, 1500, 1500, 1500, 1000, 1000, 1000, 1000]


first, I believed that the drone would respond if I modified the override value.
but, the drone did not move. So while I was looking for a solution, I saw an article In google that responded by tying terminal the ' rosrun mavros mavparm set SYSID_MYGCS 1 ', but it didn't work for me. In px4, SYSID_MYGCS did not exist.

After, I saw your article and tried it. I typed in another terminal.

rostopic pub -r 5 /mavros/manual_control/send mavros_msgs/ManualControl "header:
  seq: 0
  stamp: {secs: 0, nsecs: 0}
  frame_id: ''
x: 0.0
y: 0.0
z: 1000.0
r: 200.0
buttons: 0" 

I entered in terminal to confirm that the message was delivered.

rostopic echo mavros/manual_control/send :

header: 
  seq: 1675
  stamp: 
    secs: 0
    nsecs:         0
  frame_id: ''
x: 0.0
y: 0.0
z: 1000.0
r: 200.0
buttons: 0
---

But my drone(in gazebo) did not move on the ground.

I definitely want to launch the drone through the rc value in gazebo. I want to know what my problem is.

thanks....

okalachev:
Basically you have to arm your drone before it will handle virtual RC commands. Maybe this is the problem?

Also, if you want you drone to fly autonomously, it's generally better to utilize the OFFBOARD mode and setpoint commands.

Issue #822, opened 2017-09-24
Segfault on startup on ARM


JMare:
### Issue details

Segfault on startup causing mavros_node to die.
Problem is only seen on Raspberry Pi 3, doesn't happen on laptop.

Problem is intermittent, happens approximately 70% of the time when I launch mavros.

I have built with `catkin build --cmake-args -DCMAKE_BUILD_TYPE=Debug` and run with GDB, but the segfault backtrace only shows that the problem occurs in `ld-linux-armhf.so.3` and doesnt show the trace into mavros source.

### MAVROS version and platform

Mavros: 0.21.1 Built from source
ROS: Kinetic
Ubuntu: 16.04

### Autopilot type and version

[ x  ] ArduPilot
[  ] PX4

Version: 3.5

### Node logs

```
avionics@avionics:~$ roslaunch mavros apm2.launch fcu_url:=udp://localhost:14560@
... logging to /home/avionics/.ros/log/02c0d100-a0eb-11e7-b5ef-b827eb529962/roslaunch-avionics-6866.log
Checking log directory for disk usage. This may take awhile.
Press Ctrl-C to interrupt
Done checking log file disk usage. Usage is <1GB.

started roslaunch server http://avionics:44969/

SUMMARY
========

CLEAR PARAMETERS
 * /mavros/

PARAMETERS
 * /mavros/cmd/use_comp_id_system_control: False
 * /mavros/conn/heartbeat_rate: 1.0
 * /mavros/conn/system_time_rate: 0.0
 * /mavros/conn/timeout: 10.0
 * /mavros/conn/timesync_rate: 0.0
 * /mavros/distance_sensor/rangefinder_pub/field_of_view: 0.0
 * /mavros/distance_sensor/rangefinder_pub/frame_id: lidar
 * /mavros/distance_sensor/rangefinder_pub/id: 0
 * /mavros/distance_sensor/rangefinder_pub/send_tf: False
 * /mavros/distance_sensor/rangefinder_pub/sensor_position/x: 0.0
 * /mavros/distance_sensor/rangefinder_pub/sensor_position/y: 0.0
 * /mavros/distance_sensor/rangefinder_pub/sensor_position/z: -0.1
 * /mavros/distance_sensor/rangefinder_sub/id: 1
 * /mavros/distance_sensor/rangefinder_sub/orientation: PITCH_270
 * /mavros/distance_sensor/rangefinder_sub/subscriber: True
 * /mavros/fake_gps/eph: 2.0
[...]

PR #924, opened 2018-01-22
Blacklist HIL for APM since it is not relevent

JMare:
Ardupilot doesn't support hardware in the loop so this plugin is not needed.

Additionally, this prevents, but doesn't fix, the issue in #822


khancyr:
LGTM!

PR #925, opened 2018-01-23
Serial: better exit

khancyr:
Hello,

This should prevent the "Error: serial0: receive: End of file" problem !

### DIFF
--- a/libmavconn/src/serial.cpp
+++ b/libmavconn/src/serial.cpp
@@ -82,11 +82,15 @@ void MAVConnSerial::close()
 	if (!is_open())
 		return;
 
-	io_service.stop();
+	serial_dev.cancel();
 	serial_dev.close();
+
+	io_service.stop();
 
 	if (io_thread.joinable())
 		io_thread.join();
+
+	io_service.reset();
 
 	if (port_closed_cb)
 		port_closed_cb();
--- a/libmavconn/src/serial.cpp
+++ b/libmavconn/src/serial.cpp
@@ -54,6 +54,14 @@ MAVConnSerial::MAVConnSerial(uint8_t system_id, uint8_t component_id,
 		serial_dev.set_option(SPB::parity(SPB::parity::none));
 		serial_dev.set_option(SPB::stop_bits(SPB::stop_bits::one));
 		serial_dev.set_option(SPB::flow_control( (hwflow) ? SPB::flow_control::hardware : SPB::flow_control::none));
+#ifdef __linux__
+		// Set serial port to "raw" mode. Prevent the EOF exit
+		int fd = serial_dev.native_handle();
+		termios tio;
+		tcgetattr(fd, &tio);
+		cfmakeraw(&tio);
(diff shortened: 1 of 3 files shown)


libmavconn/src/serial.cpp
```diff
@@ -54,6 +54,15 @@ MAVConnSerial::MAVConnSerial(uint8_t system_id, uint8_t component_id,
...
 		serial_dev.set_option(SPB::flow_control( (hwflow) ? SPB::flow_control::hardware : SPB::flow_control::none));
+		#ifdef __linux__
+		int fd = serial_dev.native_handle();
+		struct termios tio;
+		tcgetattr(fd, &tio);
+		tio.c_cc[VTIME] = 0;
+		tio.c_cc[VMIN] = 1;
+		tcflush(fd, TCIFLUSH);
+		tcsetattr(fd, TCSANOW, &tio);
+		#endif
```
vooon: It is possible to do that via ASIO API? Just `set_option()` looks a bit better (but it also does tty calls).

Ifdef should not have any indentation. And maybe enclosure that hack to block?
Also it definitely should have comment on what problem solves that hack.

TSC21:
@khancyr any updates here?

khancyr:
I will do it before the end of the week, tcp and udp would gains from this too

khancyr:
@vooon set_option() doesn't allow to set those option. I have change to set to the serial port to raw mode that should prevent the EOF error.
The other commit is just to ensure that everything is close cleanly.

libmavconn/src/serial.cpp
```diff
@@ -82,12 +90,16 @@ void MAVConnSerial::close()
...
-	io_service.stop();
+	serial_dev.cancel();
 	serial_dev.close();
 
+	io_service.stop();
+
 	if (io_thread.joinable())
 		io_thread.join();
 
+	io_service.reset();
```
vooon: That is not necessary, destructor should free automatically.

libmavconn/src/tcp.cpp
```diff
@@ -131,14 +131,18 @@ void MAVConnTCPClient::close()
 	lock_guard lock(mutex);
 	if (!is_open())
 		return;
+	socket.shutdown(boost::asio::ip::tcp::socket::shutdown_send);
```
vooon: Blank line before lost.

vooon:
Thank you! I think i do release with that fix.

libmavconn/src/serial.cpp
khancyr: Normally yes, but that way we ensure that it is really done ! Using serial on asio should also put the port on raw mode but isn't done properly ;-P see https://github.com/chriskohlhoff/asio/commit/ff9d69e2fc002f62828d1c23b5eaae993427bc07

mhkabir:
@vooon Please hold for release. I have another hotfix for flow control incoming. This is fixed in Boost >= 1.66, but we cannot have that as default version yet till ROS supports it.

Issue #930, opened 2018-01-26
Silence "Table has different entry" console messages


lamping7:
### Issue details

Is there any plan to remove or silence the numerous "Table has different entry" console messages?

https://github.com/mavlink/mavros/blob/7426a6a75ceea449824b3a9c681ea8b955d3ae45/libmavconn/src/mavlink_helpers.cpp.em#L39

Change to `logDebug`? #931 

### MAVROS version and platform

Mavros: 0.22.0
ROS: Kinetic
Ubuntu: 16.04

### Autopilot type and version

[  ] ArduPilot
[ X ] PX4

Version: 1.7.3

### Node logs

```
[ WARN] [1516820172.507078551]: init: message from ASLUAV, MSG-ID 201 ignored! Table has different entry.
[ WARN] [1516820172.507314852]: init: message from autoquad, MSG-ID 150 ignored! Table has different entry.
[ WARN] [1516820172.507328406]: init: message from autoquad, MSG-ID 152 ignored! Table has different entry.
[ WARN] [1516820172.507539650]: init: message from matrixpilot, MSG-ID 150 ignored! Table has different entry.
[ WARN] [1516820172.507551888]: init: message from matrixpilot, MSG-ID 151 ignored! Table has different entry.
[ WARN] [1516820172.507563988]: init: message from matrixpilot, MSG-ID 152 ignored! Table has different entry.
[ WARN] [1516820172.507575543]: init: message from matrixpilot, MSG-ID 153 ignored! Table has different entry.
[ WARN] [1516820172.507591210]: init: message from matrixpilot, MSG-ID 155 ignored! Table has different entry.
[ WARN] [1516820172.507602556]: init: message from matrixpilot, MSG-ID 156 ignored! Table has different entry.
[ WARN] [1516820172.507614265]: init: message from matrixpilot, MSG-ID 157 ignored! Table has different entry.
[ WARN] [1516820172.507625593]: init: message from matrixpilot, MSG-ID 158 ignored! Table has different entry.
[ WARN] [1516820172.507636917]: init: message from matrixpilot, MSG-ID 170 ignored! Table has different entry.
[ WARN] [1516820172.507648235]: init: message from matrixpilot, MSG-ID 171 ignored! Table has different entry.
[ WARN] [1516820172.507661663]: init: message from matrixpilot, MSG-ID 172 ignored! Table has different entry.
[...]

PR #931, opened 2018-01-26
libmavconn: warn->debug table entry message

lamping7:
#930


vooon:
Actually i have idea to add parameter to select one dialect, not merge all as it done now.
But that solution is fine.

PR #949, opened 2018-02-19
obstacle_distance: measurament units 

mrivi:
`min_distance` and `max_distance` are `uint16_t` in the obstacle_distance mavlink message. It makes more sense to keep them as units and transform them into cm where the message gets used. 
I'll update also the mavlink message description. @TSC21 what do you think? 

FYI @ChristophTobler


TSC21:
Works for me. Please update Mavlink side so I can merge it. Also doesn't it sound better `one unit per cm`?

mrivi:
@TSC21 done :)

vooon:
Nope! You violate [LaserScan.msg][1] docs.

[1]: http://docs.ros.org/api/sensor_msgs/html/msg/LaserScan.html

vooon:
Hmm, `LaserScan.ranges` also should be meters. Bug!

TSC21:
@vooon I get your point about the docs, but what bug are you refering to?
@mrivi, @vooon is right, though we can keep the logic by dividing by 10 and rounding the values of the distances so you have units per cm.

PR #953, opened 2018-02-20
compile also with boost >= 1.66.0

bulwahn:
In boost 1.66.0, which includes boost-asio 1.12.0, the asio interfaces have been changed to follow the "C++ Extensions for Networking" Technical Specification [1]. As a consequence, resolvers now produce ranges rather than iterators.

This commits retains the original behaviour selecting the begin iterator from the provided range when using boost >= 1.66.0.

A preprocessing directive ensures backwards compatibility using the provided boost-asio version [2].

The issue was identified in a build with the cross-compilation tool chain provided in the meta-ros OpenEmbedded layer [3].

[1] http://www.boost.org/doc/libs/1_66_0/doc/html/boost_asio/net_ts.html
[2] https://github.com/boostorg/asio/commit/0c9cbdfbf217146c096265b5eb56089e8cebe608
[3] http://github.com/bmwcarit/meta-ros

### DIFF
--- a/libmavconn/src/tcp.cpp
+++ b/libmavconn/src/tcp.cpp
@@ -41,13 +41,19 @@ static bool resolve_address_tcp(io_service &io, size_t chan, std::string host, u
 	error_code ec;
 
 	tcp::resolver::query query(host, "");
-	std::for_each(resolver.resolve(query, ec), tcp::resolver::iterator(),
-			[&](const tcp::endpoint & q_ep) {
-				ep = q_ep;
-				ep.port(port);
-				result = true;
-				logDebug(PFXd "host %s resolved as %s", chan, host.c_str(), to_string_ss(ep).c_str());
-			});
+
+	auto fn = [&](const tcp::endpoint & q_ep) {
+		ep = q_ep;
+		ep.port(port);
+		result = true;
+		logDebug(PFXd "host %s resolved as %s", chan, host.c_str(), to_string_ss(ep).c_str());
+	};
+
+#if BOOST_ASIO_VERSION >= 101200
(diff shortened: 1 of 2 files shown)


vooon:
Does it support `for (auto v : range)`? I don't remember why i use `std::for_each()`.

bulwahn:
Good point. I only wanted to minimally change the code, but I will test using `for (auto v : range)` for boost 1.65.0 and boost 1.66.0 and come back to you on this pull request. This might take a day or two though.

bulwahn:
Hmm... so `for (auto v : range)` works for boost 1.66.0; but I could not figure out how to get that working with boost < 1.66.0. I am not really a C++ expert and I am still puzzled if there is an easy way to turn the iterator into a proper range or how to do that correctly.

bulwahn:
Are the build error in the travis CI related to this patch? From the log, I cannot really judge if it related to the commit, which should be perfectly backwards compatible; hence, I am a bit surprised.

TSC21:
Lukas do not worry about Travis. Our travis CI needs an update.

vooon:
Ah, ok. I think we may do that:

```c++

auto fn = [&](tcp::endpoint q_ep) {
...
};

#if ASIO > version
for (auto q_ep : resolver.resolve()) fn(q_ep);
#else
std::for_each(...);
#endif
```

// G++ should be smart enough to optimeze-out fn().

bulwahn:
Yes, that could actually work. I will try out that tomorrow morning (CET) and if it does work, provide an update of the pull request here.

bulwahn:
I followed your suggestion, compile-tested it on boost 1.65.0 and boost 1.66.0 and now it should be ready to be merged.

PR #960, opened 2018-03-01
launch: add optional respawn_mavros arg

lamping7:
I can see cases were we don't want our system to kill all the nodes if the mavros node dies. Instead, allow the user to specify a new `respawn_mavros` argument if they want the mavros node to respawn rather than kill the whole roslaunch.

The argument name was chosen based on gazebo's launch file (see [empty_world.launch](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/9cbc4bfd9b0caf763387cb203f6bf695071cae07/gazebo_ros/launch/empty_world.launch#L17)), so this should feel familiar to users.


vooon:
LGTM. Thanks!

PR #966, opened 2018-03-06
setpoint_attitude: added support for roll, pitch, yawrate and thrust setpoints.

Pedro-Roque:


TSC21:
Jenkins test this please.

TSC21:
@vooon any idea why Jenkins would fail to get `mav_msgs` as a dependency? It may actually be installed by `rosdep` and `apt-get`

mavros/launch/px4_config.yaml
```diff
@@ -106,6 +106,12 @@ setpoint_position:
     rate_limit: 50.0
   mav_frame: LOCAL_NED
 
+# setpoint_raw
+setpoint_raw:
```
vooon: Not raw, it is attitude.

mavros/src/plugins/setpoint_attitude.cpp
```diff
@@ -26,7 +27,7 @@
 #include <geometry_msgs/PoseStamped.h>
 #include <geometry_msgs/TwistStamped.h>
 #include <mavros_msgs/Thrust.h>
-
+#include <mav_msgs/RollPitchYawrateThrust.h>
```
vooon: Blank line missed. Do you run uncrustify?

mavros/src/plugins/setpoint_attitude.cpp
```diff
@@ -93,6 +94,23 @@ class SetpointAttitudePlugin : public plugin::PluginBase,
 			sync_twist.reset(new SyncTwistThrust(SyncTwistThrustPolicy(10), twist_sub, th_sub));
 			sync_twist->registerCallback(boost::bind(&SetpointAttitudePlugin::attitude_twist_cb, this, _1, _2));
 		}
+
+		rpyt_sub = sp_nh.subscribe("roll_pitch_yawrate_thrust", 10, &SetpointAttitudePlugin::rpyt_cb, this);
+		if(!sp_nh.getParam("thrust_scaling_factor", thrust_scaling)){
+			ROS_FATAL("No thrust scaling factor found, DO NOT FLY");
```
vooon: This only fatal for RPYT handler.

mavros/src/plugins/setpoint_attitude.cpp
```diff
@@ -215,6 +237,43 @@ class SetpointAttitudePlugin : public plugin::PluginBase,
...
+	 	        "the most likely cause of this is a failure to specify the thrust_scaling_factor,	"
+			"yaw_rate_scaling_factor or system_mass_kg parameters");
+		     	return;
+		}
+
+		// Set mask to ignore everything but thrust and attitude setpoint
+		uint8_t ignore_all_except_q_and_thrust = (7 << 0);
+
+		geometry_msgs::Quaternion orientation = 
+			tf::createQuaternionMsgFromRollPitchYaw( msg->roll, msg->pitch, 0);
```
vooon: We have `ftf::quaternion_from_rpy()` in `frame_tf.h`.

vooon: Why Yaw is ignored?

mavros/src/plugins/setpoint_attitude.cpp
```diff
@@ -215,6 +237,43 @@ class SetpointAttitudePlugin : public plugin::PluginBase,
...
+		tf::quaternionMsgToEigen(orientation, desired_orientation);
+		
+		// Transform desired orientation to represent aircraft->NED,
+		// MAVROS operates on orientation of base_link->ENU
+		auto ned_desired_orientation = ftf::transform_orientation_enu_ned(
+				ftf::transform_orientation_baselink_aircraft(desired_orientation));
+		
+		body_rate.x() = 0;
+		body_rate.y() = 0;
+		body_rate.z() = yaw_rate_scaling * ftf::detail::transform_frame_yaw(msg->yaw_rate);
```
vooon: Why not:
```cpp
auto body_rate = Eigen::Vector(0, 0, yaw...);
```

mavros/src/plugins/setpoint_attitude.cpp
Pedro-Roque: The idea is that we can apply a body rate to yaw, as usually the yaw reference can be different than the one onboard (for intance, if we use an external controller and just need to send an actuation to the UAV), while the roll/pitch are the same.

mavros/src/plugins/setpoint_attitude.cpp
Pedro-Roque: Will fix

mavros/src/plugins/setpoint_attitude.cpp
Pedro-Roque: In fact I didn't, I will run before the next commit.

mavros/launch/px4_config.yaml
Pedro-Roque: Yes, I forgot to move this when I changed it to the attitude.

mavros/src/plugins/setpoint_attitude.cpp
Pedro-Roque: Without the scaling, you mean?

mavros/src/plugins/setpoint_attitude.cpp
```diff
@@ -215,6 +237,43 @@ class SetpointAttitudePlugin : public plugin::PluginBase,
...
+			"yaw_rate_scaling_factor or system_mass_kg parameters");
+		     	return;
+		}
+
+		// Set mask to ignore everything but thrust and attitude setpoint
+		uint8_t ignore_all_except_q_and_thrust = (7 << 0);
+
+		geometry_msgs::Quaternion orientation = 
+			tf::createQuaternionMsgFromRollPitchYaw( msg->roll, msg->pitch, 0);
+		// Transforms from thrust acceleration to a thrust force (scaling adjusts to the UAV
```
mhkabir: This is incorrect. The thrust in the message, according to the definition in mav_msgs, is in Newtons (i.e a force) already. Can you also describe here how you compute the scaling factor in order to get a thrust value normalized from 0 to 1?

Pedro-Roque: I'm trying to figure out the best way to apply that scaling, from newtons [0,1]. On the UAV, 1 is the maximum thrust it can generate right? If so, a previously calibration needs to be set, and thrust_scaling set.  

The scaling can be done by putting the drone on a scale. Let's say the drone weights 2kg [thrust = 0] :
F=ma, so F=2*9.81= 19,62

When it lifts one kg [say at thrust.z = 0.1], then we know that it applied 9.81N, At maximum, it can do 98.1N of Force. 

So, 98.1N ----> thrust. = 1.0
         req N  ----->  x , x = req*1.0/98.1 , and so the scaling factor is  1/98.1 =  0.01019 . 

So, finally, we can set thrust = std::max(0.0, msg->thrust.z * thrust_scaling) , on which thrust_scaling is given by the user.

mavros/src/plugins/setpoint_attitude.cpp
vooon: Initializing, it looks a bit better than later fill zeros. Other possible code:

```
auto body_rate = Eigen::Vector3d::Zero();

body_rate.z() = yaw...;
```

TSC21:
Guys if you check this closely, this ends up to do a similar thing as done by [`send_attitude_quaternion()`](https://github.com/mavlink/mavros/blob/master/mavros/src/plugins/setpoint_attitude.cpp#L156). The main difference here is that this depends on the data received from just on topic, while the `send_attitude_quaternion()` depends of two subscribers being synced.
I do believe that any type of scaling should be applied on the publisher side, and not in our/plugin side. Though I would keep exactly the same structure as the one in `send_attitude_quaternion()`. The main difference also relies on the `RollPitchYawrateThrust` msg spec, where Roll and Pitch are angles and Yaw is a rate, though the control structure applied makes sense - but again, the thrust value scale should be done on the publisher side, with a verification done by [`is_normalized()`](https://github.com/mavlink/mavros/blob/master/mavros/src/plugins/setpoint_attitude.cpp#L130). The docs about how to apply the scale should be referenced on the plugin itself as a comment (note).

TSC21:
And also, as @vooon said, since the yaw rate is also sent, it should not be ignored, though the applied bitmask has to be changed to `ignore_all_except_yaw_q_and_thrust`.

mavros/src/plugins/setpoint_attitude.cpp
mhkabir: Okay, all this makes sense. We keep the thrust scaling in the plugin.

> this implies that the thrust is linearly increasing

I think this is a fair assumption. We can handle linearisation on the autopilot side.

Pedro-Roque: Thanks @mhkabir , I'll update this later in the day.

Pedro-Roque:
@vooon @mhkabir all set now?

mavros/package.xml
```diff
@@ -51,11 +51,13 @@
   <depend>diagnostic_msgs</depend>
   <depend>geometry_msgs</depend>
   <depend>mavros_msgs</depend>
+  <depend>mav_msgs</depend>
   <depend>nav_msgs</depend>
   <depend>sensor_msgs</depend>
   <depend>geographic_msgs</depend>
   <depend>std_msgs</depend>
   <depend>std_srvs</depend>
+  <depend>mav_msgs</depend>
```
vooon: Why 2 times?

mavros/launch/px4_config.yaml
vooon: I suppose that whole entry for sp-raw not needed.

mavros/launch/px4_config.yaml
```diff
@@ -96,6 +96,7 @@ setpoint_attitude:
     frame_id: "map"
     child_frame_id: "target_attitude"
     rate_limit: 50.0
+  thrust_scaling: 0.0
```
vooon: Better move it before tf.

mavros/src/plugins/setpoint_attitude.cpp
```diff
@@ -215,6 +222,56 @@ class SetpointAttitudePlugin : public plugin::PluginBase,
 			send_attitude_ang_velocity(req->header.stamp, ang_vel, thrust_msg->thrust);
 	}
 
+	/**
+	 * @brief Setpoint for Roll, Pitch, Yawrate and Thrust (RPYT).
+	 * Message specification: http://docs.ros.org/indigo/api/mav_msgs/html/msg/RollPitchYawrateThrust.html
+	 * Thrust is set in Newtons and scaled by 1/MAX_THRUST_N (MAX_THRUST_N is UAV dependent).
+	 * @param msg		Received RPYT setpoint
+	 */
+	void rpyt_cb(const mav_msgs::RollPitchYawrateThrustConstPtr msg)
```
vooon: To keep consistency with other sub-cb's:
```cpp
void rpyt_cb(const mav_msgs::RollPitchYawrateThrust::ConstPtr &msg)
```

mavros/src/plugins/setpoint_attitude.cpp
```diff
@@ -215,6 +222,56 @@ class SetpointAttitudePlugin : public plugin::PluginBase,
...
+		// Set mask to ignore everything but thrust and attitude setpoint
+		uint8_t ignore_all_except_q_and_thrust = 0; //(7 << 0);
+
+		geometry_msgs::Quaternion orientation = 
+			tf::createQuaternionMsgFromRollPitchYaw( msg->roll, msg->pitch, 0);
+		// Transforms from thrust acceleration to a thrust force (scaling adjusts to the UAV
+		// propeller thrust, can be calculated from UAV calibration
+		double thrust = std::min(1.0, std::max(0.0, msg->thrust.z * thrust_scaling));
+		Eigen::Quaterniond desired_orientation;
+		tf::quaternionMsgToEigen(orientation, desired_orientation);
```
vooon: I already said. Should be:
```cpp
auto desired_orientation = ftf::quaternion_from_rpy(msg->roll, msg->pitch, 0.0);
```

mavros/src/plugins/setpoint_raw.cpp
```diff
@@ -217,6 +218,7 @@ class SetpointRawPlugin : public plugin::PluginBase,
 				body_rate,
 				req->thrust);
 	}
+
```
vooon: This PR shouldn't touch that file.

AlexisTM:
@mhkabir @Pedro-Roque What about leaving the thrust scaling factor but defaulting it to 1?

Indeed, the `mav_control_rw` library depends on the copter kinematics and outputs the thrust in Newton. Therefore, leaving a thrust factor allows using that input normalised for the Pixhawk on Mavros side.

> command/roll_pitch_yawrate_thrust of type mav_msgs/RollPitchYawrateThrust. This is the command to the low level controller. Angles are in rad and thrust is in N.

mavros/package.xml
AlexisTM: This is the reason it fails to build on Jenkins:

```log
05:04:18 Traceback (most recent call last):
05:04:18   File "/tmp/ros_buildfarm/scripts/devel/create_devel_task_generator.py", line 250, in <module>
05:04:18     main()
05:04:18   File "/tmp/ros_buildfarm/scripts/devel/create_devel_task_generator.py", line 76, in main
05:04:18     pkgs.update(find_packages(source_space))
05:04:18   File "/usr/lib/python3/dist-packages/catkin_pkg/packages.py", line 86, in find_packages
05:04:18     packages = find_packages_allowing_duplicates(basepath, exclude_paths=exclude_paths, exclude_subspaces=exclude_subspaces, warnings=warnings)
05:04:18   File "/usr/lib/python3/dist-packages/catkin_pkg/packages.py", line 146, in find_packages_allowing_duplicates
05:04:18     xml, filename=filename, warnings=warnings)
05:04:18   File "/usr/lib/python3/dist-packages/catkin_pkg/package.py", line 662, in parse_package_string
05:04:18     raise InvalidPackage('Error(s) in %s:%s' % (filename, ''.join(['\n- %s' % e for e in errors])))
05:04:18 catkin_pkg.package.InvalidPackage: Error(s) in /tmp/catkin_workspace/src/mavros/mavros/package.xml:
05:04:18 - The generic dependency on 'mav_msgs' is redundant with: build_depend, build_export_depend, exec_depend
05:04:18 Build step 'Execute shell' marked build as failure
```

vooon: `<depend />` = `<{build,exec,...}_depend />`
And i'm sure that second depend is not solution.

mavros/launch/px4_config.yaml
vooon: And good idea to add comment that this parameter used only for RPYT.

Pedro-Roque: True! Will fix.

Pedro-Roque:
@AlexisTM a warning is sent if you use the plugin and the thrust_scaling is 0. The reason to default to zero is merely safety - in this way we make sure that the user actually knows what the thrust_scaling is for, and we prevent him of sending Newtons directly to the attitude_target.

AlexisTM:
@Pedro-Roque But if we remove the scaling factor and use only normalized thrust, there will be no prevention such problem. 

I would leave 1 so it is not bothering people developing with normalized thrust and send a message on plugin instantiation that it needs to be normalized and if not, set the scaling factor accordingly?

mavros/package.xml
AlexisTM: What I mean is: if we just put the `<depend>mav_msgs</depend>` just once it builds fine.

Pedro-Roque:
@AlexisTM , the documentation for RollPitchYawrateThrust is that the Thrust comes in Newtons, and that is the reference we should follow. As it needs conversion, I believe it should be done inside and the user should be aware of that.

Now, it is implemented as:
- user uses Newton? It get's a message saying the thrust scaling is zero, so it needs to be set and he needs to understand how to set it;
- user uses normalized thrust? It will get the message, and after reading what it is, he just needs to put it to one.

In your argument, I could argue the opposite: it bothers people that are using non-normalized thrust to not know that on the vehicle it is normalized, and if you set a thrust input in newtons, you would get the maximum thrust right away, which is not safe. In this case, I think we should follow the standard documentation, and the for the type of message being used, it is Newtons. As it is converted, the user should be aware of this.

mavros/package.xml
Pedro-Roque: Yep I realized this on a local build as well.

Pedro-Roque:
@vooon The jenkins still gives the same build problem as previously mentioned by @TSC21 - caused by keyerror on rosdep_refs.

vooon:
Fuck! [mav_msgs][1] not released for Lunar! I cannot accept that PR.

[1]: http://wiki.ros.org/mav_msgs

TSC21:
We need to ask for a release

mavros/src/plugins/setpoint_attitude.cpp
```diff
@@ -215,6 +222,53 @@ class SetpointAttitudePlugin : public plugin::PluginBase,
...
+		}
+
+		if(thrust_scaling == 0.0){
+			ROS_WARN("thrust_scaling parameter is set to zero.");
+		}
+
+		// Set mask to ignore everything but thrust and attitude setpoint
+		uint8_t ignore_all_except_q_and_thrust = 0; //(7 << 0);
+
+		// Transforms from thrust acceleration to a thrust force (scaling adjusts to the UAV
```
mhkabir: Comment is incorrect. It is already a force.

mhkabir: "Normalizes thrust force (in Newtons) to a range of 0-1" is more appropriate.

Pedro-Roque:
Holy :( Alright, I'll keep it here meanwhile.

mavros/src/plugins/setpoint_attitude.cpp
Pedro-Roque: True, I changed to zero just for trying, forget to put it back. Will do soon :)

mavros/src/plugins/setpoint_raw.cpp
```diff
@@ -17,6 +17,7 @@
 #include <mavros/mavros_plugin.h>
 #include <mavros/setpoint_mixin.h>
 #include <eigen_conversions/eigen_msg.h>
+#include <tf/transform_datatypes.h>
```
vooon: Not needed any more, not here nor setpoint_attitude.cpp.

TSC21:
FYI: https://github.com/ethz-asl/mav_comm/issues/51

Pedro-Roque:
@TSC21 @vooon @mhkabir  I removed the rpyt topic and instead fixed the attitude_cb on setpoint_raw/attitude , which already provides the functionality.

Moreover, changes have to be made to the PX4 firmware to support it. That will go as a PR to the PX4 stack.

mavros/src/plugins/setpoint_raw.cpp
```diff
@@ -217,25 +222,45 @@ class SetpointRawPlugin : public plugin::PluginBase,
...
 		// Transform desired orientation to represent aircraft->NED,
 		// MAVROS operates on orientation of base_link->ENU
 		auto ned_desired_orientation = ftf::transform_orientation_enu_ned(
 					ftf::transform_orientation_baselink_aircraft(desired_orientation));
 
-		auto body_rate = ftf::transform_frame_baselink_aircraft(baselink_angular_rate);
-
-		tf::vectorMsgToEigen(req->body_rate, body_rate);
+		// transform body-rate
+		tf::vectorMsgToEigen(req->body_rate, body_rate);
```
TSC21: Can you please run uncrustify here?

vooon: Better replace all `tf::xxxMsgToEigen(msg, eigen)` with `ftf::to_eigen(msg) -> eigen`

TSC21:
@Pedro-Roque the PR became a little bit messy, with redundant commits and not even being able to rebase given conflicts. Do you mind creating a new branch with a single commit where you add all the changes you have here? Also, please address https://github.com/mavlink/mavros/pull/966#discussion_r192245200. Thanks!

TSC21:
@Pedro-Roque FYI https://github.com/ethz-asl/mav_comm/issues/51

TSC21:
@Pedro-Roque can you please rebase and update this PR? The `mav_comm` package seems to be updated already with the required msgs in Melodic. Thanks

Pedro-Roque:
@TSC21 Yes, I'll come back to this later as I had to shift my attention to other things for the moment. By the end of the week I'll take a look at this!

soldierofhell:
Hi, anyone interested in finalizing this PR?

AlexisTM:
Feel free to fix the last comments, ensure functionality (2y old PR) and probably create a new one instead of using this one.

soldierofhell:
Ok, thanks for info

Pedro-Roque:
This is finalized @soldierofhell @AlexisTM , you can use it through the setpoint_raw/attitude and proper masks.

PR #971, opened 2018-03-08
libmavconn : fix hardware flow control setting for Boost < 1.66

mhkabir:
This commit fixes handling of hardware flow control. Due to bugs in Boost, `set_option()` would not work for flow control settings, and mavros would die if you tried a `serial-hwfc` link. 

This is fixed in Boost v1.66, but we need this workaround for now. Relevant Boost commit : https://github.com/boostorg/asio/commit/619cea4356

### DIFF
--- a/libmavconn/src/serial.cpp
+++ b/libmavconn/src/serial.cpp
@@ -21,6 +21,8 @@
 #include <mavconn/thread_utils.h>
 #include <mavconn/serial.h>
 
+#include <boost/version.hpp>
+
 namespace mavconn {
 using boost::system::error_code;
 using boost::asio::io_service;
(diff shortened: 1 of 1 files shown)


mhkabir:
@sharvashish this will enable you to use flow control.

khancyr:
LGTM but should be tested !

mhkabir:
Already tested with flow control enabled hardware. That is what motivated the development. :)

LorenzMeier (approved):

LorenzMeier:
Looking correct.

libmavconn/src/serial.cpp
```diff
@@ -53,15 +55,35 @@ MAVConnSerial::MAVConnSerial(uint8_t system_id, uint8_t component_id,
 		serial_dev.set_option(SPB::character_size(8));
 		serial_dev.set_option(SPB::parity(SPB::parity::none));
 		serial_dev.set_option(SPB::stop_bits(SPB::stop_bits::one));
-		serial_dev.set_option(SPB::flow_control( (hwflow) ? SPB::flow_control::hardware : SPB::flow_control::none));
+		// Flow control setting in older versions of Boost is broken, use workaround (below) for now.
+		if (BOOST_VERSION >= 106600) {
```
vooon: I think better to use preprocessor, like in #953.

libmavconn/src/serial.cpp
```diff
@@ -53,15 +55,35 @@ MAVConnSerial::MAVConnSerial(uint8_t system_id, uint8_t component_id,
...
+			if (hwflow) {
+				tio.c_iflag &= ~(IXOFF | IXON);
+				tio.c_cflag |= CRTSCTS;
+			} else {
+				tio.c_iflag &= ~(IXOFF | IXON);
+				tio.c_cflag &= ~CRTSCTS;
+			}
+
+			// Set serial port to "raw" mode to prevent EOF exit.
+			cfmakeraw(&tio);
```
vooon: Does it fixed in 1.66 (ASIO 1.12)?

vooon: Yes, fixed.

PR #972, opened 2018-03-10
Add ability to send STATUSTEXT messages

anassinator:
This adds the ability to listen to and send STATUSTEXT messages to the FCU. Sent STATUSTEXT can then be received by QGroundControl (and read out if the severity is high enough) automatically over telemetry as if they were sent by the FCU.

To make use of this, one needs to enable mavlink forwarding on the FCU. On PX4, that can be achieved with the `-f` command when running `mavlink start`.

### DIFF
--- a/mavros/src/plugins/sys_status.cpp
+++ b/mavros/src/plugins/sys_status.cpp
@@ -21,6 +21,7 @@
 #include <mavros_msgs/StreamRate.h>
 #include <mavros_msgs/SetMode.h>
 #include <mavros_msgs/CommandLong.h>
+#include <mavros_msgs/StatusText.h>
 
 #ifdef HAVE_SENSOR_MSGS_BATTERYSTATE_MSG
 #include <sensor_msgs/BatteryState.h>
@@ -466,6 +467,8 @@ public:
 		state_pub = nh.advertise<mavros_msgs::State>("state", 10, true);
 		extended_state_pub = nh.advertise<mavros_msgs::ExtendedState>("extended_state", 10);
 		batt_pub = nh.advertise<BatteryMsg>("battery", 10);
+		statustext_pub = nh.advertise<mavros_msgs::StatusText>("statustext/recv", 10);
+		statustext_sub = nh.subscribe("statustext/send", 10, &SystemStatusPlugin::statustext_cb, this);
 		rate_srv = nh.advertiseService("set_stream_rate", &SystemStatusPlugin::set_rate_cb, this);
 		mode_srv = nh.advertiseService("set_mode", &SystemStatusPlugin::set_mode_cb, this);
 
@@ -502,6 +505,8 @@ private:
 	ros::Publisher state_pub;
 	ros::Publisher extended_state_pub;
 	ros::Publisher batt_pub;
+	ros::Publisher statustext_pub;
+	ros::Subscriber statustext_sub;
 	ros::ServiceServer rate_srv;
 	ros::ServiceServer mode_srv;
 
@@ -710,8 +715,13 @@ private:
 	void handle_statustext(const mavlink::mavlink_message_t *msg, mavlink::common::msg::STATUSTEXT &textm)
 	{
 		auto text = mavlink::to_string(textm.text);
-
 		process_statustext_normal(textm.severity, text);
+
+		auto st_msg = boost::make_shared<mavros_msgs::StatusText>();
+		st_msg->header.stamp = ros::Time::now();
+		st_msg->severity = textm.severity;
+		st_msg->text = text;
+		statustext_pub.publish(st_msg);
 	}
 
 	void handle_meminfo(const mavlink::mavlink_message_t *msg, mavlink::ardupilotmega::msg::MEMINFO &mem)
@@ -901,6 +911,20 @@ private:
 			// publish connection change
 			publish_disconnection();
 		}
+	}
+
+	/* -*- subscription callbacks -*- */
+
+	void statustext_cb(const mavros_msgs::StatusText::ConstPtr &req) {
+		mavlink::common::msg::STATUSTEXT statustext {};
+		statustext.severity = req->severity;
+
+		// Limit the length of the string by null-terminating at the 50-th character
+		ROS_WARN_COND_NAMED(req->text.length() >= statustext.text.size(), "sys",
+				"Status text too long: truncating...");
+		mavlink::set_string_z(statustext.text, req->text);
+
+		UAS_FCU(m_uas)->send_message_ignore_drop(statustext);
 	}
 
 	/* -*- ros callbacks -*- */
--- a/mavros_msgs/msg/StatusText.msg
+++ b/mavros_msgs/msg/StatusText.msg
@@ -1,0 +1,17 @@
+# STATUSTEXT message representation
+# https://mavlink.io/en/messages/common.html#STATUSTEXT
+
+# Severity levels
+uint8 EMERGENCY = 0
+uint8 ALERT = 1
+uint8 CRITICAL = 2
+uint8 ERROR = 3
+uint8 WARNING = 4
+uint8 NOTICE = 5
+uint8 INFO = 6
+uint8 DEBUG = 7
+
+# Fields
+std_msgs/Header header
+uint8 severity
+string text
--- a/mavros_msgs/CMakeLists.txt
+++ b/mavros_msgs/CMakeLists.txt
@@ -35,6 +35,7 @@ add_message_files(
   RCOut.msg
   RadioStatus.msg
   State.msg
+  StatusText.msg
   Thrust.msg
   VFR_HUD.msg
   Vibration.msg


mhkabir:
Nice, thanks!

TSC21:
The question is: why? https://github.com/mavlink/mavros/blob/master/mavros/src/plugins/sys_status.cpp#L710 did you actually checked this?

anassinator:
Yeah, but that just outputs it to `/rosout` and loses the severity, which makes it hard to parse/debug. The main use of this is the ability to send STATUSTEXT messages to the FCU though; that allows the OBC to send status messages to the operator for QGC to read out. Receiving them was just the cherry on top.

TSC21:
So or you extend the sys_status plugin or or remove it from from it. Does not make sense two handlers for the same msg.

anassinator:
You mean I should I move all of this to `sys_status`? Or should I just remove one of the handlers?

TSC21:
Extend `sys_status` plugin but keep its current functionality there

anassinator:
Should be good now

TSC21:
Jenkins test this please.

anassinator:
Do I need to do something to fix the build?

mavros/src/plugins/sys_status.cpp
```diff
@@ -903,6 +912,21 @@ class SystemStatusPlugin : public plugin::PluginBase
 		}
 	}
 
+	/* -*- subscription callbacks -*- */
+
+	void statustext_cb(const mavros_msgs::StatusText::ConstPtr &req) {
+		mavlink::common::msg::STATUSTEXT statustext {};
+
+		if (req->text.length() >= mavlink::common::msg::STATUSTEXT::LENGTH) {
```
vooon: Wrong. Should be:
```cpp
req->text.lenght() >= statustext.text.size() - 1
```

mavros/src/plugins/sys_status.cpp
```diff
@@ -903,6 +912,21 @@ class SystemStatusPlugin : public plugin::PluginBase
...
+			ROS_WARN("Status text too long: truncating...");
+		}
+
+		statustext.severity = req->severity;
+		mavlink::set_string(statustext.text, req->text);
```
vooon:
```
mavlink::set_string_z()
```

mavros/src/plugins/sys_status.cpp
```diff
@@ -466,6 +467,8 @@ class SystemStatusPlugin : public plugin::PluginBase
 		state_pub = nh.advertise<mavros_msgs::State>("state", 10, true);
 		extended_state_pub = nh.advertise<mavros_msgs::ExtendedState>("extended_state", 10);
 		batt_pub = nh.advertise<BatteryMsg>("battery", 10);
+		statustext_pub = nh.advertise<mavros_msgs::StatusText>("statustext/get", 10);
+		statustext_sub = nh.subscribe("statustext/send", 10, &SystemStatusPlugin::statustext_cb, this);
```
vooon: send/recv.

mavros/src/plugins/sys_status.cpp
vooon: Also that should be in `ROS_WARN_NAMED_COND()`

mavros_msgs/msg/StatusText.msg
```diff
@@ -0,0 +1,15 @@
...
+
+uint8 severity
+uint8 SEVERITY_EMERGENCY = 0
+uint8 SEVERITY_ALERT = 1
+uint8 SEVERITY_CRITICAL = 2
+uint8 SEVERITY_ERROR = 3
+uint8 SEVERITY_WARNING = 4
+uint8 SEVERITY_NOTICE = 5
+uint8 SEVERITY_INFO = 6
+uint8 SEVERITY_DEBUG = 7
```
vooon: May be better without prefix.

vooon: Like in std rosout: http://docs.ros.org/api/rosgraph_msgs/html/msg/Log.html

vooon: And that would be better if you use Cog to generate enum from pymavlink.

mavros/src/plugins/sys_status.cpp
```diff
@@ -710,8 +715,12 @@ class SystemStatusPlugin : public plugin::PluginBase
 	void handle_statustext(const mavlink::mavlink_message_t *msg, mavlink::common::msg::STATUSTEXT &textm)
 	{
 		auto text = mavlink::to_string(textm.text);
-
 		process_statustext_normal(textm.severity, text);
+
+		auto st_msg = boost::make_shared<mavros_msgs::StatusText>();
+		st_msg->severity = textm.severity;
+		st_msg->text = mavlink::to_string(textm.text);
```
vooon: Why not to use `text` obj?

anassinator:
Thanks @voon. I think I addressed everything except for the cog part, I'll look into that later tonight.

mavros/src/plugins/sys_status.cpp
```diff
@@ -710,8 +715,12 @@ class SystemStatusPlugin : public plugin::PluginBase
 	void handle_statustext(const mavlink::mavlink_message_t *msg, mavlink::common::msg::STATUSTEXT &textm)
 	{
 		auto text = mavlink::to_string(textm.text);
-
 		process_statustext_normal(textm.severity, text);
+
+		auto st_msg = boost::make_shared<mavros_msgs::StatusText>();
+		st_msg->severity = textm.severity;
```
vooon: Add timestamp:

```
st_msg->header.stamp = ros::Time::now();
```

mavros_msgs/msg/StatusText.msg
```diff
@@ -0,0 +1,16 @@
+# Status text
```
vooon: A bit better header comment:
```
# STATUSTEXT message representation
# https://mavlink.io/en/messages/common.html#STATUSTEXT
```

mavros/src/plugins/sys_status.cpp
```diff
@@ -903,6 +912,19 @@ class SystemStatusPlugin : public plugin::PluginBase
...
 
+	/* -*- subscription callbacks -*- */
+
+	void statustext_cb(const mavros_msgs::StatusText::ConstPtr &req) {
+		mavlink::common::msg::STATUSTEXT statustext {};
+		ROS_WARN_COND_NAMED(req->text.length() >= statustext.text.size() - 1,
+				"sys", "Status text too long: truncating...");
+
+		statustext.severity = req->severity;
+		mavlink::set_string_z(statustext.text, req->text);
```
vooon: Here we can add comment, that message defined "without null termination",
but quite often software treat it as null terminated.
And since it is not so limited like param name, better to keep 50-th character for terminator.

mavros/src/plugins/sys_status.cpp
```diff
@@ -903,6 +912,19 @@ class SystemStatusPlugin : public plugin::PluginBase
 		}
 	}
 
+	/* -*- subscription callbacks -*- */
+
+	void statustext_cb(const mavros_msgs::StatusText::ConstPtr &req) {
+		mavlink::common::msg::STATUSTEXT statustext {};
+		ROS_WARN_COND_NAMED(req->text.length() >= statustext.text.size() - 1,
```
vooon: Hmm, i'm a bit miss. Strings with 50 or more symbols will be truncated to 49.
So `s/>=/>/` or `s/\ -\ 1//`.

anassinator: oops, you're right

mavros/src/plugins/sys_status.cpp
anassinator: not sure I understood this completely, but I added a comment

vooon:
LGTM. Thanks!

anassinator:
Thanks! Do you still want the Cog stuff?

vooon:
Not necessary, but it might be helpful in your development.

4ert:
Hello! I tested this feature and it works nice with px4 v1.7.0 in SITL, but it doesn't with px4 v1.7.3 in real Pixhawk.
Can someone help me with this?
Thanks!

anassinator:
You probably need to add the `-f` flag like here:

https://github.com/mcgill-robotics/Firmware/commit/b8dfb5a47ecaf56136a6a662e62e66a50b4fbba6#diff-e5c16c5209413f30e964883f74027b51R715

That change will soon make it to the latest `master`.

4ert:
I made a microsd/etc/config.txt with:
```
mavlink start -f -r 800000 -d /dev/ttyACM0 -m config -x
mavlink start -f -d /dev/ttyS2 -b 57600 -m osd -r 1000
```
And I can see in console with "ps" command that all my mavlink instances run with "-f" flag.
My FCU is quadrocopter with pixhawk connected to onboard raspberry pi by usb. QGroundControl connects to it by UDP through mavproxy.

anassinator:
I'm not sure if there is any additional setup required for mavproxy as I've never used that, but I am 100% sure it's a configuration issue if it works in SITL. Is the `send` topic not working or is the problem with `recv`?

4ert:
The problem is with `send` topic.

4ert:
And on what versions of pixhawk, firmware and qgroundcontrol did you test the ability of sending statustext?

anassinator:
We just flew with it successfully on our Pixhawk Mini running PX4 v1.7.3 and QGroundControl 3.2.7. I didn't have to do anything other than adding the `-f` flags so I don't know where the problem could be. I'd suggest trying it on a simpler setup and building it back up to what you have now until you find the point of failure.

4ert:
I have just tested with raspberry pi with mavros connected to pixhawk through microusb and QGroundControl connected to pixhawk through telem2 to usb - and it works! I have received messages in QGC which I sent via mavros.
So, I think, the problem is in the UDP connection.

TiffanyKalin:
Hi, does this work if I am using ArduPilot instead of PX4? I am publishing successfully but I'm not sure Ardupilot is receiving the statustexts.

jiejiezhang:
Hi！ I'm not sure why it's not work on sitl. My Firmware version is V1.11.0

PR #975, opened 2018-03-15
libmavconn : enable low-latency mode on Linux

mhkabir:
Some common USB-UART converters like the FTDI accumulate individual bytes from the serial link in order to send them in a single USB packet ("Nagling" - https://en.wikipedia.org/wiki/Nagle%27s_algorithm). This commit sets the ASYNC_LOW_LATENCY flag, which the FTDI kernel driver interprets as a request to drop the Nagling timer to 1ms (i.e send all accumulated bytes after 1ms.)

This reduces average link RTT to under 5ms at 921600 baud (it is 20-30 ms on master), and enables the use of mavros in systems where low latency is required to get good performance for e.g estimation and controls. As a side effect, this also greatly increases performance of timesync on serial links and prevents the hard-resets we've been seeing.

### DIFF
--- a/libmavconn/src/serial.cpp
+++ b/libmavconn/src/serial.cpp
@@ -47,6 +47,7 @@ MAVConnSerial::MAVConnSerial(uint8_t system_id, uint8_t component_id,
 
 	try {
 		serial_dev.open(device);
+		int fd = serial_dev.native_handle();
 
 		// Set baudrate and 8N1 mode
 		serial_dev.set_option(SPB::baud_rate(baudrate));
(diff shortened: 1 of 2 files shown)


mhkabir:
@TSC21 
> Good commit! Can you share some results?

> This reduces average link RTT to under 5ms at 921600 baud (it is 20-30 ms on master)

TSC21:
Yeah I read that. I was asking if you have some pratical example where you tested this (ex: VIO)

libmavconn/src/serial.cpp
```diff
@@ -61,7 +62,6 @@ MAVConnSerial::MAVConnSerial(uint8_t system_id, uint8_t component_id,
 		// Workaround to set some options for the port manually. This is done in
 		// Boost.ASIO, but until v1.12.0 (Boost 1.66) there was a bug which doesn't enable relevant
 		// code. Fixed by commit: https://github.com/boostorg/asio/commit/619cea4356
-		int fd = serial_dev.native_handle();
```
vooon: This is not portable outside of POSIX systems.

PR #983, opened 2018-03-28
[WIP] Trajectory Plugin

mrivi:
Plugin to receive a path from the FCU and send back the collision free path.
Works with: https://github.com/mavlink/mavlink/pull/856

FYI @ChristophTobler  @Stifael @TSC21


mavros_extras/src/plugins/obstacle_avoidance.cpp
```diff
@@ -0,0 +1,237 @@
...
+        obstacle_avoidance.point_5[1] = req->point_5.position.x;
+        obstacle_avoidance.point_5[2] = -req->point_5.position.z;
+        obstacle_avoidance.point_5[3] = req->point_5.velocity.y;
+        obstacle_avoidance.point_5[4] = req->point_5.velocity.x;
+        obstacle_avoidance.point_5[5] = -req->point_5.velocity.z;
+        obstacle_avoidance.point_5[6] = req->point_5.acceleration_or_force.y;
+        obstacle_avoidance.point_5[7] = req->point_5.acceleration_or_force.x;
+        obstacle_avoidance.point_5[8] = -req->point_5.acceleration_or_force.z;
+        obstacle_avoidance.point_5[9] = wrap_pi(-req->point_5.yaw + (M_PI / 2.0f));
+        obstacle_avoidance.point_5[10] = req->point_5.yaw_rate;
```
vooon: That should be generated with Cog. Too error-prone.

mavros_extras/src/plugins/obstacle_avoidance.cpp
```diff
@@ -0,0 +1,237 @@
...
+        obstacle_avoidance_input->point_5.position.y = avoid_input.point_5[0];
+        obstacle_avoidance_input->point_5.position.z = -avoid_input.point_5[2];
+        obstacle_avoidance_input->point_5.velocity.x = avoid_input.point_5[4];
+        obstacle_avoidance_input->point_5.velocity.y = avoid_input.point_5[3];
+        obstacle_avoidance_input->point_5.velocity.z = -avoid_input.point_5[5];
+        obstacle_avoidance_input->point_5.acceleration_or_force.x = avoid_input.point_5[7];
+        obstacle_avoidance_input->point_5.acceleration_or_force.y = avoid_input.point_5[6];
+        obstacle_avoidance_input->point_5.acceleration_or_force.z = -avoid_input.point_5[8];
+        obstacle_avoidance_input->point_5.yaw = wrap_pi((M_PI / 2.0f) - avoid_input.point_5[9]);
+        obstacle_avoidance_input->point_5.yaw_rate = avoid_input.point_5[10];
```
vooon: Cog.

mavros_msgs/msg/ObstacleAvoidance.msg
```diff
@@ -0,0 +1,18 @@
...
+
+uint8 type # See enum MAV_TRAJECTORY_REPRESENTATION.
+uint8 MAV_TRAJECTORY_REPRESENTATION_WAYPOINTS = 0
+uint8 MAV_TRAJECTORY_REPRESENTATION_BEZIER = 1
+
+mavros_msgs/PositionTarget point_1
+mavros_msgs/PositionTarget point_2
+mavros_msgs/PositionTarget point_3
+mavros_msgs/PositionTarget point_4
+mavros_msgs/PositionTarget point_5
```
vooon: Why not:
```
mavros_msgs/PositionTarget[5] points
```

mrivi:
@vooon  the message is not yet in `common.xml`, but I have put the corresponding mavlink PR in the description.
> (Works with: mavlink/mavlink#856

This is part of a larger effort on the obstacle avoidance interace. This is the proposal document https://docs.google.com/document/d/1BQp1a6yszl9f6LDrxrkKUDDGyXxBs5C86BzvJwVbRrU/edit?usp=sharing

mavros_msgs/msg/ObstacleAvoidance.msg
mrivi: I think is help with readability to have a bit more text and less indexes.

mavros_extras/src/plugins/obstacle_avoidance.cpp
```diff
@@ -68,167 +67,263 @@ class ObstacleAvoidancePlugin : public plugin::PluginBase {
...
+		// [[[cog:
+		// for p in ('1', '2', '3', '4', '5'):
+		//     cog.outl("Eigen::Vector3d position_point_%s, velocity_point_%s, acceleration_point_%s;" % (p, p, p))
+		//     cog.outl("tf::pointMsgToEigen(req->point_%s.position, position_point_%s);" % (p, p))
+		//     cog.outl("tf::vectorMsgToEigen(req->point_%s.velocity, velocity_point_%s);" % (p, p))
+		//     cog.outl("tf::vectorMsgToEigen(req->point_%s.acceleration_or_force, acceleration_point_%s);" % (p, p))
+		//     cog.outl("position_point_%s = ftf::transform_frame_enu_ned(position_point_%s);" % (p, p))
+		//     cog.outl("velocity_point_%s = ftf::transform_frame_enu_ned(velocity_point_%s);" % (p, p))
+		//     cog.outl("acceleration_point_%s = ftf::transform_frame_enu_ned(acceleration_point_%s);" % (p, p))
+		//     cog.outl("double yaw_ned_point_%s = wrap_pi(-req->point_%s.yaw + (M_PI / 2.0f));;" % (p, p))
```
vooon: Use `ftf::to_eigen()` 	https://github.com/mavlink/mavros/blob/master/mavros/include/mavros/frame_tf.h#L414 instead of `tf::`.
That allow chain functions.

Add white string between iterations.

mavros_extras/src/plugins/obstacle_avoidance.cpp
```diff
@@ -68,167 +67,263 @@ class ObstacleAvoidancePlugin : public plugin::PluginBase {
...
+		auto enu_position_point_5 = ftf::transform_frame_ned_enu(Eigen::Vector3d(avoid_input.point_5[0], avoid_input.point_5[1], avoid_input.point_5[2]));
+		auto enu_velocity_point_5 = ftf::transform_frame_ned_enu(Eigen::Vector3d(avoid_input.point_5[3], avoid_input.point_5[4], avoid_input.point_5[5]));
+		auto enu_acceleration_point_5 = ftf::transform_frame_ned_enu(Eigen::Vector3d(avoid_input.point_5[6], avoid_input.point_5[7], avoid_input.point_5[8]));
+		double yaw_enu_point_5 = wrap_pi((M_PI / 2.0f) - avoid_input.point_5[9]);
+		// [[[end]]] (checksum: 2b44b491fcaa6d23ac19857ed9ff504a)
+
+		// [[[cog:
+		// for p in ('1', '2', '3', '4', '5'):
+		//     for a, b in zip ("xyz", ('0', '1', '2')):
+		//         cog.outl("obstacle_avoidance_input->point_%s.position.%s = enu_position_point_%s(%s);" % (p, a, p, b))
```
vooon: Use Vector3d::x(), y, z() instead of index.

vooon: Also better to use python's 3 string.format(). E.g.
```python
cog.outl("m->point_{p}.position.{axis} = v_p{p}.{axis}();".format(**locals()))
```

mavros_msgs/msg/ObstacleAvoidance.msg
mhkabir: Just my 2 cents - I prefer the array.

mavros_extras/src/plugins/obstacle_avoidance.cpp
```diff
@@ -0,0 +1,330 @@
...
+public:
+	ObstacleAvoidancePlugin() : PluginBase(),
+		avoidance_nh("~avoidance")
+	{ }
+
+	void initialize(UAS &uas_)
+	{
+		PluginBase::initialize(uas_);
+
+		avoidance_outout_sub = avoidance_nh.subscribe("output", 10, &ObstacleAvoidancePlugin::avoidance_cb, this);
```
vooon: Misprint.

mavros_msgs/msg/ObstacleAvoidance.msg
TSC21: Actuallly @mrivi I think it would fit better also.

mrivi:
@vooon We have decide to rename the mavlink message `TRAJECTORY` since it a general way to represent a trajectory that could come from an obstacle avoidance module or a trajectory generator running on a companion computer.

TSC21:
@mrivi the plugin also needs to be adapted to the changes on the msg spec, specially to what concerns how the data is initialized and used (check latest @ChristophTobler commits).

mrivi:
I have added the support for `nav_msgs::Path` and I now distinguish between how the message is filled for `MAV_TRAJECTORY_REPRESENTATION::WAYPOINTS` and `(MAV_TRAJECTORY_REPRESENTATION::BEZIER` . I think I still need to fix the fact that in `mavros_msgs::Trajectory` there isn't any specific field for the time horizon

mavros_extras/src/plugins/trajectory.cpp
```diff
@@ -0,0 +1,579 @@
...
+		}
+
+		while (a >= M_PI) {
+			a -= (M_PI * 2.0f);
+		}
+
+		while (a < -M_PI) {
+			a += (M_PI * 2.0f);
+		}
+
```
AlexisTM: I would propose the following implementation, using modulo; ensuring constant execution time.

```c++
return fmod(a + M_PI, 2.0*M_PI) - M_PI;
```

mrivi:
I realized that the solution for the field `time_horizon` is not the most elegant. Still thinking how to improve it

mavros_extras/src/plugins/trajectory.cpp
```diff
@@ -172,50 +167,74 @@ class TrajectoryPlugin : public plugin::PluginBase {
...
+			// [[[end]]] (checksum: de8ef970a43d4b07fb7581c856c95980)
 		} else {
 			//[[[cog:
-			//for p in "12345":
-			//     cog.outl("trajectory.point_{p}[3] = req->point_{p}.velocity.x;".format(**locals()))
-			//     cog.outl("double yaw_ned_point_{p} = wrap_pi(-req->point_{p}.yaw + (M_PI / 2.0f)); \n".format(**locals()))
-			//     cog.outl("trajectory.point_{p}[4] = yaw_ned_point_{p};".format(**locals()))
-			//     cog.outl("trajectory.point_{p}[5] = req->point_{p}.yaw_rate; \n".format(**locals()))
+			//for p, index in zip("12345", "01234"):
+			//     cog.outl("trajectory.point_{p}[3] = req->time_horizon[{index}];".format(**locals()))
```
vooon: Note: follow python style, 4 spaces. Use one space between comment slashes and code.

mavros_extras/src/plugins/trajectory.cpp
```diff
@@ -172,50 +167,74 @@ class TrajectoryPlugin : public plugin::PluginBase {
...
+			//[[[cog:
+			//for p in "12345":
+			//     cog.outl("trajectory.point_{p}[4] = wrap_pi(-req->point_{p}.yaw + (M_PI / 2.0f));".format(**locals()))
+			//     cog.outl("trajectory.point_{p}[5] = NAN;".format(**locals()))
+			//     cog.outl("trajectory.point_{p}[6] = NAN;".format(**locals()))
+			//     cog.outl("trajectory.point_{p}[7] = NAN;".format(**locals()))
+			//     cog.outl("trajectory.point_{p}[8] = NAN;".format(**locals()))
+			//     cog.outl("trajectory.point_{p}[9] = NAN;".format(**locals()))
+			//     cog.outl("trajectory.point_{p}[10] = NAN;\n".format(**locals()))
+			// ]]]
```
vooon: That looks like may be better write that way:
```cpp

auto fill_points_variant = [](std::array<float, N?> &point, double yaw) {
  // cog...
}

// cog...
fill_points_variant(traectory.point_1, req->point_1.yaw);
```

mrivi: I am not sure I understand what you would like to achieve. Would you like to have a separate method to fill each of the field, e.g. yaw, yaw_speed, etc?

vooon: With lambda function you can reduce amount of generated code. That may look more readable.

mavros_extras/src/plugins/trajectory.cpp
```diff
@@ -255,133 +248,115 @@ class TrajectoryPlugin : public plugin::PluginBase {
...
+			point[10] = NAN;
+			// [[[end]]] (checksum: 78bad8df6dced32241db3b734ffd1361)
+		};
+
+		auto fill_points_unused = [] (std::array<float, LEN> &point) {
+			// [[[cog:
+			// for p in range(0, 11):
+			//     cog.outl("point[{p}] = NAN;".format(**locals()))
+			// ]]]
+			point[0] = NAN;
```
TSC21: `std::fill(point, array + LEN, NAN);`

mavros_extras/src/plugins/trajectory.cpp
```diff
@@ -255,133 +248,115 @@ class TrajectoryPlugin : public plugin::PluginBase {
...
+			point[9] = wrap_pi(-orientation_wp(2) + (M_PI / 2.0f));
+		};
+
+		auto fill_points_unused_fields = [] (std::array<float, LEN> &point) {
+			// [[[cog:
+			// for p in range(3, 9):
+			//     cog.outl("point[{p}] = NAN;".format(**locals()))
+			// cog.outl("point[10] = NAN;".format(**locals()))
+			// ]]]
+			point[3] = NAN;
```
TSC21: `std::fill(point + 3, array + LEN, NAN);`

mavros_extras/src/plugins/trajectory.cpp
```diff
@@ -73,175 +81,160 @@ class TrajectoryPlugin : public plugin::PluginBase {
 		trajectory.time_usec = req->header.stamp.toNSec() / 1000;	//!< [milisecs]
 		trajectory.type = req->type;	//!< trajectory type (waypoints, bezier)
 
-		/* Transformation from ENU to NED */
+		auto fill_points_position = [] (std::array<float, LEN> &point_output, const float x, const float y, const float z) {
+			// [[[cog:
+			//     cog.outl("auto position_ned = ftf::transform_frame_enu_ned(Eigen::Vector3d(x, y, z));")
+			//     for index, axis in zip ("012", "xyz"):
+			//         cog.outl("point_output[{index}] = position_ned.{axis}();".format(**locals()))
```
vooon: I think you can define function for that `outl()`. Else it is not necessary to include first line.
Also better to use ros geometry_msgs / Vector / Point as input. And anyway it should be `double`.

mavros_extras/src/plugins/trajectory.cpp
```diff
@@ -60,6 +63,11 @@ class TrajectoryPlugin : public plugin::PluginBase {
 
 	ros::Publisher trajectory_desired_pub;
 
+	// auto fill_points_variant = [](std::array<float, N?> &point, const float yaw) {
+	//      //point[9] = yaw;
+
+	//      print("%f %f \n", point[0], yaw);
+	// };
```
vooon: Debris?

mavros_extras/src/plugins/trajectory.cpp
```diff
@@ -73,175 +81,160 @@ class TrajectoryPlugin : public plugin::PluginBase {
...
+		};
+
+		auto fill_points_yaw_wp = [this](std::array<float, LEN> &point, float yaw) {
+			// cog...
+			point[9] = wrap_pi(-yaw + (M_PI / 2.0f));
+		};
+
+		auto fill_points_yaw_bezier = [this](std::array<float, LEN> &point, float yaw) {
+			// cog...
+			point[4] = wrap_pi(-yaw + (M_PI / 2.0f));
```
vooon: For single line cog is not necessary.

mavros_extras/src/plugins/trajectory.cpp
```diff
@@ -19,6 +19,9 @@
 #include <mavros_msgs/PositionTarget.h>
 #include <nav_msgs/Path.h>
 
+//! Point array length
+static constexpr int LEN = 11;
```
vooon: Call it `POINTS_LEN`, LEN is too general.

vooon: And that line may be computed from pymavlink (by cog), but not necessary for now.

mavros_msgs/msg/Trajectory.msg
```diff
@@ -7,12 +7,10 @@ uint8 type # See enum MAV_TRAJECTORY_REPRESENTATION.
...
-mavros_msgs/PositionTarget point_1
-mavros_msgs/PositionTarget point_2
-mavros_msgs/PositionTarget point_3
-mavros_msgs/PositionTarget point_4
-mavros_msgs/PositionTarget point_5
+float32[11] point_1
+float32[11] point_2
+float32[11] point_3
+float32[11] point_4
+float32[11] point_5
```
vooon: That looks bad, can it be defined in new message?

mavros_extras/src/plugins/trajectory.cpp
mrivi: Ok, I'll revert back to have each point in `Trajectory` to be a `mavros_msgs::PositionTarget` and have the time_horizon for bezier in the Trajectory. I changed everything to a general array because the 4th element of the array is either `vx` or `time_horizon` depending on the type. I totally see your point of using the mavros geometry_msgs etc and in fact that was my initial choice.

mavros_extras/src/plugins/trajectory.cpp
```diff
@@ -0,0 +1,517 @@
...
+
+		// [[[end]]] (checksum: 2f3c59967b0cd4b9167720a16adc1471)
+
+
+		// check that either x and y are finite or z to set the position waypoint as valid
+		if ((std::isfinite(trajectory.point_1[0]) && std::isfinite(trajectory.point_1[1])) || std::isfinite(trajectory.point_1[2])) {
+			trajectory.point_valid[0] = true;
+		} else {
+			trajectory.point_valid[0] = false;
+		}
```
vooon: That code:

```cpp

constexpr auto NUM_POINTS = 5; // ??? suggest better name

TR tr{};
// [[[cog:
// cog.outl(".... points {{ {tr_points} }};".format(",".join('tr.point_%s' % n for n in range(1, 6)))) 
// ]]]
std::array<&TR, NUM_POINTS> points { tr.point_1... };

// fill using points array... 

auto is_point_valid = [](... &point) -> bool {
return (std::isfinite(point[0]) ....);
}

for (size_t i = 0; i < points.size(); i++)
    tr.point_valid[i] = is_point_valid(points[i]);
```

mavros_extras/src/plugins/trajectory.cpp
```diff
@@ -0,0 +1,517 @@
...
+			acceleration.z = enu_acceleration.z();
+			// [[[end]]] (checksum: 0365bcb69209af709be922d638b7e473)
+		};
+
+		auto fill_points_yaw = [this](float &yaw_output, float yaw_input) {
+			yaw_output = wrap_pi((M_PI / 2.0f) - yaw_input);
+		};
+
+		auto fill_points_yaw_speed = [] (float yaw_speed, std::array<float, POINT_LEN> &point) {
+			yaw_speed = point[10];
```
vooon: Don't work, `yaw_speed` should be ref.

mavros_extras/src/plugins/trajectory.cpp
```diff
@@ -0,0 +1,517 @@
...
+		};
+
+		auto fill_points_time_horizon = [] (float time_horizon, std::array<float, POINT_LEN> &point){
```
vooon: Same problem.

vooon:
I think i can fix things that i don't like myself. So i'll merge as is.
@mrivi later i need your help to check plugin.

PR #986, opened 2018-03-31
cmake: fix FindGeographicLib for Debian systems

TSC21:
This adds support for CMake finding GeographicLib when the shared libs are in `/usr/lib/x86_64-linux-gnu/`. It restores the functionality that was lost with the simplification done in f2bda9739e202efdc61adc28acb2490051e35960.


TSC21:
@vooon it' passing Travis CI, so seems good to go.

TSC21:
Had some legacy src build that was putting a static lib on my OS environment (`/usr/local/lib/libGeographic.a`). A quick clean does makes it build with the current cmake module. Closing this.

PR #988, opened 2018-03-31
extras: add covariance parsing to VISION_POSITION_ESTIMATE

TSC21:
Updates `vision_pose_estimate` plugin so it can send the covariance matrix also. Follows the Mavlink spec update in https://github.com/mavlink/mavlink/pull/724. Requires a `mavlunk-gbp-release` update.
Also, generalizes `covariance9d_urt_to_mavlink` to `covariance_urt_to_mavlink` so to be able to parse URT values for matrices of other sizes.

### DIFF
--- a/mavros/include/mavros/frame_tf.h
+++ b/mavros/include/mavros/frame_tf.h
@@ -377,7 +377,7 @@ inline Eigen::Quaterniond mavlink_to_quaternion(const std::array<float, 4> &q)
 }
 
 /**
- * @brief Store Covariance matrix to MAVLink float[n] format
+ * @brief Convert covariance matrix to MAVLink float[n] format
  */
 template<class T, std::size_t SIZE>
 inline void covariance_to_mavlink(const T &cov, std::array<float, SIZE> &covmsg)
@@ -386,12 +386,12 @@ inline void covariance_to_mavlink(const T &cov, std::array<float, SIZE> &covmsg)
 }
 
 /**
- * @brief Store half upper right triangular of 9d Covariance matrix to MAVLink float[n] format
+ * @brief Convert half upper right triangular of a covariance matrix to MAVLink float[n] format
  */
-inline void covariance9d_urt_to_mavlink(const Covariance9d &cov, std::array<float, 45> &covmsg)
+template<class T, std::size_t SIZE>
+inline void covariance_urt_to_mavlink(const T &covmap, std::array<float, SIZE> &covmsg)
 {
-	EigenMapConstCovariance9d m(cov.data());
-
+	auto m = covmap;
 	auto out = covmsg.begin();
 
 	for (size_t x = 0; x < m.cols(); x++)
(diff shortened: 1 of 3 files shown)


mavros/include/mavros/frame_tf.h
```diff
@@ -386,12 +386,12 @@ inline void covariance_to_mavlink(const T &cov, std::array<float, SIZE> &covmsg)
 }
 
 /**
- * @brief Store half upper right triangular of 9d Covariance matrix to MAVLink float[n] format
+ * @brief Convert half upper right triangular of a covariance matrix to MAVLink float[n] format
  */
-inline void covariance9d_urt_to_mavlink(const Covariance9d &cov, std::array<float, 45> &covmsg)
+template<class T, std::size_t SIZE>
+inline void covariance_urt_to_mavlink(const T &covmap, std::array<float, SIZE> &covmsg)
```
vooon: Now it should have assert that number of URT elements of covmap == SIZE.

Also it may be a good idea to keep `covariance9d_...()` function (just change to call new one).

TSC21: > Now it should have assert that number of URT elements of covmap == SIZE.

Added now.

> Also it may be a good idea to keep `covariance9d_...()` function (just change to call new one).

For what reason, given that this generalizes the implementation?

TSC21:
@ChristophTobler, @mrivi this may also interest you

mavros/include/mavros/frame_tf.h
vooon: Use `ROS_ASSERT()`.

TSC21: > Use `ROS_ASSERT()`.

Makes sense. Done.

TSC21:
@vooon does the latest `mavlink-gbp-release` supports this new spec for the msg?

TSC21:
Can you wait for an update on the odometry plugin? I will do it today.

vooon:
It's too huge, let it be in .1 patch.

TSC21:
OK sounds reasonable

okalachev:
@TSC21 , I wonder, what would be the ROS way of passing the accelerations (ax, ay, az)? Should it be a nearby synchronised `AccelWithCovarianceStamped` topic, or a more general message with the both sub-messages?

TSC21:
> @TSC21 , I wonder, what would be the ROS way of passing the accelerations (ax, ay, az)? Should it be a nearby synchronised `AccelWithCovarianceStamped` topic, or a more general message with the both sub-messages?

On the `odom` plugin currently there is a possibility of passing them through the `LOCAL_POSITION_NED_COV` MAVLink msg, but since there is no handling on the autopilot side for the accels (I mean the estimated, not setpoints), being there or not is exactly the same. I changed this on #991. But answering your question, ROS may handle the accels using `[geometry_msgs/AccelWithCovarianceStamped](http://docs.ros.org/kinetic/api/geometry_msgs/html/msg/AccelWithCovarianceStamped.html), but in the autopilot perspective, it isn't used. It can be useful though if one adds a parser to `LOCAL_POSITION_NED_COV` on `local_position` and add a way of passing the accel data from the autopilot side, but I wonder what where would you get the two parts of the `geometry_msgs/Accel` (the linear and the angular).

PR #991, opened 2018-04-05
Extras: update odom plugin to send Mavlink ODOMETRY msgs

TSC21:
This updates the `odom` plugin to send [`ODOMETRY`](http://mavlink.org/messages/common#ODOMETRY) Mavlink msgs instead of [`LOCAL_POSITION_NED_COV`](http://mavlink.org/messages/common#LOCAL_POSITION_NED_COV) + [`ATT_COV`](http://mavlink.org/messages/common#ATTITUDE_QUATERNION_COV), which required an higher bandwidth given the size of `LOCAL_POSITION_NED_COV`. Also, this merges the odometry data into the same message, removing the need for posteriors sync of the pose and velocity data, which would get to the FCU at different time frames.
This will be followed by the respective support for the PX4 Pro firmware.

### DIFF
--- a/mavros_extras/src/plugins/odom.cpp
+++ b/mavros_extras/src/plugins/odom.cpp
@@ -25,7 +25,6 @@
 namespace mavros {
 namespace extra_plugins {
 using mavlink::common::MAV_FRAME;
-
 /**
  * @brief Odometry plugin
  *
@@ -50,6 +49,43 @@ public:
 
 		// subscribers
 		odom_sub = odom_nh.subscribe("odom", 10, &OdometryPlugin::odom_cb, this);
+
+		//! Map from param string to local MAV_FRAME
+		const std::unordered_map<std::string, MAV_FRAME> lf_map{
+			{ "vision_enu", MAV_FRAME::VISION_ENU },
+			{ "vision_ned", MAV_FRAME::VISION_NED },
+			{ "mocap_enu", MAV_FRAME::MOCAP_ENU },
+			{ "mocap_ned", MAV_FRAME::MOCAP_NED }
+		};
+
+		//! Map from param string to body MAV_FRAME
+		const std::unordered_map<std::string, MAV_FRAME> bf_map{
+			{ "ned", MAV_FRAME::BODY_NED },
+			{ "frd", MAV_FRAME::BODY_FRD },
+			{ "flu", MAV_FRAME::BODY_FLU }
+		};
+
+		auto local_frame_it = lf_map.find(local_frame);
+		auto body_frame_it = bf_map.find(body_frame_orientation);
+
+		/** Determine frame_id naming - considering the ROS msg frame_id
+		 * as "odom" by default
+		 */
+		if (local_frame_it == lf_map.end()) {
+			ROS_ERROR_NAMED("odom", "Odometry: invalid local frame \"%s\"", local_frame.c_str());
+			return;
+		}
+		else
+			lf_id = utils::enum_value(local_frame_it->second);
+
+		/** Determine child_frame_id naming
+		 */
+		if (body_frame_it == bf_map.end()) {
+			ROS_ERROR_NAMED("odom", "Odometry: invalid body frame orientation \"%s\"", body_frame_orientation.c_str());
+			return;
+		}
+		else
+			bf_id = utils::enum_value(local_frame_it->second);
 	}
 
 	Subscriptions get_subscriptions()
@@ -64,20 +100,8 @@ private:
 	std::string local_frame;		//!< orientation and source of the local frame (pose)
 	std::string body_frame_orientation;	//!< orientation of the body frame (velocity)
 
-	//! Map for local frame
-	std::unordered_map<std::string, MAV_FRAME> lf_map{
- 		{ "vision_enu", MAV_FRAME::VISION_ENU },
-		{ "vision_ned", MAV_FRAME::VISION_NED },
-		{ "mocap_enu", MAV_FRAME::MOCAP_ENU },
-		{ "mocap_ned", MAV_FRAME::MOCAP_NED }
-	};
-
-	//! Map for body frame
-	std::unordered_map<std::string, MAV_FRAME> bf_map{
-		{ "ned", MAV_FRAME::BODY_NED },
-		{ "frd", MAV_FRAME::BODY_FRD },
-		{ "flu", MAV_FRAME::BODY_FLU }
-	};
+	MAV_FRAME lf_id;			//!< local frame (pose) ID
+	MAV_FRAME bf_id;			//!< body frame (pose) ID
 
 	/**
 	 * @brief Sends odometry data msgs to the FCU.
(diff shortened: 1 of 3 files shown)


TSC21:
@ChristophTobler FYI.

mhkabir:
Woohoo!

mavros_extras/src/plugins/odom.cpp
```diff
@@ -64,134 +65,154 @@ class OdometryPlugin : public plugin::PluginBase {
...
+			if (frame_source == "vision")
+				msg.frame_id = utils::enum_value(MAV_FRAME::VISION_ENU);
+			else if (frame_source == "mocap")
+				msg.frame_id = utils::enum_value(MAV_FRAME::MOCAP_ENU);
+			else {
+				ROS_ERROR_NAMED("odom", "Odometry: invalid frame source \"%s\"", frame_source.c_str());
+				return;
+			}
+		}
+		else {
```
vooon: I think that may be a bit simplifyed by that method:
```cpp
std::unordered_map<const std::string, MAV_FRAME> m{{
 { "vision-enu", MAV_FRAME:VISION_ENU },
...
};

auto frame_it = m.find(local_frame_orientation + "-" + frame_source);
...
```

mavros_extras/src/plugins/odom.cpp
```diff
@@ -64,134 +65,154 @@ class OdometryPlugin : public plugin::PluginBase {
...
+							odom->child_frame_id, "fcu_" + body_frame_orientation, ros::Time(0)));
 
 			// transform lookup WRT body frame
-			tf_parent2ned = tf2::transformToEigen(m_uas->tf2_buffer.lookupTransform(
-							odom->header.frame_id, "local_origin_" + desired_frame, ros::Time(0)));
+			tf_parent2local = tf2::transformToEigen(m_uas->tf2_buffer.lookupTransform(
+							odom->header.frame_id, "local_origin_" + local_frame_orientation, ros::Time(0)));
 		} catch (tf2::TransformException &ex) {
 			ROS_WARN("odom: %s",ex.what());
 			ros::Duration(1.0).sleep();
```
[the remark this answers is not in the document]
TSC21: Yeah actually I didn't caught that, since it was @jgoppert implementation. I will replace with `ROS_ERROR_THROTTLE_NAMED()`.

mavros_extras/src/plugins/odom.cpp
TSC21: Yeah sounds good.

vooon:
Do string param things in `init()`, store ready to use enum. Mappings should be const.

mavros_extras/src/plugins/odom.cpp
```diff
@@ -50,6 +49,43 @@ class OdometryPlugin : public plugin::PluginBase {
...
+
+		auto local_frame_it = lf_map.find(local_frame);
+		auto body_frame_it = bf_map.find(body_frame_orientation);
+
+		/** Determine frame_id naming - considering the ROS msg frame_id
+		 * as "odom" by default
+		 */
+		if (local_frame_it == lf_map.end()) {
+			ROS_ERROR_NAMED("odom", "Odometry: invalid local frame \"%s\"", local_frame.c_str());
+			return;
```
vooon: Not return. Set to some default.

mavros_extras/src/plugins/odom.cpp
```diff
@@ -64,20 +100,8 @@ class OdometryPlugin : public plugin::PluginBase {
...
-	};
-
-	//! Map for body frame
-	std::unordered_map<std::string, MAV_FRAME> bf_map{
-		{ "ned", MAV_FRAME::BODY_NED },
-		{ "frd", MAV_FRAME::BODY_FRD },
-		{ "flu", MAV_FRAME::BODY_FLU }
-	};
+	uint8_t lf_id;				//!< local frame (pose) ID
+	uint8_t bf_id;				//!< body frame (pose) ID
```
vooon:
```
MAV_FRAME lf_id;
MAV_FRAME bf_id;
```

mavros_extras/src/plugins/odom.cpp
```diff
@@ -162,27 +186,8 @@ class OdometryPlugin : public plugin::PluginBase {
...
-		 */
-		auto body_frame_it = bf_map.find(body_frame_orientation);
-		if (body_frame_it == bf_map.end()) {
-			ROS_ERROR_NAMED("odom", "Odometry: invalid body frame orientation \"%s\"", body_frame_orientation.c_str());
-			return;
-		}
-		else
-			msg.child_frame_id = utils::enum_value(body_frame_it->second);
+		msg.frame_id = lf_id;
+		msg.child_frame_id = bf_id;
```
vooon:
```
= utils::enum_value(lf_id);
```

mavros_extras/src/plugins/odom.cpp
```diff
@@ -50,6 +49,43 @@ class OdometryPlugin : public plugin::PluginBase {
...
+
+		//! Map from param string to body MAV_FRAME
+		const std::unordered_map<std::string, MAV_FRAME> bf_map{
+			{ "ned", MAV_FRAME::BODY_NED },
+			{ "frd", MAV_FRAME::BODY_FRD },
+			{ "flu", MAV_FRAME::BODY_FLU }
+		};
+
+		auto local_frame_it = lf_map.find(local_frame);
+		auto body_frame_it = bf_map.find(body_frame_orientation);
```
vooon: Why not to use `mav_frame_from_str()`?

mavros_extras/src/plugins/odom.cpp
TSC21: Actually, I don't want it to send the wrong frame_id to the flight controller. That's why this is an error and not a warning.

mavros_extras/src/plugins/odom.cpp
TSC21: Because, as I explained to you, not all `MAV_FRAME` should be available to be passed, only odometry and body frame related ones. So a mapping is much faster than filtering `mav_frame_from_str()`.

vooon: Ok, but then why you use different strings (lower case, w/o BODY_ prefix)?

TSC21:
I am working on that now

TSC21:
Respective PR on PX4 side: https://github.com/PX4/Firmware/pull/9301

ArkadiuszNiemiec:
@TSC21 Excuse me for asking here but can you confirm if I can use `pose` from old script that was using `vision_pose_estimate` plugin, put it in `nav_msgs/Odometry` with correct frame and publish it to be used by new plugin? The only data from my EV system is x,y,z.

TSC21:
> @TSC21 Excuse me for asking here but can you confirm if I can use `pose` from old script that was using `vision_pose_estimate` plugin, put it in `nav_msgs/Odometry` with correct frame and publish it to be used by new plugin? The only data from my EV system is x,y,z.

I don't think I understand what are you trying to do exactly. What old script? And what was using `vision_pose_estimate`? `vision_pose_estimate` continues to work as it should and you can still use it.

ArkadiuszNiemiec:
> > @TSC21 Excuse me for asking here but can you confirm if I can use `pose` from old script that was using `vision_pose_estimate` plugin, put it in `nav_msgs/Odometry` with correct frame and publish it to be used by new plugin? The only data from my EV system is x,y,z.
> 
> I don't think I understand what are you trying to do exactly. What old script? And what was using `vision_pose_estimate`? `vision_pose_estimate` continues to work as it should and you can still use it.

Oh, so the current master of PX4 is still fusing mavlink message from `vision_pose_estimate`? PX4s mavlink receiver is no longer publishing it back to GCS so I assumed that it's deprecated.
I have a setup which was using `vision_pose_estimate` to fuse EV on stable PX4, I thought that I should switch to the new plugin.

TSC21:
> Oh, so the current master of PX4 is still fusing mavlink message from `vision_pose_estimate`?

Yes.

> PX4s mavlink receiver is no longer publishing it back to GCS so I assumed that it's deprecated.

Check: https://dev.px4.io/en/advanced/parameter_reference.html#MAV_ODOM_LP, though it just works for the `ODOMETRY` Mavlink msg.

> I have a setup which was using `vision_pose_estimate` to fuse EV on stable PX4, I thought that I should switch to the new plugin.

No need to.

ArkadiuszNiemiec:
@TSC21 Got it. I will keep using my current setup and in the mean time try to implement a branch working with your new plugin. I have wrote a crude test like this: `Odometry.pose.pose = PoseStamped.pose` with `MAV_ODOM_LP` set to `true` but it did not work. Guess I have to check if setting only `pose` in `Odometry` message is enough and learn in what frame should it be :wink: Thank you for your work!

TSC21:
The odom plugin does listen to PoseStamped topics. It expects a nav_msgs/Odometry topic.

ArkadiuszNiemiec:
@TSC21 Yes, I know, that's what I am sending. Anyway: `gazebo_iris_vision` with `MAV_ODOM_LP` set to `true` should publish `Odometry` back to QGroundControl, right? I cannot make it work, even in gazebo.

TSC21:
On sitl_gazebo, there is a CMake option to set the ODOMETRY stream. But please move this out of this PR discussion to PX4 Slack or similar.

PR #1001, opened 2018-04-12
imu plugin: publish differential pressure

tstastny:
publishes the differential pressure member of the highres imu message from px4.

PR #1002, opened 2018-04-13
imu plugin: fix pressure units

tstastny:
Sorry - converted units wrong on the differential and static pressure in #1001

PX4 also using new uorb for the static pressure which is in pa now (used to be millibar) https://github.com/PX4/Firmware/blob/master/msg/vehicle_air_data.msg#L4


TSC21:
Thanks

PR #1006, opened 2018-04-17
[WIP] extras: odom: adapt the transform according to the frames

TSC21:
The msgs spec states that the position and orientation should be on the local frame, while the twist data should be on the body frame.

**UPDATE:** this is not actually right -> it depends on the frames where the data is sent on. *example:* If a VIO app sends the velocity of a vehicle (the `child_frame_id` that navigates in an `odom` `frame_id`), then the velocity of the vehicle needs to be parsed on the the same frame of the `child_frame_id`, which means the vehicle body frame. The orientation should though be rotated from the FLU to the FRD orientation.

### DIFF
--- a/mavros_extras/src/plugins/odom.cpp
+++ b/mavros_extras/src/plugins/odom.cpp
@@ -97,9 +97,9 @@ private:
 	ros::NodeHandle odom_nh;
 	ros::Subscriber odom_sub;
 
-	std::string local_frame;		//!< orientation and source of the local frame (pose)
-	std::string body_frame_orientation;	//!< orientation of the body frame (velocity)
-	std::string local_frame_orientation;
+	std::string local_frame;		//!< orientation and source of the local frame
+	std::string local_frame_orientation;	//!< orientation of the local fram
+	std::string body_frame_orientation;	//!< orientation of the body frame
 
 	MAV_FRAME lf_id;			//!< local frame (pose) ID
 	MAV_FRAME bf_id;			//!< body frame (pose) ID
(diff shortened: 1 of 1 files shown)


TSC21:
On the FCU side, PX4 at least requires that the velocity is in the local frame, while the orientation represents the rotation from the local NED to the body frame, so in the FCU side those transforms should be handled.

TSC21:
Actually, this may be flexible! In http://docs.ros.org/api/nav_msgs/html/msg/Odometry.html:
>*# The pose in this message should be specified in the coordinate frame given by header.frame_id.
# The twist in this message should be specified in the coordinate frame given by the child_frame_id*

So there is no real obligation of one being on the local frame and the other on the body frame - it really depends on how we parse the data on the ROS side and how it is propagated on the Mavlink side. Depending on the frames, then one applies the respective transforms. That also means that we need to update the description on the Mavlink side.
@jgoppert what's your view on this?

jgoppert:
I believe that the translation is from frame_id to child_id so that locks in the velocities to be in the body frame for us.

TSC21:
> I believe that the translation is from frame_id to child_id so that locks in the velocities to be in the body frame for us.

@jgoppert I think that really depends on the source of the odometry data. I updated the PR so it actually iterates over the `frame_id` and `child_frame_id` (of the ROS msg) naming and decide on the type transform to apply, considering always that pose is WRT the `frame_id` reference (`odom` or `world`) and the twist is WRT the `child_frame_id` reference. I think it makes sense now and gives more flexibility in terms of the data we want to parse.
Can you review the last update? Thanks

TSC21:
After evaluating the `MAV_FRAME` enum values we have available and the increase of complexity with the possibility of having the pose and the velocity in both body and local frames, I set that, according to the available frames, the `frame_id` is always WRT a local frame, which means that the poses is WRT a local frame, and the `child_frame_id` can be WRT a local or a body frame, which means that the linear and angular velocities can be parsed WRT a local or a body frame.

mavros_extras/src/plugins/odom.cpp
```diff
@@ -116,98 +116,121 @@ class OdometryPlugin : public plugin::PluginBase {
...
-		/** Apply linear transforms
+		/** Apply transforms:
+		 * According to nav_msgs/Odometry.
+		 */
+		Eigen::Vector3d position {};		//!< Position vector. WRT frame_id
+		Eigen::Quaterniond orientation {};	//!< Attitude quaternion. WRT frame_id
+		Eigen::Vector3d lin_vel {};		//!< Linear velocity vector. WRT child_frame_id
+		Eigen::Vector3d ang_vel {};		//!< Angular velocity vector. WRT child_frame_id
+		Eigen::Matrix<double, 6, 6> r_pose {};	//!< Pose 6-D Covariance matrix. WRT frame_id
+		Eigen::Matrix<double, 6, 6> r_vel {};	//!< Velocity 6-D Covariance matrix. WRT child_frame_id
```
vooon: Define type somewhere.
```cpp
using Matrix6d = Eigen::Matrix<double, 6, 6>;
```

then

```cpp
auto r_pose = Matrix6d::Zero();
```

mavros_extras/src/plugins/odom.cpp
```diff
@@ -116,98 +116,121 @@ class OdometryPlugin : public plugin::PluginBase {
...
+			ang_vel = Eigen::Vector3d(tf_child2local.linear() * ftf::to_eigen(odom->twist.twist.angular));
+			r_vel.block<3, 3>(0, 0) = r_vel.block<3, 3>(3, 3) = tf_child2local.linear();
+
+			// the child_frame_id would be the same reference frame as frame_id
+			msg.child_frame_id = msg.frame_id;
+		}
+		else {
+			lin_vel = Eigen::Vector3d(tf_child2body.linear() * ftf::to_eigen(odom->twist.twist.linear));
+			ang_vel = Eigen::Vector3d(tf_child2body.linear() * ftf::to_eigen(odom->twist.twist.angular));
+			r_vel.block<3, 3>(0, 0) = r_vel.block<3, 3>(3, 3) = tf_child2body.linear();
```
vooon: Same code with only difference in Affine object. Redo that way:

```cpp

auto set_vel = [&](auto affine) {
   lin_vel = ....;
};

if (1) {
 set_vel(tf_child2local);
}
```

TSC21:
@vooon Travis is not liking the lambda expressions. Maybe related with the GCC compiler version?

Issue #1000, opened 2018-04-12
ROS2: Overview what deps needed to start porting


vooon:
Subj.

TSC21:
Here goes 1000 issues!!

TSC21:
Here was my start for this: https://github.com/mavlink/mavros/projects/1. Probably needs some updates.

LorenzMeier:
Hi - this is super relevant and a PR would be appreciated. What flight controller are you using with it?

TSC21:
Well while MAVROS is not ported to ROS2, the `ros1_bridge` is an acceptable solution. But to what repo are the PR's addressed to?

mikaelarguedas:
I have code changes for 3 packages ATM (more details in the gist above):
- uuid_msgs
- geographic_msgs
- mavros_msgs

AlexisTM:
Porting to ROS2, wouldn't it be the correct solution to move from MAVLink to FastRTPS for the backend communications?

TSC21:
@AlexisTM that's already an ongoing work yes.

ArkadiuszNiemiec:
Is there a branch with ROS2 code? I would like to check if I am able to help with porting and developing FastRTPS part.

TSC21:
There's no ROS2 branch code yet. It will be made public as soon it is found stable enough for it (at least for the PX4 side)

dlarue:
Looking for ROS2 also.

evandp:
Hello, what is the status of porting to ROS2? Are there plans to make the ROS2 branch public any time soon? Thanks!

TSC21:
There's no effort to port to ROS2 atm.

TSC21:
If you are interested though on the bridge between PX4 and ROS2 check https://dev.px4.io/v1.9.0/en/middleware/micrortps.html

PR #1007, opened 2018-04-17
convert mavros_msgs to ROS 2

mikaelarguedas:
This should target a `ros2` branch or equivalent

First commit ignores the other packages of this repository
Second commit changes package.xml and CMake to build with ROS2
Third commit renames the message failing to compile in ROS2 and adds a ros1_bridge mapping rule file to be able to use it over the ros1_bridge.
I'm open to suggestions regarding the message renaming. The issue is that underscores are not allowed in message names in ROS2, so I changed it to CamelCase

Note: I didn't port the utilities from the include directories in this PR, this can be done in this PR if necessary or in a follow-up

connects to https://github.com/mavlink/mavros/issues/1000


mikaelarguedas:
This PR took existing ROS 2 message packages as example.
The diff is still significant and hopefully could be reduce further in the future but is a good example (IMO) of porting message packages to ROS 2.

TSC21:
Yes it is! Thanks. Havr you checked Project #1 for MAVROS? I have mapped there for some time already most of the things that require porting. It will be a thing that would take time but for sure useful in the end.

mikaelarguedas:
> Have you checked Project 1 for MAVROS?

Oh that's a great initiative!

Does one need to be a member of the mavlink organization to add/modify items on that board ?

PR #1014, opened 2018-04-22
add wind estimation plugin

### LINKED ISSUE #1008: handle wind_cov (px4) for wind estimation publisher
I'd like to handle and publish values from PX4's wind_cov msg [WIND_COV](https://mavlink.io/en/messages/common.html#WIND_COV) on mavros. The APM specific wind msg is handled currently in the vfr_hud plugin [L76](https://github.com/mavlink/mavros/blob/master/mavros/src/plugins/vfr_hud.cpp#L76). However, the PX4 specific message also includes some variances on the estimate.

I'm motivated to PR this addition -- but would appreciate your preference on how to include it (to avoid remaking the wheel a few times):
1. Should the additional variances (horizontal and vertical) on the wind estimate be put into a separate ros message type, and leave the twist as is? Or should I make a new ros msg specifically for the wind_estimation publisher, and convert the current APM handler to populate this msg?
2. Does it make sense to keep this in vfr_hud? Or rather should it be in it's own plugin?

Let me know your thoughts please. Thanks! 

@TSC21

tstastny:
Makes a new wind estimation plugin handling both APM and PX4 specific wind messages.
- moves old APM wind publisher from vfr_hud to the new wind plugin
- the PX4 message contains a *summed* horizontal wind speed convariance (i.e. x cov + y cov). This value is placed in the first covariance array slot for now, and the vertical in the 3rd.

resolves #1008

### DIFF
--- a/mavros/src/plugins/wind_estimation.cpp
+++ b/mavros/src/plugins/wind_estimation.cpp
@@ -1,0 +1,96 @@
+/**
+ * @brief wind estimation plugin
+ * @file wind_estimation.cpp
+ * @author Thomas Stastny <thomas.stastny@mavt.ethz.ch>
+ *
+ * @addtogroup plugin
+ * @{
+ */
+/*
+ * Copyright 2018 Thomas Stastny <thomas.stastny@mavt.ethz.ch>
+ *
+ * This file is part of the mavros package and subject to the license terms
+ * in the top-level LICENSE file of the mavros repository.
+ * https://github.com/mavlink/mavros/tree/master/LICENSE.md
+ */
+
+#include <angles/angles.h>
+#include <mavros/mavros_plugin.h>
+
+#include <geometry_msgs/TwistWithCovarianceStamped.h>
+
+namespace mavros {
+namespace std_plugins {
+/**
+ * @brief Wind estimation plugin.
+ */
+class WindEstimationPlugin : public plugin::PluginBase {
+public:
+	WindEstimationPlugin() : PluginBase(),
+		nh("~")
+	{ }
+
+	/**
+	 * Plugin initializer. Constructor should not do this.
+	 */
+	void initialize(UAS &uas_)
+	{
+		PluginBase::initialize(uas_);
+
+		wind_pub = nh.advertise<geometry_msgs::TwistWithCovarianceStamped>("wind_estimation", 10);
+	}
+
+	Subscriptions get_subscriptions()
+	{
+		return {
+			make_handler(&WindEstimationPlugin::handle_APM_wind),
+			make_handler(&WindEstimationPlugin::handle_PX4_wind),
+		};
+	}
+
+private:
+	ros::NodeHandle nh;
+
+	ros::Publisher wind_pub;
+
+	/**
+	 * Handle APM specific wind direction estimation message
+	 */
+	void handle_APM_wind(const mavlink::mavlink_message_t *msg, mavlink::ardupilotmega::msg::WIND &wind)
+	{
+		const double speed = wind.speed;
+		const double course = angles::from_degrees(wind.direction);
+
+		auto twist_cov = boost::make_shared<geometry_msgs::TwistWithCovarianceStamped>();
+		twist_cov->header.stamp = ros::Time::now();
+		// TODO: check math's
+		twist_cov->twist.twist.linear.x = speed * std::sin(course);
+		twist_cov->twist.twist.linear.y = speed * std::cos(course);
+		twist_cov->twist.twist.linear.z = wind.speed_z;
+
+		wind_pub.publish(twist_cov);
+	}
+
+	/**
+	 * Handle APM specific wind direction estimation message
+	 */
+	void handle_PX4_wind(const mavlink::mavlink_message_t *msg, mavlink::common::msg::WIND_COV &wind)
+	{
+		auto twist_cov = boost::make_shared<geometry_msgs::TwistWithCovarianceStamped>();
+		twist_cov->header.stamp = ros::Time::now();
+
+		twist_cov->twist.twist.linear.x = wind.wind_x;
+		twist_cov->twist.twist.linear.y = wind.wind_y;
+		twist_cov->twist.twist.linear.z = wind.wind_z;
+
+		twist_cov->twist.covariance[0] = wind.var_horiz; // NOTE: this is a summed covariance for both x and y horizontal wind components
+		twist_cov->twist.covariance[2] = wind.var_vert;
+
+		wind_pub.publish(twist_cov);
+	}
+};
+}	// namespace std_plugins
+}	// namespace mavros
+
+#include <pluginlib/class_list_macros.h>
+PLUGINLIB_EXPORT_CLASS(mavros::std_plugins::WindEstimationPlugin, mavros::plugin::PluginBase)
--- a/mavros/src/plugins/wind_estimation.cpp
+++ b/mavros/src/plugins/wind_estimation.cpp
@@ -45,8 +45,8 @@ public:
 	Subscriptions get_subscriptions()
 	{
 		return {
-			make_handler(&WindEstimationPlugin::handle_apm_wind),
-			make_handler(&WindEstimationPlugin::handle_px4_wind),
+			       make_handler(&WindEstimationPlugin::handle_apm_wind),
+			       make_handler(&WindEstimationPlugin::handle_px4_wind),
 		};
 	}
 
@@ -61,14 +61,14 @@ private:
 	void handle_apm_wind(const mavlink::mavlink_message_t *msg, mavlink::ardupilotmega::msg::WIND &wind)
 	{
 		const double speed = wind.speed;
-		const double course = -angles::from_degrees(wind.direction); // direction "from" -> direction "to"
+		const double course = -angles::from_degrees(wind.direction);	// direction "from" -> direction "to"
 
 		auto twist_cov = boost::make_shared<geometry_msgs::TwistWithCovarianceStamped>();
 		twist_cov->header.stamp = ros::Time::now();
 		// TODO: check math's
-		twist_cov->twist.twist.linear.x = speed * std::sin(course); // E
-		twist_cov->twist.twist.linear.y = speed * std::cos(course); // N
-		twist_cov->twist.twist.linear.z = -wind.speed_z; // D -> U
+		twist_cov->twist.twist.linear.x = speed * std::sin(course);	// E
+		twist_cov->twist.twist.linear.y = speed * std::cos(course);	// N
+		twist_cov->twist.twist.linear.z = -wind.speed_z;// D -> U
 
 		// covariance matrix unknown in APM msg
 		ftf::EigenMapCovariance6d cov_map(twist_cov->twist.covariance.data());
@@ -92,7 +92,7 @@ private:
 		// fill available covariance elements
 		ftf::EigenMapCovariance6d cov_map(twist_cov->twist.covariance.data());
 		cov_map.setZero();
-		cov_map(0, 0) = wind.var_horiz; // NOTE: this is a summed covariance for both x and y horizontal wind components
+		cov_map(0, 0) = wind.var_horiz;	// NOTE: this is a summed covariance for both x and y horizontal wind components
 		cov_map(2, 2) = wind.var_vert;
 
 		wind_pub.publish(twist_cov);
--- a/mavros/src/plugins/wind_estimation.cpp
+++ b/mavros/src/plugins/wind_estimation.cpp
@@ -1,5 +1,5 @@
 /**
- * @brief wind estimation plugin
+ * @brief Wind estimation plugin
  * @file wind_estimation.cpp
  * @author Thomas Stastny <thomas.stastny@mavt.ethz.ch>
  *
@@ -14,8 +14,10 @@
  * https://github.com/mavlink/mavros/tree/master/LICENSE.md
  */
 
-#include <angles/angles.h>
 #include <mavros/mavros_plugin.h>
+
+#include <angles/angles.h>
+#include <eigen_conversions/eigen_msg.h>
 
 #include <geometry_msgs/TwistWithCovarianceStamped.h>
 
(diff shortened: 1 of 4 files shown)


mavros/src/plugins/wind_estimation.cpp
```diff
@@ -0,0 +1,96 @@
...
+		// TODO: check math's
+		twist_cov->twist.twist.linear.x = speed * std::sin(course);
+		twist_cov->twist.twist.linear.y = speed * std::cos(course);
+		twist_cov->twist.twist.linear.z = wind.speed_z;
+
+		wind_pub.publish(twist_cov);
+	}
+
+	/**
+	 * Handle APM specific wind direction estimation message
```
TSC21: Typo

mavros/src/plugins/wind_estimation.cpp
```diff
@@ -0,0 +1,96 @@
...
+	void handle_PX4_wind(const mavlink::mavlink_message_t *msg, mavlink::common::msg::WIND_COV &wind)
+	{
+		auto twist_cov = boost::make_shared<geometry_msgs::TwistWithCovarianceStamped>();
+		twist_cov->header.stamp = ros::Time::now();
+
+		twist_cov->twist.twist.linear.x = wind.wind_x;
+		twist_cov->twist.twist.linear.y = wind.wind_y;
+		twist_cov->twist.twist.linear.z = wind.wind_z;
+
+		twist_cov->twist.covariance[0] = wind.var_horiz; // NOTE: this is a summed covariance for both x and y horizontal wind components
```
TSC21: What about splitting this on the Firmware side also?

mavros/src/plugins/wind_estimation.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+	/**
+	 * Handle APM specific wind direction estimation message
+	 */
+	void handle_PX4_wind(const mavlink::mavlink_message_t *msg, mavlink::common::msg::WIND_COV &wind)
+	{
+		auto twist_cov = boost::make_shared<geometry_msgs::TwistWithCovarianceStamped>();
+		twist_cov->header.stamp = ros::Time::now();
+
+		twist_cov->twist.twist.linear.x = wind.wind_x;
```
TSC21: Are you sure we don't have to do frame transforms here?

mavros/src/plugins/wind_estimation.cpp
```diff
@@ -0,0 +1,96 @@
...
+	 */
+	void handle_APM_wind(const mavlink::mavlink_message_t *msg, mavlink::ardupilotmega::msg::WIND &wind)
+	{
+		const double speed = wind.speed;
+		const double course = angles::from_degrees(wind.direction);
+
+		auto twist_cov = boost::make_shared<geometry_msgs::TwistWithCovarianceStamped>();
+		twist_cov->header.stamp = ros::Time::now();
+		// TODO: check math's
+		twist_cov->twist.twist.linear.x = speed * std::sin(course);
```
TSC21: No frame transforms?

mavros/src/plugins/wind_estimation.cpp
```diff
@@ -0,0 +1,96 @@
...
+		wind_pub.publish(twist_cov);
+	}
+
+	/**
+	 * Handle APM specific wind direction estimation message
+	 */
+	void handle_PX4_wind(const mavlink::mavlink_message_t *msg, mavlink::common::msg::WIND_COV &wind)
+	{
+		auto twist_cov = boost::make_shared<geometry_msgs::TwistWithCovarianceStamped>();
+		twist_cov->header.stamp = ros::Time::now();
```
TSC21: You have a timestamp on the msg. So you can do add timesync:
```c++
twist_cov->header.stamp = synchronise_stamp(wind.time_usec)
```

mavros/src/plugins/wind_estimation.cpp
```diff
@@ -0,0 +1,96 @@
...
+		const double speed = wind.speed;
+		const double course = angles::from_degrees(wind.direction);
+
+		auto twist_cov = boost::make_shared<geometry_msgs::TwistWithCovarianceStamped>();
+		twist_cov->header.stamp = ros::Time::now();
+		// TODO: check math's
+		twist_cov->twist.twist.linear.x = speed * std::sin(course);
+		twist_cov->twist.twist.linear.y = speed * std::cos(course);
+		twist_cov->twist.twist.linear.z = wind.speed_z;
+
```
TSC21: You should initialize the covariance matrix of the ROS msg to unknown by setting element 0 of the associated covariance matrix to -1.

mavros/src/plugins/wind_estimation.cpp
tstastny: Problem is in the mavlink message itself, (WIND_COV)[https://mavlink.io/en/messages/common.html#WIND_COV]. Only has "horizontal" and "vertical" as members. As I understand it, Mavlink is a "write only" process, correct? Otherwise, I'd for sure be in favor of splitting them.

mavros/src/plugins/wind_estimation.cpp
tstastny: wind x y z from PX4 is in NED local frame. should be ok as long as mavros also follows this convention. let me know if it does not.

mavros/src/plugins/wind_estimation.cpp
tstastny: this is a copy paste of the APM handler that was in VFR_HUD before. I assume this should be ok? Unless someone with APM experience could point out any changes since that handler was written.

mavros/src/plugins/wind_estimation.cpp
```diff
@@ -68,21 +68,26 @@ class WindEstimationPlugin : public plugin::PluginBase {
 		twist_cov->twist.twist.linear.y = speed * std::cos(course);
 		twist_cov->twist.twist.linear.z = wind.speed_z;
 
+		// fill available covariance elements (-1 = unknown)
+		for (int i = 0; i < 36; ++i) twist_cov->twist.covariance[i] = -1.0;
```
TSC21: Use `std::array` and `std::copy()` instead. And only index 0 is to be set to -1. The others should be 0.

mavros/src/plugins/wind_estimation.cpp
TSC21: MAVLink can be extended. One does not have to replace fields. You can so update the description of the fields.

mavros/src/plugins/wind_estimation.cpp
tstastny: You want to represent a *complete* unknown covariance matrix by only putting -1 in the first element of the array (and the rest zeros), and then on the other (PX4) implementation, just fill unknowns with zeros? I dont understand, can you clarify please? I thought you wanted any unknown as a -1 (which is all for APM, and all but two for PX4).

TSC21: ROS convention states that only the first value of the array (covariance matrix) should be valued -1 when the values of the covariance matrix are unknown or invalid. There is no convention for the autopilot side or to MAVLink, though I am also planning to open a RFC so that can be discussed.

TSC21: Where is that convention for unkown covariance matrix values stated on the autopilot side? I am not aware of those. For MAVLink that does not exist.

mavros/src/plugins/wind_estimation.cpp
tstastny: I guess that would be a bit of an odd update though. "var_horiz" and "var_horiz2"? My guess would be that they originally put only the summed covariance for horizontal as they weren't expecting the horizontal components to have that much different of an uncertainty anyway (a somewhat reasonable expectation looking at how the wind is estimated). Though for sure vertical and horizontal will have different uncertainties. Do you think it is worth to add a field at this point? Or just keep it as it is for now? If you want to change it, could you recommend a field name?

TSC21: If it is a sum, would it be reasonable to consider a split by half to the two components (X and Y)?

mavros/src/plugins/wind_estimation.cpp
tstastny: I have no idea. That's why I asked. I originally just assumed one would default the values to zero if not populated. Was just confused at the first comment. I'll implement with the ROS convention for APM message (since there is no covariance getting sent in the msg), and fill the numbers present in the PX4 message, and the rest zeros. Let me know if I'm still not getting it right.

TSC21: Well probably that need some clarification. Can you open an issue on PX4 so someone can clarify it? I suppose the answer should be in the wind estimator for that case, but better having someone clarify it. But yes for APM makes sense to apply the ROS convention. For PX4, the question remains regarding the XY components.

mavros/src/plugins/wind_estimation.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+private:
+	ros::NodeHandle nh;
+
+	ros::Publisher wind_pub;
+
+	/**
+	 * Handle APM specific wind estimation message
+	 */
+	void handle_APM_wind(const mavlink::mavlink_message_t *msg, mavlink::ardupilotmega::msg::WIND &wind)
```
vooon: Should be named `handle_apm_wind()`.

mavros/src/plugins/wind_estimation.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+		auto twist_cov = boost::make_shared<geometry_msgs::TwistWithCovarianceStamped>();
+		twist_cov->header.stamp = ros::Time::now();
+		// TODO: check math's
+		twist_cov->twist.twist.linear.x = speed * std::sin(course);
+		twist_cov->twist.twist.linear.y = speed * std::cos(course);
+		twist_cov->twist.twist.linear.z = wind.speed_z;
+
+		// fill available covariance elements (-1 = unknown)
+		for (int i = 0; i < 36; ++i) twist_cov->twist.covariance[i] = -1.0;
```
vooon: Should be:
```cpp

ftf::EigenMapCovariance6d cov_map(twist_cov->twist->covariance.data());

cov_map.setZero();
cov_map(0, 0) = -1.0;
```

mavros/src/plugins/wind_estimation.cpp
tstastny: So I checked a recent flight log -- looks like maybe not the best to assume they are equal (though they are somehow close). see
wind_cov1 and (zoomed)
wind_cov2

I would probably be for leaving it as is for the time being, and if someone wants more ekf detailed info -- maybe we would need to make a new mavlink msg specifically for that, leaving this message as a common default one for people who dont care to have every piece of info. Otherwise your other suggestion of extending somehow the current message with a few extra covariance values could also be an option.

I actually don't think vertical wind is even estimated on EKF2 currently. Though I may look into some model-based extensions to help with this later this summer. But for now PX4 populates that field = 0.

mavros/src/plugins/wind_estimation.cpp
```diff
@@ -0,0 +1,101 @@
...
+		twist_cov->header.stamp = m_uas->synchronise_stamp(wind.time_usec);
+
+		twist_cov->twist.twist.linear.x = wind.wind_x;
+		twist_cov->twist.twist.linear.y = wind.wind_y;
+		twist_cov->twist.twist.linear.z = wind.wind_z;
+
+		// fill available covariance elements (-1 = unknown)
+		for (int i = 0; i < 36; ++i) twist_cov->twist.covariance[i] = -1.0;
+		twist_cov->twist.covariance[0] = wind.var_horiz; // NOTE: this is a summed covariance for both x and y horizontal wind components
+		twist_cov->twist.covariance[2] = wind.var_vert;
```
vooon: Use `ftf::EigenMapCovariance6d` to fill matrix.

tstastny:
@TSC21 @vooon 
- fixed the conventions (again, maybe APM needs to be double checked, but I believe after looking at their code briefly, seems to be correct now)
- used frame tf for the PX4 side, direct NED -> ENU. but not on the APM side, as the components were anyway getting calculated in the handler. let me know if this is an issue.

still open question on:
- [ ] x y components of PX4 wind covariance. I would vote for leaving it as is for now. then updating later if someone fills out the covariance matrix more. I don't think it is such a good idea to divide it in two (see [this](https://github.com/mavlink/mavros/pull/1014#discussion_r183327751) comment).

mavros/src/plugins/wind_estimation.cpp
```diff
@@ -54,37 +56,44 @@ class WindEstimationPlugin : public plugin::PluginBase {
...
+	void handle_px4_wind(const mavlink::mavlink_message_t *msg, mavlink::common::msg::WIND_COV &wind)
 	{
 		auto twist_cov = boost::make_shared<geometry_msgs::TwistWithCovarianceStamped>();
-		twist_cov->header.stamp = ros::Time::now();
+		twist_cov->header.stamp = m_uas->synchronise_stamp(wind.time_usec);
 
-		twist_cov->twist.twist.linear.x = wind.wind_x;
-		twist_cov->twist.twist.linear.y = wind.wind_y;
-		twist_cov->twist.twist.linear.z = wind.wind_z;
+		tf::vectorEigenToMsg(ftf::transform_frame_ned_enu(Eigen::Vector3d(wind.wind_x, wind.wind_y, wind.wind_z)),
```
TSC21: Did you run `uncrustify` on the code?

TSC21: https://github.com/mavlink/mavros/blob/master/CONTRIBUTING.md#contributing

TSC21:
> - used frame tf for the PX4 side, direct NED -> ENU. but not on the APM side, as the components were anyway getting calculated in the handler. let me know if this is an issue.

@khancyr anything you can look into?

> - [ ] x y components of PX4 wind covariance. I would vote for leaving it as is for now. then updating later if someone fills out the covariance matrix more. I don't think it is such a good idea to divide it in two (see [this](https://github.com/mavlink/mavros/pull/1014#discussion_r183327751) comment).

@tstastny I actually think it is a good idea to extend it, for the reason you saw - they are different and should be treated differently. But first, a RFC should be opened [here](https://github.com/mavlink/rfcs) so this can be discussed.

tstastny:
> But first, a RFC should be opened here so this can be discussed.

Ok I will try to do an RFC some time this week. @TSC21 do you want to wait with this plugin until that whole process has been done? Or do we put what's there now (still useful for anyone simply wanting the wind estimates and not caring about variances) and update accordingly later?

vooon:
I think we may merge that version and update later.

PR #1016, opened 2018-04-23
glob pos plugin: correct gps velocity convention (NEU->ENU)

tstastny:


tstastny:
> We have functions to handle the frame tf.

Yes but this seems unnecessary in this case. To use the frame tf, one would anyway need to negate the vz on the input argument, then use the extra operations to again negative the already negated vz and flip the vx and vy. Note the vz comes in positive already from MAVLink on [GLOBAL_POSITION_INT](https://mavlink.io/en/messages/common.html#GLOBAL_POSITION_INT), it is not in NED format, but NEU.

Is this still preferable?

TSC21:
Actually you are right. I feel akward though that we are using Z-up, since it does not match the aircraft convention. Will merge as is.

mhkabir:
@TSC21 @tstastny I believe this is incorrect. The message description was originally incorrect and is now fixed : https://github.com/mavlink/mavlink/pull/892

tstastny:
@mhkabir good catch. you're right -- the uorb topics give NED. the mavlink.io page is still not updated though -- where could one PR this?

TSC21:
Then this is not right either https://github.com/mavlink/mavlink/pull/892/files#diff-ceacc1e2390186650146a45be8fcb691R3092. I think there's some inconsistency here. @julianoes can you comment on this? Thanks

Issue #1020, opened 2018-05-02
GPS horizontal accuracy


pavloblindnology:
What is the reason of using HDOP value for computing horizontal accuracy of `GPS_RAW_INT` message instead of using `h_acc` values contained in the message? HDOP value is a scale factor, so, at least it should be multiplied by some metric accuracy estimation. Otherwise, it gives unrealistically small values.
https://github.com/mavlink/mavros/blob/38c6fff5aebf8d474e2fc6cad429ec99382bc0dc/mavros/src/plugins/global_position.cpp#L175

vooon:
Just because that code done for a very long time before GPS_RAW_INT have extended with that fields.
And note, extension do not works with mavlink v1.0.

pavloblindnology:
Understood.
Can we use extension values if on mavlink v2.0 and HDOP otherwise?
Also, I'd add some metric accuracy estimation parameter to config file to be scaled by HDOP, VDOP.
Is it feasible?

TSC21:
> Understood.
Can we use extension values if on mavlink v2.0 and HDOP otherwise?
Also, I'd add some metric accuracy estimation parameter to config file to be scaled by HDOP, VDOP.
Is it feasible?

@pavloblindnology that seems reasonable to me. Can you please do a PR with it? Thank you!

PR #1034, opened 2018-05-24
GPS accuracy wo approximations

pavloblindnology:
GPS accuracy published by mavros is now based on accuracies reported by FCU in `GPS_RAW_INT` message if using mavlink v2.0. Otherwise they are approximated by HDOP & VDOP values, scaled by GPS User Equivalent Range Error (UERE) measured in meters.
According to #1020.

### DIFF
--- a/mavros/src/plugins/global_position.cpp
+++ b/mavros/src/plugins/global_position.cpp
@@ -59,6 +59,7 @@ public:
 		gp_nh.param<std::string>("frame_id", frame_id, "map");
 		gp_nh.param<std::string>("child_frame_id", child_frame_id, "base_link");
 		gp_nh.param("rot_covariance", rot_cov, 99999.0);
+		gp_nh.param("gps_uere", gps_uere, 1.0);
 		gp_nh.param("use_relative_alt", use_relative_alt, true);
 		// tf subsection
 		gp_nh.param("tf/send", tf_send, false);
@@ -127,6 +128,7 @@ private:
 	bool is_map_init;
 
 	double rot_cov;
+	double gps_uere;
 
 	Eigen::Vector3d map_origin {};	//!< geodetic origin of map frame [lla]
 	Eigen::Vector3d ecef_origin {};	//!< geocentric origin of map frame [m]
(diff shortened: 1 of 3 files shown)


mavros/src/plugins/global_position.cpp
```diff
@@ -168,12 +170,17 @@ class GlobalPositionPlugin : public plugin::PluginBase {
...
 
-		if (!std::isnan(eph)) {
-			const double hdop = eph;
-
-			// From nmea_navsat_driver
-			fix->position_covariance[0 + 0] = fix->position_covariance[3 + 1] = std::pow(hdop, 2);
-			fix->position_covariance[6 + 2] = std::pow(2 * hdop, 2);
+		// With FCU protocol v2.0 use accuracies reported by sensor
+		if (UAS_FCU(m_uas)->get_protocol_version() == mavconn::Protocol::V20 &&
+				raw_gps.h_acc > 0 && raw_gps.v_acc > 0) {
```
vooon: That is ok, but i think that bit faster: `msg->magic == MAVLINK_STX`.

- `MAVLINK_STX` = 0xfd (v2.0)
- `MAVLINK_STX_MAVLINK1` = 0xfe (v1.0)

vooon: Also that check is not depend on link configuration, because it always can receive v1 and v2 msgs.

mavros/src/plugins/global_position.cpp
```diff
@@ -168,12 +170,17 @@ class GlobalPositionPlugin : public plugin::PluginBase {
...
+			fix->position_covariance[0 + 0] = fix->position_covariance[3 + 1] = std::pow(raw_gps.h_acc / 1E3, 2);
+			fix->position_covariance[6 + 2] = std::pow(raw_gps.v_acc / 1E3, 2);
+			fix->position_covariance_type = sensor_msgs::NavSatFix::COVARIANCE_TYPE_DIAGONAL_KNOWN;
```
vooon: About filling cov matrix, i'd like to see Eigen::Map here ([ftf::EigenMapCovariance3d][1]), old code done long before we start to use Eigen.

[1]: http://docs.ros.org/lunar/api/mavros/html/group__nodelib.html#gabafbfa3aa489d1c304f320a417ce91d1

vooon: Pseudocode:
```cpp
ftf::EigenMapCovariance3d cov(fix->position_covariance.data());

xy = pow(...);
cov.diagonal() << xy, xy, z;
```

pavloblindnology:
Updated according to review.

mavros/src/plugins/global_position.cpp
vooon: Hmm, perhaps it is good idea to add `is_mavlink_v10` and `is_mavlink_v20()` inline helpers...
Will add that after merge.

PR #1042, opened 2018-05-31
Advanced Time Synchronisation

mhkabir:
Mavros version of https://github.com/PX4/Firmware/pull/9365.

Tested in SITL. Pending test on vehicle, but everything should be working correctly already.

### DIFF
--- a/mavros/src/plugins/sys_time.cpp
+++ b/mavros/src/plugins/sys_time.cpp
@@ -18,6 +18,7 @@
 
 #include <sensor_msgs/TimeReference.h>
 #include <std_msgs/Duration.h>
+#include <mavros_msgs/TimesyncStatus.h>
 
 namespace mavros {
 namespace std_plugins {
@@ -37,9 +38,9 @@ public:
 		tolerance_(0.1),
 		times_(win_size),
 		seq_nums_(win_size),
-		last_dt(0),
-		dt_sum(0),
-		last_ts(0),
+		last_rtt(0),
+		rtt_sum(0),
+		last_remote_ts(0),
 		offset(0)
 	{
 		clear();
@@ -51,7 +52,7 @@ public:
 
 		ros::Time curtime = ros::Time::now();
 		count_ = 0;
-		dt_sum = 0;
+		rtt_sum = 0;
 
 		for (int i = 0; i < window_size_; i++)
 		{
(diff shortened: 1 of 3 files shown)


mavros/src/plugins/sys_time.cpp
```diff
@@ -165,12 +171,45 @@ class SystemTimePlugin : public plugin::PluginBase {
...
+		// in the estimate. Larger values will cause low-amplitude oscillations.
+		//
+		// Beta : Used to smooth the clock skew estimate. Smaller values will lead to a
+		// tighter estimation of the skew (derivative), but will negatively affect how fast the
+		// filter reacts to clock skewing (e.g cause by temperature changes to the oscillator).
+		// Larger values will cause large-amplitude oscillations.
+		nh.param("time/timesync_alpha_initial", filter_alpha_initial, 0.05f);
+		nh.param("time/timesync_beta_initial", filter_beta_initial, 0.05f);
+		nh.param("time/timesync_alpha_final", filter_alpha_final, 0.003f);
+		nh.param("time/timesync_alpha_final", filter_beta_final, 0.003f);
```
vooon: Bug.

vooon: Use cog.

mavros/src/plugins/sys_time.cpp
```diff
@@ -289,30 +354,122 @@ class SystemTimePlugin : public plugin::PluginBase {
...
+			}
+		} else {
+			// Increment counter if round trip time is too high for accurate timesync
+			high_rtt_count++;
+
+			if (high_rtt_count > max_cons_high_rtt) {
+				ROS_WARN_NAMED("time", "TM : RTT too high for timesync: %0.2f ms", rtt_ns / 1000000.0);
+				// Reset counter to rate-limit warnings
+				high_rtt_count = 0;
+			}
```
vooon: ROS already have THROTTLEd logger. setting not needed.

mavros_msgs/msg/TimesyncStatus.msg
```diff
@@ -0,0 +1,7 @@
+# Status of the MAVLink time synchroniser
+
+std_msgs/Header header
+uint64 remote_timestamp		# remote system timestamp (nanoseconds)
+int64 observed_offset		# raw time offset directly observed from this timesync packet (nanoseconds)
+int64 estimated_offset		# smoothed time offset between companion system and Mavros (nanoseconds)
+float32 round_trip_time		# round trip time of this timesync packet (milliseconds)
```
vooon: Perhaps good idea to add `_ns/_ms` suffixes.

mavros/src/plugins/sys_time.cpp
mhkabir: Fixed, but seems too overkill for cog.

mavros_msgs/msg/TimesyncStatus.msg
mhkabir: Done!

mavros/src/plugins/sys_time.cpp
mhkabir: Should've made comment clearer. The intention is to not just rate-limit, but rather to inform the user when the RTT is high _n times in a row_. This would point to an issue with the system which needs to be fixed by the user.

mhkabir:
image

Good to go!

PR #1043, opened 2018-06-01
Add permanent UDP broadcasting for GCS bridge

okalachev:
For now, the GCS bridge has a «UDP broadcast» mode, but this name is quite confusing, because it’s not real broadcasting. In fact, it’s just an automatic GCS discovery. 

But, for some environments, simple UDP broadcasting is preferable. For example:

1. It allows viewing and controlling the vehicle from multiple GCS simultaneously (tested)
2. Sometimes, it better fits multiple vehicles communications purposes (e. g. drone - charging station - GCS, drones swarms, etc)
3. It can be faster over Wi-Fi connections, because broadcasting doesn’t have Wi-Fi level acks (unlike of unicasting).

The protocol name (udp-bb) is quite random, can be discussed.

### DIFF
--- a/libmavconn/README.md
+++ b/libmavconn/README.md
@@ -18,7 +18,8 @@ Supported schemas:
   - Serial: `serial:///path/to/serial/device[:baudrate][?ids=sysid,compid]`
   - Serial with hardware flow control: `serial-hwfc:///path/to/serial/device[:baudrate][?ids=sysid,compid]`
   - UDP: `udp://[bind_host][:port]@[remote_host][:port][/?ids=sysid,compid]`
-  - UDP broadcast: `udp-b://[bind_host][:port]@[:port][/?ids=sysid,compid]`
+  - UDP broadcast until GCS discovery: `udp-b://[bind_host][:port]@[:port][/?ids=sysid,compid]`
+  - UDP broadcast (permanent): `udp-pb://[bind_host][:port]@[:port][/?ids=sysid,compid]
   - TCP client: `tcp://[server_host][:port][/?ids=sysid,compid]`
   - TCP server: `tcp-l://[bind_port][:port][/?ids=sysid,compid]`
 
--- a/libmavconn/src/interface.cpp
+++ b/libmavconn/src/interface.cpp
@@ -284,7 +284,7 @@ static MAVConnInterface::Ptr url_parse_serial(
 
 static MAVConnInterface::Ptr url_parse_udp(
 		std::string hosts, std::string query,
-		uint8_t system_id, uint8_t component_id, bool is_udpb)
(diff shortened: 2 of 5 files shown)


libmavconn/src/udp.cpp
```diff
@@ -107,8 +107,13 @@ MAVConnUDP::MAVConnUDP(uint8_t system_id, uint8_t component_id,
 		socket.set_option(udps::send_buffer_size(256_KiB));
 		socket.set_option(udps::receive_buffer_size(512_KiB));
 
-		if (remote_host == BROADCAST_REMOTE_HOST)
+		if (remote_host == BROADCAST_REMOTE_HOST) {
```
vooon: Add to class initializer list `permanent_broadcast(false)`.

libmavconn/src/interface.cpp
```diff
@@ -392,6 +392,8 @@ MAVConnInterface::Ptr MAVConnInterface::open_url(std::string url,
 		return url_parse_udp(host, query, system_id, component_id, false);
 	else if (proto == "udp-b")
 		return url_parse_udp(host, query, system_id, component_id, true);
+	else if (proto == "udp-bb")
+		return url_parse_udp(host, query, system_id, component_id, true, true);
```
vooon: In that internal funcs better to be more explicit. I mean always pass permanent-broadcast.

libmavconn/README.md
```diff
@@ -18,7 +18,8 @@ Supported schemas:
   - Serial: `serial:///path/to/serial/device[:baudrate][?ids=sysid,compid]`
   - Serial with hardware flow control: `serial-hwfc:///path/to/serial/device[:baudrate][?ids=sysid,compid]`
   - UDP: `udp://[bind_host][:port]@[remote_host][:port][/?ids=sysid,compid]`
-  - UDP broadcast: `udp-b://[bind_host][:port]@[:port][/?ids=sysid,compid]`
+  - UDP broadcast until GCS discovery: `udp-b://[bind_host][:port]@[:port][/?ids=sysid,compid]`
+  - UDP broadcast (permanent): `udp-bb://[bind_host][:port]@[:port][/?ids=sysid,compid]
```
vooon: Maybe better name proto `udp-pb`?

libmavconn/src/interface.cpp
okalachev: I guess, the only explicit way to pass booleans is Pythonic `foo(bar=True, baz=False)`, sadly this is not Python :)

libmavconn/README.md
okalachev: OK.

okalachev:
Done.

vooon:
Thanks!

PR #1046, opened 2018-06-05
GCS bridge: pass only HEARTBEATs when GCS is offline

okalachev:
When using GCS bridge with UDP (or UDP broadcast) mode on Wi-Fi, even when QGC is off, it's a lot of traffic (MAVLink messages) making other Wi-Fi connections slow.

This PR adds a parameter `gcs_quiet_mode`, turning more intelligent mode: the bridge retransmits only `HEARTBEAT` messages before any other device (GCS) sends any message. After that, all telemetry is switched on. If the GCS is not sending anything for 20 seconds, telemetry is switched off again.

This has been tested with QGC and a PixHawk drone, and it works quite well.

### DIFF
--- a/mavros/src/lib/mavros.cpp
+++ b/mavros/src/lib/mavros.cpp
@@ -18,6 +18,7 @@
 
 // MAVLINK_VERSION string
 #include <mavlink/config.h>
+#include <mavconn/mavlink_dialect.h>
 
 using namespace mavros;
 using mavconn::MAVConnInterface;
@@ -32,6 +33,7 @@ MavRos::MavRos() :
 	fcu_link_diag("FCU connection"),
 	gcs_link_diag("GCS bridge"),
 	plugin_loader("mavros", "mavros::plugin::PluginBase"),
+	last_message_received_from_gcs(0),
 	plugin_subscriptions{}
 {
 	std::string fcu_url, gcs_url;
@@ -46,6 +48,8 @@ MavRos::MavRos() :
 
 	nh.param<std::string>("fcu_url", fcu_url, "serial:///dev/ttyACM0");
 	nh.param<std::string>("gcs_url", gcs_url, "udp://@");
+	nh.param<bool>("gcs_quiet_mode", gcs_quiet_mode, false);
+
 	nh.param<std::string>("fcu_protocol", fcu_protocol, "v2.0");
 	nh.param("system_id", system_id, 1);
 	nh.param<int>("component_id", component_id, mavconn::MAV_COMP_ID_UDP_BRIDGE);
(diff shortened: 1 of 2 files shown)


mavros/src/lib/mavros.cpp
```diff
@@ -135,8 +139,14 @@ MavRos::MavRos() :
 		mavlink_pub_cb(msg, framing);
 		plugin_route_cb(msg, framing);
 
-		if (gcs_link)
+		if (gcs_link) {
+			if (this->gcs_quiet_mode && msg->msgid != mavlink::common::msg::HEARTBEAT::MSG_ID &&
+				(ros::Time::now() - this->last_message_received_from_gcs > ros::Duration(20))) {
```
TSC21: The duration should be a parameter aswell.

okalachev: I'm not sure this should be a parameter, because it's too many parameters. @voon?

TSC21: If you are adding new features and adding new params for them, at least make the implementation flexible so the user can setup the feature itself without having to rebuild the code. Please have this changed.

mhkabir:
@okalachev Nice contribution! I was also thinking about implementing a "message filter" for GCS in the  `gcs_bridge`. For example, there is no need to have 200Hz IMU messages coming to mavros to be forwarded to GCS. The filtered messages could be read in from a YAML file like we do for the plugin blacklist. @vooon What do you think?

vooon:
@mhkabir that PR is more about saving link bandwidth when GCS not connected (or closed).

@TSC21 timeout parameter is not necessary, but we can use fcu timeout as well.

TSC21:
> @TSC21 timeout parameter is not necessary, but we can use fcu timeout as well.

It's a bad practice to hardcode this type of values bad that's your call. I have done my review.

okalachev:
@mhkabir , I'm experimenting with some more or less complicated multi-vehicle environments, I'm coming to a conclusion that there should be something like "MAVLink-Hub" program. It should transmit MAVLink message betweens "endpoints", with rich and configurable set of rules.

There exists https://github.com/intel/mavlink-router, but it lacks a lot of useful features:
* It can't transmit messages between endpoints, only between endpoints and master;
* filtering messages (blacklisting or whitelisting), limiting maximum rates;
* "quiet" mode, when only HEARTBEATs are transmitted;
* transmitting rules for different endpoints and systems.

I'm thinking about such a project.

okalachev:
@mhkabir , @anassinator , I've written some draft on how this software should work: https://gist.github.com/okalachev/5b6836999bcae2d4893b9773401a6759.

Can you comment, please?

Issue #1068, opened 2018-07-27
Adding RTK support (RTCM messages forward) 


AlexisTM:
### Issue details

To allow many copters to fly with the RTK, we need to forward the RTCM messages to all of them. We cannot have QGC running so we would like to send the RTCM messages from ROS.

https://github.com/AlexisTM/mavros/commit/6338bdef3359ff0dbfa7c2b3bbe6a889c7483dae

That specific message is handled differently in MAVLink v1 and v2 (zeroes removed). Is this implementation enough to have both working?

QGC Implementation: https://github.com/mavlink/qgroundcontrol/blob/master/src/GPS/RTCM/RTCMMavlink.cc

vooon:
No, you incorrectly implement fragmentation.

Message should looks following:
```
Header header
uint8[] data
```

Handler should:
1. Check data length, if it < message payload (180): goto 2 else 3
2. Copy data, set len, fill remaining with 0 (optional), set flags = (seq << 3) | 0;
3. Copy first 180 bytes to first message, set falgs = (seq << 3) | fragment << 1 | 1; repeat up to 4 times.

vooon:
Also you can use Header.seq as sequence counter, but appy `& 0x1f`.

vooon:
And data[] may vary up to 180*4 bytes. So add assert `m.data.size() <= 180*4`.

vooon:
Also i think better name plugin `navsat_rtk` or `gps_rtk`

AlexisTM:
I was thinking of a 3rd party node handling the fragmentation for fast prototyping.
But it makes indeed more sense is this is done on mavros side of the RTCM messages are up to 700 bytes

ArkadiuszNiemiec:
I have implemented a setup that configures an ublox RTK base, survey-in the position over set time and then sends RTCM3 messages over ROS2. On our vehicle, we have ROS2->ROS1 bridge and a ROS1 plugin that sends it to PX4 that handles the rest.
Please tell me if anything would be useful for you 😄

AlexisTM:
That would be awesome! Feel free to share it :)

ArkadiuszNiemiec:
@AlexisTM https://github.com/ArkadiuszNiemiec/rtcm_inject
...but I just realized that it is a very similar code that you have in your PR. The RTCM3 stream is published on ROS2 and bridged to ROS1 on both sides but it's on this repo. Do you need anything else?

AlexisTM:
[...]

PR #1069, opened 2018-07-27
GPS_RTK plugin to forward RTCM corrections

AlexisTM:
Followup of #1068 

This plugin allows forwarding RTCM messages from the RTK base station to the Pixhawk via mavros.

NOTE: This plugin still requires a node that publishes the RTCM data on ROS.


TSC21:
@AlexisTM I think this should be on `mavros_extras` as this is not a standard plugin people would be using.

AlexisTM:
The plugin can for sure be moved

TSC21:
> The plugin can for sure be moved 

Ok, please do it then.

mavros/src/plugins/gps_rtk.cpp
```diff
@@ -0,0 +1,86 @@
+/**
+ * @brief GPS RTK plugin
+ * @file gps_rtk.cpp
+ * @author Alexis Paques <alexis.paques@gmail.com>
+ *
+ * @addtogroup plugin
+ * @{
+ */
+/*
+ * Copyright 2018 Nuno Marques.
```
vooon: Update.

mavros/src/plugins/gps_rtk.cpp
```diff
@@ -0,0 +1,86 @@
...
+		return {};
+	}
+
+private:
+	ros::NodeHandle gps_rtk_nh;
+	ros::Subscriber gps_rtk_sub;
+
+	void rtk_cb(const mavros_msgs::RTCM::ConstPtr &msg)
+	{
+		const int maxMessageLength = 180;	// Should be replaced by MAVLINK_MSG_GPS_RTCM_DATA_FIELD_DATA_LEN
```
vooon: I think you can use `auto max_len = mavlink::common::msg::GPS_RTCM_DATA::data::size()` as it constexpr.

mavros/src/plugins/gps_rtk.cpp
```diff
@@ -0,0 +1,86 @@
...
+		if (msg->data.size() > 4 * maxMessageLength) {
+			ROS_FATAL("gps_rtk: RTCM message received is bigger than the maximal possible size");
+			return;
+		}
+
+		if (msg->data.size() < maxMessageLength) {
+			mavlink::common::msg::GPS_RTCM_DATA rtcm_data;
+			rtcm_data.len = msg->data.size();
+			rtcm_data.flags = (seq & 0x1F) << 3;
+			memcpy(&rtcm_data.data, &msg->data, msg->data.size());
```
vooon: Use `std::copy`. Something like that:

```cpp
std::copy(msg->data.begin(), msg->data.end(), rtcm_data.data.begin());
std::fill(rtcm_data.data.begin() + rtcm_data.len, rtcm_data.data.end(), 0);
```

mavros/src/plugins/gps_rtk.cpp
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+private:
+	ros::NodeHandle gps_rtk_nh;
+	ros::Subscriber gps_rtk_sub;
+
+	void rtk_cb(const mavros_msgs::RTCM::ConstPtr &msg)
+	{
+		const int maxMessageLength = 180;	// Should be replaced by MAVLINK_MSG_GPS_RTCM_DATA_FIELD_DATA_LEN
+		int seq = msg->header.seq;
```
vooon: I think better apply 0x1F mask here. And call it `seq_u5`, use uint8_t.

mavros/src/plugins/gps_rtk.cpp
vooon: Also i don't like CamelCase for anything except class names.

mavros/src/plugins/gps_rtk.cpp
```diff
@@ -0,0 +1,86 @@
...
+				mavlink::common::msg::GPS_RTCM_DATA rtcm_data;
+				int length = std::min((int)msg->data.size() - start, maxMessageLength);
+				rtcm_data.flags = 1;				// LSB set indicates message is fragmented
+				rtcm_data.flags |= fragmentId++ << 1;		// Next 2 bits are fragment id
+				rtcm_data.flags |= (seq & 0x1F) << 3;		// Next 5 bits are sequence id
+				rtcm_data.len = length;
+				memcpy(&rtcm_data.data, &msg->data + start, length);
+				UAS_FCU(m_uas)->send_message(rtcm_data);
+				start += length;
+			}
```
vooon: I think for is a bit better. Something like that:

```cpp
auto data_it = msg->data.begin();
auto end_it = msg->data.end();

for (uint8_t fragment_id = 0; fragment_id < 4 && data_it < end_it; fragment_id++) {
    size_t len = std::min(std::distance(data_t, end_it), MAX_FRAG_LEN);

    std::copy(data_it, data_it + len, rtcm_data.data.begin();
    std:fill(...);

    std::advance(data_it, len);
}
```

mavros_msgs/msg/RTCM.msg
```diff
@@ -0,0 +1,3 @@
+# RTCM message for the RTK plugin
+Header header
+uint8[] data
```
vooon: Add comment note, that data should be <= 4*180 = 720 bytes.

mavros/src/plugins/gps_rtk.cpp
AlexisTM: I tried to use `mavlink::common::msg::GPS_RTCM_DATA::data::size()` but it fails to compile: 

```bash
/home/alexis/Workspace/src/mavros/mavros/src/plugins/gps_rtk.cpp: In member function ‘void mavros::std_plugins::GpsRtkPlugin::rtk_cb(const ConstPtr&)’:
/home/alexis/Workspace/src/mavros/mavros/src/plugins/gps_rtk.cpp:52:53: error: ‘mavlink::common::msg::GPS_RTCM_DATA::data’ is not a class, namespace, or enumeration
   auto max_len = mavlink::common::msg::GPS_RTCM_DATA::data::size();
```

As data is a vector, I tried: `auto max_len = mavlink::common::msg::GPS_RTCM_DATA::data.size();` but it fails as it is not instanciated.

```bash
/home/alexis/Workspace/src/mavros/mavros/src/plugins/gps_rtk.cpp: In member function ‘void mavros::std_plugins::GpsRtkPlugin::rtk_cb(const ConstPtr&)’:
/home/alexis/Workspace/src/mavros/mavros/src/plugins/gps_rtk.cpp:52:55: error: invalid use of non-static data member ‘mavlink::common::msg::GPS_RTCM_DATA::data’
   auto max_len = mavlink::common::msg::GPS_RTCM_DATA::data.size();
                                                       ^
In file included from /opt/ros/kinetic/include/mavlink/v2.0/common/common.hpp:1310:0,
                 from /home/alexis/Workspace/src/mavros/libmavconn/include/mavconn/mavlink_dialect.h:26,
                 from /home/alexis/Workspace/src/mavros/libmavconn/include/mavconn/interface.h:37,
                 from /home/alexis/Workspace/src/mavros/mavros/include/mavros/mavros_plugin.h:24,
                 from /home/alexis/Workspace/src/mavros/mavros/src/plugins/gps_rtk.cpp:17:
/opt/ros/kinetic/include/mavlink/v2.0/common/./mavlink_msg_gps_rtcm_data.hpp:24:30: note: declared here
     std::array<uint8_t, 180> data; /*< RTCM message (may be fragmented) */
```

AlexisTM: It works as: 

```cpp
	mavlink::common::msg::GPS_RTCM_DATA rtcm_data;
	const int max_len = rtcm_data.data.size(); // 180 bytes
```

Yet, this would be useful if we are sending the message once; Here we could send it 4 times. 
I guess that  `UAS_FCU(m_uas)->send_message(rtcm_data);` is non blocking, so there could be a race condition issue if I use the same object. 

Is it safe to keep the same object and send it 4 times with different data?

AlexisTM:
@vooon Can I send 4 times a message from the same object with different data using `UAS_FCU(m_uas)->send_message(rtcm_data_frag);` safely or should I use different objects when the plugin fragments the message?

https://github.com/AlexisTM/mavros/blob/aa42b6d7a598dd26c89c9cc5bc086e98c2d15b2b/mavros/src/plugins/gps_rtk.cpp#L73-L74

mavros_extras/src/plugins/gps_rtk.cpp
```diff
@@ -0,0 +1,94 @@
...
+			rtcm_data.len = msg->data.size();
+			rtcm_data.flags = seq_u5;
+			std::copy(data_it, end_it, rtcm_data.data.begin());
+			std::fill(rtcm_data.data.begin() + rtcm_data.len, rtcm_data.data.end(), 0);
+			UAS_FCU(m_uas)->send_message(rtcm_data);
+		} else {
+			for (uint8_t fragment_id = 0; fragment_id < 4 && data_it < end_it; fragment_id++) {
+				mavlink::common::msg::GPS_RTCM_DATA rtcm_data_frag;	// Can we send rtcm_data 4 times with different data safely?
+
+				uint8_t len = std::min((int)std::distance(data_it, end_it), max_frag_len);
```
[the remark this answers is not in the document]
AlexisTM: It could not compile because `max_frag_len` was an int. Converting it to size_t allows removing the casting.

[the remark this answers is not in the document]
AlexisTM:
```bash
/home/alexis/Workspace/src/mavros/mavros_extras/src/plugins/gps_rtk.cpp:74:72: error: no matching function for call to ‘min(std::__iterator_traits<__gnu_cxx::__normal_iterator<const unsigned char*, std::vector<unsigned char, std::allocator<unsigned char> > >, void>::difference_type, const size_t&)’
     uint8_t len = std::min(std::distance(data_it, end_it), max_frag_len);
```

mavros_extras/CMakeLists.txt
```diff
@@ -78,6 +78,7 @@ add_library(mavros_extras
   src/plugins/vibration.cpp
   src/plugins/vision_pose_estimate.cpp
   src/plugins/vision_speed_estimate.cpp
+  src/plugins/gps_rtk.cpp
```
TSC21: can you please put this on the correct alphabetic order?

mavros_extras/mavros_plugins.xml
```diff
@@ -41,4 +41,7 @@
 	<class name="vision_speed_estimate" type="mavros::extra_plugins::VisionSpeedEstimatePlugin" base_class_type="mavros::plugin::PluginBase">
 		<description>Send vision speed estimate to FCU.</description>
 	</class>
+	<class name="gps_rtk" type="mavros::extra_plugins::GpsRtkPlugin" base_class_type="mavros::plugin::PluginBase">
```
TSC21: same as above.

mavros_extras/src/plugins/gps_rtk.cpp
```diff
@@ -0,0 +1,92 @@
...
+	Subscriptions get_subscriptions()
+	{
+		return {};
+	}
+
+private:
+	ros::NodeHandle gps_rtk_nh;
+	ros::Subscriber gps_rtk_sub;
+
+	void rtk_cb(const mavros_msgs::RTCM::ConstPtr &msg)
```
TSC21: can you please add `doxygen` compatible documentation to clarify the msg purpose and details? examples can be found in the `imu` plugin

TSC21: Also shouldn't be this called `rtcm_cb`? The reason I question this is because you actually have https://mavlink.io/en/messages/common.html#GPS_RTK, so this could actually have an handler by itself in this same plugin.

AlexisTM: This is right, I'll change it. 

This plugin should also be extended to subscribe to the GPS_RTK MAVLink message, maybe not in the first iteration.

AlexisTM:
Note: The GPS_RTK (and GPS2_RTK) MAVLink messages seems to be unused in the PX4 Firmware.

TSC21:
> Note: The GPS_RTK (and GPS2_RTK) MAVLink messages seems to be unused in the PX4 Firmware.

True, but I guess that is something that can be extended later.

TSC21:
@vooon we definitely have to review the CI system. I don't think it's good to have CI always failing, even if it has nothing to do with the code itself. CI needs to be clear and useful to us as maintainers, but it's currently something that is failing for whatever reasons. We should definitely bet in docker container-based CI.

vooon:
I have no idea why travis fail to install packages. Regarding ROS CI - i don't know why our tests often fail on them.
Travis with industrial-ci do not work at all.

vooon:
Build our own docker based ci scripts is not easy. I'll be happy if we can use something like gitlab-ci where docker comes out of the box.

TSC21:
> Build our own docker based ci scripts is not easy. I'll be happy if we can use something like gitlab-ci where docker comes out of the box.

It's something already done for PX4 CI. Can't see why we won't be able to do it the same way. We can also base our containers in base containers and add our own specific container depend on a base container. And that would actually allows us to build and test MAVROS over multiple ROS and OS versions.

TSC21:
@AlexisTM thanks for your clear and reasonable contribution! Looking forward for more!

AlexisTM:
You're welcome! I will test it as soon as I have a node getting RTCM data successfully. It's pretty hard to test as I don't have a proper location to place my base station.

JonasPerolini:
Good afternoon, did you end up testing the plugin? @AlexisTM

PR #1076, opened 2018-08-18
Add support for log transfer related messages

shylent:
This adds basic support for messages, that allow to implement efficient log transfer even for very large log files. These messages are used by QGroundControl to download log files, but there are times, when using QGC to download logs is inappropriate. For example, when one is conducting automated testing of UAVs (without human intervention).

The endpoints (services, topics) are called `raw/...`, because I intend to add "high-level" functionality, that will add convenient abstraction over this functionality. That abstraction, however, needs not be a part of this request.

### DIFF
--- a/libmavconn/CMakeLists.txt
+++ b/libmavconn/CMakeLists.txt
@@ -96,7 +96,7 @@ install(DIRECTORY cmake/Modules
 if(CATKIN_ENABLE_TESTING)
 
 catkin_add_gtest(mavconn-test test/test_mavconn.cpp)
-target_link_libraries(mavconn-test mavconn)
+target_link_libraries(mavconn-test mavconn pthread)
 
 endif()
 
--- a/mavros_extras/src/plugins/log_transfer.cpp
+++ b/mavros_extras/src/plugins/log_transfer.cpp
@@ -1,0 +1,125 @@
+#include <mavros/mavros_plugin.h>
+#include <mavros_msgs/LogData.h>
+#include <mavros_msgs/LogEntry.h>
+#include <mavros_msgs/LogRequestData.h>
+#include <mavros_msgs/LogRequestEnd.h>
+#include <mavros_msgs/LogRequestList.h>
+
+namespace mavros {
+namespace extra_plugins {
+class LogTransferPlugin : public plugin::PluginBase {
+public:
+	LogTransferPlugin() :
+		nh("~log_transfer") {}
+
+	void initialize(UAS& uas) override
+	{
+		PluginBase::initialize(uas);
+
+		log_entry_pub = nh.advertise<mavros_msgs::LogEntry>("raw/log_entry", 1000);
+		log_data_pub = nh.advertise<mavros_msgs::LogData>("raw/log_data", 1000);
+
+		log_request_list_srv = nh.advertiseService("raw/log_request_list",
+					&LogTransferPlugin::log_request_list_cb, this);
+		log_request_data_srv = nh.advertiseService("raw/log_request_data",
+					&LogTransferPlugin::log_request_data_cb, this);
+		log_request_end_srv = nh.advertiseService("raw/log_request_end",
+					&LogTransferPlugin::log_request_end_cb, this);
+	}
+
+	Subscriptions get_subscriptions() override
+	{
+		return {
+			       make_handler(&LogTransferPlugin::handle_log_entry),
+			       make_handler(&LogTransferPlugin::handle_log_data),
+		};
+	}
+
+private:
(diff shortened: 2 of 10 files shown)


mavros_extras/src/plugins/log_transfer.cpp
```diff
@@ -0,0 +1,125 @@
...
+	void handle_log_data(const mavlink::mavlink_message_t*, mavlink::common::msg::LOG_DATA& ld)
+	{
+		mavros_msgs::LogData msg;
+		msg.id = ld.id;
+		msg.offset = ld.ofs;
+
+		decltype(ld.data) ::size_type count(ld.count);
+		if (count > ld.data.max_size()) {
+			count = ld.data.max_size();
+		}
```
vooon: We prefer to use make_shared for ros messages.

Here `size_t count = std::min(ls.count, ls.data.max_size())`

mavros_extras/src/plugins/log_transfer.cpp
```diff
@@ -0,0 +1,125 @@
...
+		msg.data.insert(msg.data.cbegin(), ld.data.cbegin(), ld.data.cbegin() + count);
```
vooon: Why not `std::copy`?

mavros_extras/src/plugins/log_transfer.cpp
```diff
@@ -0,0 +1,125 @@
...
+		log_data_pub.publish(msg);
+	}
+
+	bool log_request_list_cb(mavros_msgs::LogRequestList::Request &req,
+				mavros_msgs::LogRequestList::Response &res)
+	{
+		mavlink::common::msg::LOG_REQUEST_LIST msg;
+		msg.target_system = m_uas->get_tgt_system();
+		msg.target_component = m_uas->get_tgt_component();
```
vooon: Use `UAS::msg_set_target()` instead.

mavros_msgs/msg/LogData.msg
```diff
@@ -0,0 +1,9 @@
+# Reply to LogRequestData, - a chunk of a log
+#
+#  :id: - log id
+#  :offset: - offset into the log
+#  :data: - chunk of data (if zero-sized, then there are no more chunks)
+
+uint16 id
+uint16 offset
+uint8[] data
```
vooon: We prefer to always have Header in messages.

libmavconn/CMakeLists.txt
```diff
@@ -96,7 +96,7 @@ install(DIRECTORY cmake/Modules
 if(CATKIN_ENABLE_TESTING)
 
 catkin_add_gtest(mavconn-test test/test_mavconn.cpp)
-target_link_libraries(mavconn-test mavconn)
+target_link_libraries(mavconn-test mavconn pthread)
```
vooon: I'm not sure that this is required change.

mavros_extras/src/plugins/log_transfer.cpp
shylent: because then I'll have to resize the destination as a separate step: msg.data is a std::vector which is initialized to be of zero size upon construction. std::vector::insert does this for me

mavros_extras/src/plugins/log_transfer.cpp
shylent: well, unfortunately, std::min does not compile. it could compile if we add a cast, but I'd prefer to avoid that even if it means writing extra couple of lines of code

re: make_shared - sure, I'll use boost::shared_ptr, no problem

mavros_msgs/msg/LogData.msg
shylent: done

mavros_extras/src/plugins/log_transfer.cpp
shylent: sure

shylent:
Thank you for the prompt response! I've implemented fixes according to your comments.

Regarding the usage of services: I feel, that in this case it is more appropriate to use services from the purely semantic point of view. A request for log entries or log data chunks does not fit well with the notion of a "topic" (message stream). As for the messages being lost in transit I could add a clarification in the doc for the srv file, so that it is clear what "success = True" actually means.

Of course, if this reasoning does not resonate with you and you insist that I use topics for LogRequestList / LogRequestData / LogRequestEnd, I have no choice, but to do as you say, but I feel, that in this case this will make the public API less clear.

shylent:
Thanks for the merge! I was intending to rebase --autosquash those "fixup!" commits before the merge, though. Hope, that is not a problem

PR #1077, opened 2018-08-20
setpoint_raw/target_attitude bodyrate commands should be converted to NED from ENU

Jaeyoung-Lim:
This PR fixes a coordinate frame mismatch of the setpoint_raw/target_attitude.

The current setpoint_raw/target_attitude body_rate commands directly sends the bodyrate commands from the ROS coordinate system(ENU) without converting it to NED->roll pitch yaw rate commands.

### DIFF
--- a/mavros/src/plugins/setpoint_raw.cpp
+++ b/mavros/src/plugins/setpoint_raw.cpp
@@ -217,18 +217,18 @@ private:
 	void attitude_cb(const mavros_msgs::AttitudeTarget::ConstPtr &req)
 	{
 		Eigen::Quaterniond desired_orientation;
-		Eigen::Vector3d baselink_angular_rate;
+		Eigen::Vector3d desired_angular_rate;
 
 		tf::quaternionMsgToEigen(req->orientation, desired_orientation);
+
+		tf::vectorMsgToEigen(req->body_rate, desired_angular_rate);
 
 		// Transform desired orientation to represent aircraft->NED,
 		// MAVROS operates on orientation of base_link->ENU
 		auto ned_desired_orientation = ftf::transform_orientation_enu_ned(
 					ftf::transform_orientation_baselink_aircraft(desired_orientation));
 
-		auto body_rate = ftf::transform_frame_baselink_aircraft(baselink_angular_rate);
-
-		tf::vectorMsgToEigen(req->body_rate, body_rate);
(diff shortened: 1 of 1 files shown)


AlexisTM:
@Jaeyoung-Lim 
The body frame in ENU is: Front Left Up (FLU)
The body frame in NED is: Front Right Down (FRD)

Can you first verify that the transformation is wrong? This went through many many discussions.

Validity check: 
* ROS side body_rate is `0.5 rad/s` then the MAVLink message will become `-0.5 rad/s`
* Incomming body_rate from MAVLink of `0.5 rad/s` will become `-0.5 rad/s` in ROS side

amilcarlucas:
This needs a rebase

TSC21:
@amilcarlucas I think it's more relevant to understand if this is a requires change or not than to actually rebase the PR.

Jaeyoung-Lim:
@TSC21 @AlexisTM 
My impression was that everything on the ROS message side should be represented as ENU, as this is the standard in ROS.

I assumed the bodyframe should be defined as ENU too, hence the bodyrate defined as ENU. If this was done intentionally, this is from my confusion on how the body frame should be defined.

TSC21:
Check this: http://www.ros.org/reps/rep-0103.html#id21

Jaeyoung-Lim:
@TSC21 I have checked the document.

It is just counter intuitive that all the other setpoints are in ENU frame and only the bodyrates are in NED frame.

TSC21:
@Jaeyoung-Lim the body rates are in the body frame, and that means they are aligned with the vehicle frame, not the loca/world frame. The local frame is Earth-aligned, and in ROS, uses the ENU orientation. The body frame, which is vehicle aligned, uses the FLU convention in ROS, while in the aircraft convention, you use the FRD convention. Still though, there needs to be a straight separation between body frame and local frame in terms of orientation. The body rates are vehicle-aligned, not Earth-aligned, so you don't use NED/ENU convention in that case.

TSC21:
@Jaeyoung-Lim also you can check that the velocity setpoints can also be in the local or in the body frame, and you apply different transforms respectively.

Jaeyoung-Lim:
@TSC21 Thanks for the explanation. I do understand how the coordinate systems are defined. 

If we want to maintain the definition of body_frame as what you have explained, I think this PR is not needed.

PR #1079, opened 2018-09-06
Fix odom message to use covariance from msg

dgonano:
Was testing the odom message being pulled into PX4, noticed my covariances weren't being passed through (they were just empty zero matrix instead). 

I'm assuming it's better to actually use the covars or is there a reason to override it to zeros?

### DIFF
--- a/mavros_extras/src/plugins/odom.cpp
+++ b/mavros_extras/src/plugins/odom.cpp
@@ -144,11 +144,11 @@ private:
 		}
 
 		//! Build 6x6 pose covariance matrix to be transformed and sent
-		ftf::Covariance6d cov_pose {};	// zero initialized
+		ftf::Covariance6d cov_pose = odom->pose.covariance;
 		ftf::EigenMapCovariance6d cov_pose_map(cov_pose.data());
 
 		//! Build 6x6 velocity covariance matrix to be transformed and sent
-		ftf::Covariance6d cov_vel {};	// zero initialized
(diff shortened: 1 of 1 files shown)


TSC21:
No specific reason. I actually thought this was being passed but I didn't test it with the plugin but rather with gazebo directly. Is this working for you as expected?

TSC21:
Merging now. Thanks. If you have some results, would be great if you post them here so they can be evaluated.

dgonano:
Turns out I'm not actually going to use the odom message as I found another way using minor changes to the released 1.8.0 PX4. (PR opened in PX4/ecl) but this other way might still be useful to this odom method. 

Just thought I'd do the PR here for this seeing as noticed the issue in my testing.

dgonano:
Just to clarify, the reason the odom wasn't working for me was still to do with covariance handling and things not being passed through to the EKF. So I ended up just modifying how things were passed through in 1.8.0
https://github.com/PX4/ecl/pull/507

TSC21:
This will endup being the std way of feeding odometry data through Mavlink to the FCU as soon as the PX4 PR gets in. No other implementation on PX4 firmware uses the covariance values from an external estimation.

TSC21:
If the idea is to update the ecl interface, please do all the suggestions based on the odometry interface of the PR I have to PX4 firmware, since in the end that will be the end approach for odometey handling on PX4.

PR #1092, opened 2018-10-10
sys_status : add MAV_TYPE as a parameter

Kiwa21:
add MAV_TYPE as a parameter that can be changed with the yaml config file.

### DIFF
--- a/mavros/launch/apm_config.yaml
+++ b/mavros/launch/apm_config.yaml
@@ -8,6 +8,7 @@ startup_px4_usb_quirk: false
 # sys_status & sys_time connection options
 conn:
   heartbeat_rate: 1.0    # send hertbeat rate in Hertz
+  heartbeat_mav_type: "Onboard companion controller" # mav type as the mavlink description
   timeout: 10.0          # hertbeat timeout in seconds
   timesync_rate: 0.0     # TIMESYNC rate in Hertz (feature disabled if 0.0)
   system_time_rate: 0.0  # send system time to FCU rate in Hertz (disabled if 0.0)
--- a/mavros/src/lib/enum_to_string.cpp
+++ b/mavros/src/lib/enum_to_string.cpp
@@ -152,8 +152,8 @@ static const std::array<const std::string, 33> mav_type_strings{{
 /* 16 */ "Flapping wing",                 // Flapping wing
 /* 17 */ "Kite",                          // Kite
 /* 18 */ "Onboard companion controller",  // Onboard companion controller
-/* 19 */ "Two",                           // Two-rotor VTOL using control surfaces in vertical operation in addition. Tailsitter.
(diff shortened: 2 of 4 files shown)


mavros/launch/apm_config.yaml
```diff
@@ -8,6 +8,7 @@ startup_px4_usb_quirk: false
 # sys_status & sys_time connection options
 conn:
   heartbeat_rate: 1.0    # send hertbeat rate in Hertz
+  conn_heartbeat_mav_type: 18 # ONBOARD_CONTROLLER
```
vooon: Better to extend uas lib to be able to find MAV_TYPE by string.

mavros/src/plugins/sys_status.cpp
```diff
@@ -512,6 +519,9 @@ class SystemStatusPlugin : public plugin::PluginBase
 	ros::ServiceServer rate_srv;
 	ros::ServiceServer mode_srv;
 
+	int conn_heartbeat_mav_type;
```
vooon: Should be `MAV_TYPE`, or at least underlying type for that enum.

Kiwa21:
@vooon Thanks for suggestion. I updated my PR to use MAV_TYPE and find by string.

mavros/src/lib/enum_to_string.cpp
```diff
@@ -176,6 +176,20 @@ std::string to_string(MAV_TYPE e)
...
+			std::underlying_type<MAV_TYPE>::type rv = idx;
+			return static_cast<MAV_TYPE>(rv);
+		}
+	}
+
+	ROS_ERROR_STREAM_NAMED("uas", "TYPE: Unknown MAV_TYPE: " << mav_type);
+	return MAV_TYPE::GENERIC;
+}
+
 // [[[end]]] (checksum: 31488f5970b0f82b3efef71e32590bb6)
```
vooon: Do not change generated code.

Also i think it is better to add enum entry name to value function.
For that another array is needed, or changes in current (use std::tuple).

mavros/src/plugins/sys_status.cpp
```diff
@@ -430,16 +431,23 @@ class SystemStatusPlugin : public plugin::PluginBase
...
 		nh.param("sys/disable_diag", disable_diag, false);
 
-		// rate parameter
+		// heartbeat rate parameter
 		if (nh.getParam("conn/heartbeat_rate", conn_heartbeat_d) && conn_heartbeat_d != 0.0) {
 			conn_heartbeat = ros::Duration(ros::Rate(conn_heartbeat_d));
 		}
 
+		// heartbeat mav type parameter
+		if (nh.getParam("conn/heartbeat_mav_type", conn_heartbeat_mav_type_str) && !conn_heartbeat_mav_type_str.empty()) {
```
vooon: `.empty()` check is not necessary. Heartbeat rate check is needed because we can set 0 to disable heartbeats.

PR #1096, opened 2018-10-14
udp: bind should be called after reuse_address is set

### LINKED ISSUE #1095: UDP connection fails when running under Windows Subsystem for Linux (WSL)
### Issue details

I realize that ROS is not officially supported on Windows, but in my use case it is mostly working fine on Windows 10 with the [Windows Subsystem for Linux](https://en.wikipedia.org/wiki/Windows_Subsystem_for_Linux) (WSL) compatibility layer. See [instructions](http://wiki.ros.org/Installation/Windows) on the ROS Wiki. I did encounter an issue with mavros under the WSL which can be fixed easily.

When attempting to run mavros with the udp:// option I get the following error:

```
$ rosrun mavros mavros_node _fcu_url:=udp://@
[ INFO] [1539362177.089738700]: FCU URL: udp://@
[ INFO] [1539362177.091763900]: udp0: Bind address: 0.0.0.0:14555
[FATAL] [1539362177.092218600]: FCU: DeviceError:udp:set_option: Invalid argument
```

Looks like the error originates in `libmavconn/src/udp.cpp` and is caused by calling `socket.set_option(udps::reuse_address(true))` after `socket.bind(bind_ep)`.

Note that currently setting `reuse_address` does not work (even on Linux) because SO_REUSEADDR needs to be set before calling `bind()`.

I can suggest 3 options for fixing this, all tested:

1. Move the `reuse_address` line before the `bind` line.
2. Remove the `reuse_address` line as it is not working anyway.
3. Replace the call with something like the following code and ignore any errors (less risky but I am not recommending this option).
```
boost::system::error_code ec;
socket.set_option(udps::reuse_address(true), ec); // Ignore the error
socket.set_option(udps::send_buffer_size(256_KiB));
```

I am not submitting a pull request as there are several options to fix this, but let me know if I should.

Thanks!

Mavros: 0.26.3 and also HEAD (built from source)
ROS: Kinetic
Ubuntu: 16.04

kshahar:
Fixes #1095

### DIFF
--- a/libmavconn/src/udp.cpp
+++ b/libmavconn/src/udp.cpp
@@ -101,12 +101,13 @@ MAVConnUDP::MAVConnUDP(uint8_t system_id, uint8_t component_id,
 
 	try {
 		socket.open(udp::v4());
-		socket.bind(bind_ep);
 
 		// set buffer opt. size from QGC
 		socket.set_option(udps::reuse_address(true));
 		socket.set_option(udps::send_buffer_size(256_KiB));
 		socket.set_option(udps::receive_buffer_size(512_KiB));
+
+		socket.bind(bind_ep);
 
 		if (remote_host == BROADCAST_REMOTE_HOST) {
 			socket.set_option(udps::broadcast(true));


khancyr:
LGTM

PR #1102, opened 2018-10-22
add ability to receive vision pose from FCU

EliaTarasov:
This PR adds handler for `mavlink_vision_position_estimate` msg to be received from FCU.

### DIFF
--- a/mavros_extras/src/plugins/vision_pose_estimate.cpp
+++ b/mavros_extras/src/plugins/vision_pose_estimate.cpp
@@ -59,12 +59,15 @@ public:
 		else {
 			vision_sub = sp_nh.subscribe("pose", 10, &VisionPoseEstimatePlugin::vision_cb, this);
 			vision_cov_sub = sp_nh.subscribe("pose_cov", 10, &VisionPoseEstimatePlugin::vision_cov_cb, this);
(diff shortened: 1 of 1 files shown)


EliaTarasov:
@TSC21 the point is the same as for px4flow plugin. Vision data source may be attached to px4 directly, but in the same time you may need to handle it in ROS.

TSC21:
> @TSC21 the point is the same as for px4flow plugin. Vision data source may be attached to px4 directly, but in the same time you may need to handle it in ROS.

Can you give a practical example where that already happens? I don't think that's the case. PX4 doesn't process (directly) any kind of vision source. This is usually done by an onboard computer, that **feeds** the data into the PX4 internals, usually using MAVROS or any other SDK capable of doing that. The idea of the `VISION_POSITION_ESTIMATE` is to be consumed, not to be used as feedback. The PX4Flow case is different, since the flow data is computed on the module itself, and that data can be propagated to the Mavlink stream so to be used on the onboard computer (not a typical usage since you can also use a serial interface for that and use `rosserial`), or even for debugging purposes. The MAVROS side usually runs on the OBC, and it's the OBC that produces the data (after CV algorithms processing pipelines) to be sent and consumed **on the FCU**.

EliaTarasov:
@TSC21 agree. Let's close this.

PR #1103, opened 2018-10-22
Fix of trigger setup and control via mavlink cmd interface 

### LINKED ISSUE #987: Trigger cycle_time cannot be changed
### Issue details

I tried to use "~/cmd/trigger_control" to control the trigger of pixracer, by calling service 
```
srv.request.cycle_time = (1000/framerate_hz);
srv.request.trigger_enable = true;
```
while setting the pixracer to `time based, on command` with `GPIO`.

This method can enable/disable camera trigger without problem, but the trigger rate is always 25 Hz regardless what I set in the cycle_time field.

Mavros: 0.23.0
ROS: Kinetic
Ubuntu: 16.04

[  ] ArduPilot
[ X] PX4

Version: 1.7.3

m0oz:
Adapt trigger_control service to current mavlink cmd spec. Add a new service to change trigger interval and integration time.

Fixes #987
@mhkabir

### DIFF
--- a/mavros/src/mavros/command.py
+++ b/mavros/src/mavros/command.py
@@ -20,7 +20,7 @@ __all__ = (
 import rospy
 import mavros
 
-from mavros_msgs.srv import CommandLong, CommandInt, CommandBool, CommandHome, CommandTOL, CommandTriggerControl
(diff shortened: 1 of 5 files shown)


TSC21:
The CI on MAVROS has reached a completely broken state. It definitely needs a fix and probably a complete restructure.

m0oz:
When will this be merged?

PR #1111, opened 2018-10-31
with this fix ,it will avoid eigen error on 32 bits system

GaoGeolone:
The issue and the solution I have reported here#1110 .
Here is the flexible way to avoid the eigen not perfectly aligned issue on 32 bits system.If you libeigen support SSE packet, you will need this fix on your board.

[http://eigen.tuxfamily.org/dox-devel/group__TopicStructHavingEigenMembers.html](url)here explain why this fix is needed, and why it can also fit X86-X64system.

In addition,the version is pasted here :ROS distro is above Lunar ,and GCC is above 5.4 and libeigen3 is above 3.3.4-4

### DIFF
--- a/mavros/src/plugins/hil.cpp
+++ b/mavros/src/plugins/hil.cpp
@@ -42,7 +42,7 @@ public:
 	HilPlugin() : PluginBase(),
 		hil_nh("~hil")
 	{ }
-
+	EIGEN_MAKE_ALIGNED_OPERATOR_NEW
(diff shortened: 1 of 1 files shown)


GaoGeolone:
I search the problem in check log ,I think maybe the conflict comes from the test_mavros, I am confused.

vooon:
Don't worry about CI failures. Unfortunately Travis now are broken. ROS CI most time fails on libmavconn test.

PR #1112, opened 2018-11-06
fix: a typing error "alredy" to "already"

rapsealk:


rapsealk:
I just wonder why the checks have failed.

vooon:
Travis was broken after some it's update, our code no longer works => Failure.
On ROS CI we often have problem with libmavconn test, which i can't reproduce...

Issue #450, opened 2015-12-07
Plugins for GCS Bridge


mhkabir:
This would enable (but not limited to) : 
* Receiving companion computer shutdown/reboot commands from GCS
* CPU load and other status reporting to GCS
* Reporting of link status for links connected to the companion (e.g over Ethernet, or WiFi dongles, etc.)

TSC21:
Also interested. @mhkabir maybe we can check this together later.

Issue #333, opened 2015-07-07
Support OBC reboot/shutdown using GCS commands


mhkabir:
@LorenzMeier @vooon 
We need to support commands from GCS for shutting down or rebooting the OBC. Currently handling of commands from GCS is rather difficult, since mavros has no API for that.

This is a system-critical feature, as I have managed to do disk damage by improper shutdown of OBC.

@dogmaphobic interested as well :)

vooon:
That task require processing gcs commands. So currently there no other choice than process it in gcs_bridge.

Another difficulties that halt requires root privileges, and i can't recommend to run bridge as root.
So extra step for sudoers config.

vooon:
Or you think that gcs bridge also need plugin system?

vooon:
Another option is to add somehow "power" button to OBC and setup OS to poweroff on long tap?

LorenzMeier:
I would think that ROS has already an API to stop or reboot the machine - can we just use this / send a command to it?

mhkabir:
@vooon  
I did a sudoers config for my OBC for this very purpose, and I think it's easy enough for most users to pull off.
I think a plugin system for the GCS bridge would be very welcome, especially as we are expanding our capabilities. It would be useful for e.g sending back datalink quality data, images (already done in custom bridge) and handling the GCS commands meant only for the OBC.

@LorenzMeier ROS doesn't have a proper API for controlling the system power states. I can see vague references to custom scripts for the PR2 robot which allow this, but cannot find any actual code.

dogmaphobic:
> @dogmaphobic interested as well :)

Sure, if I lived in Venus maybe. At the least the days would be long enough to have time to do everything I want to do :)

Receiving data directly from GCS opens the door to a whole new level of interaction though. Not just the ability to shut it down.

mhkabir:
@vooon I think this is mostly done by event_launcher. What else needs to be done?

vooon:
Perhaps in 0.16 we may introduce plugins for gcs_bridge. But that means that internal bridge go away.

In case of that issue you need node that listens to `/mavlink/to` and interpret `COMMAND_LONG`/`COMMAND_INT` and trigger event_launcher to shutdown/reboot.

mhkabir:
@vooon Any updates on this? Would you mind guiding me on how we can achieve this? :) Or could you maybe implement the plugin system for GCS bridge?

vooon:
[...]

PR #1117, opened 2018-11-12
GCS Bridge Plugin

arwagoner:
Converts the GCS Bridge into a plugin. 

Tested it by writing a node that publishes a test `NAMED_VALUE_FLOAT` to `/mavlink/from` and saw that it shows up on QGC. Users do have to specify a `gcs_url` in their mavros launch file to enable this plugin. 

qgc_bridge_plugin

It does not currently support OBC reboot/shutdown commands from GCS as requested in https://github.com/mavlink/mavros/issues/333 and https://github.com/mavlink/mavros/issues/450.

### DIFF
--- a/mavros/src/lib/mavros.cpp
+++ b/mavros/src/lib/mavros.cpp
@@ -101,15 +101,15 @@ MavRos::MavRos() :
 		try {
 			gcs_link = MAVConnInterface::open_url(gcs_url, system_id, component_id);
 
-			gcs_link_diag.set_mavconn(gcs_link);
-			gcs_diag_updater.setHardwareID(gcs_url);
-			gcs_diag_updater.add(gcs_link_diag);
+            gcs_link_diag.set_mavconn(gcs_link);
+            gcs_diag_updater.setHardwareID(gcs_url);
+            gcs_diag_updater.add(gcs_link_diag);
             UAS_GCS(&mav_uas) = gcs_link;
 		}
 		catch (mavconn::DeviceError &ex) {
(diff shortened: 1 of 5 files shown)


mavros/src/lib/mavros.cpp
```diff
@@ -96,23 +96,24 @@ MavRos::MavRos() :
...
-		}
-		catch (mavconn::DeviceError &ex) {
-			ROS_FATAL("GCS: %s", ex.what());
-			ros::shutdown();
-			return;
-		}
-	}
-	else
-		ROS_INFO("GCS bridge disabled");
+    if (gcs_url != "") {
```
TSC21: Indentation are 8 spaces, not 4. Please use uncrustify (`uncrustify -c ${ROS_WORKSPACE}/src/mavros/mavros/tools/uncrustify-cpp.cfg --replace --no-backup <path/to/file.ext>`).

arwagoner:
> Where changes in MavRos class? 

I added a new pointer for the GCS link. It follows the same format as the FCU link

> How that new callbacks works?

It works the same way as the callback worked in the original GCS_Bridge. The only difference is that this is a plugin.

> Where message filter?

What kind of a message filter are you looking for? 

> For now that is early WIP. Also i'm thinking that GCS plugin should have different base class to differentiate message sources.

There is not currently a good way to communicate between the companion computer and QGC. I think it makes sense to get the base functionality out there and then allow the community to continue building on it as it goes.

vooon:
That feature woun't be implemented in mavros v1. But it should be possible with v2 because here we have filters and router.

PR #1118, opened 2018-11-14
[WIP] Avoidance heartbeat

baumanta:
Send out a heartbeat from the avoidance process to the fcu. This allows monitoring the state of the individual and required process running on the companion computer.

As there exists no MAV_TYPE for such processes yet, the MAV_TYPE_SUBMARINE, which is not in use for PX4, is misused.


baumanta:
@TSC21 @thomasgubler @bkueng  FYI

thomasgubler:
@TSC21 could you help here to figure out how it would be possible to use component id instead of mav_type?

vooon:
@thomasgubler it is impossible to sent different component_id with current libmavconn.
All plugins shares same address.

In theory you can extend send_message, because MsgBuffer accepts ids, but you also need to pass different status struct. And that is main problem.

mavros_extras/src/plugins/companion_process_status.cpp
```diff
@@ -0,0 +1,85 @@
...
+	 * @param req	received CompanionProcessStatus msg
+	 */
+
+
+
+	void status_cb(const mavros_msgs::CompanionProcessStatus::ConstPtr &req)
+	{
+		mavlink::common::msg::HEARTBEAT heartbeat {};
+
+		heartbeat.type = 12;					//enum="MAV_TYPE" Type of the MAV (quadrotor, helicopter, etc.) missuse filed for submarine
```
hamishwillee: Why not create a new appropriate type. If you can come up with a good name that can be accepted in MAVLink fairly easily. Mabye MAV_TYPE_PROCESS_MONITOR

baumanta:
should we leave it in the MAV_TYPE? I could open a PR on mavlink for a MAV_TYPE_AVOIDANCE or such. What do you think? @vooon @hamishwillee

vooon:
@baumanta i think first we need answers on: https://github.com/mavlink/mavlink/pull/1022

baumanta:
@hamishwillee @vooon  
so what would you suggest we do? It seems that we cannot use component_id as @vooon explained. But we should net use MAV_TYPE. So does that mean the HEARTBEAT message is unusable for this case? Are we back at the custom message?

hamishwillee:
@baumanta I *think* you use a https://mavlink.io/en/messages/common.html#MAV_COMPONENT

mavros_msgs/msg/CompanionProcessStatus.msg
```diff
@@ -0,0 +1,59 @@
...
+uint8 MAV_STATE_BOOT = 1
+uint8 MAV_STATE_CALIBRATING = 2
+uint8 MAV_STATE_STANDBY = 3
+uint8 MAV_STATE_ACTIVE = 4
+uint8 MAV_STATE_CRITICAL = 5
+uint8 MAV_STATE_EMERGENCY = 6
+uint8 MAV_STATE_POWEROFF = 7
+uint8 MAV_STATE_FLIGHT_TERMINATION = 8
+
+uint8 MAV_COMP_ID_ALL = 0
```
TSC21: @baumanta we don't need all of these enumeration entries, since the only `MAV_COMP`  we require are the ones used for the companion processes.

mavros_extras/src/plugins/companion_process_status.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+
+
+	void status_cb(const mavros_msgs::CompanionProcessStatus::ConstPtr &req)
+	{
+		mavlink::common::msg::HEARTBEAT heartbeat {};
+
+		heartbeat.type = enum_value(mavlink::common::MAV_TYPE::ONBOARD_CONTROLLER);
+		heartbeat.autopilot = enum_value(mavlink::common::MAV_AUTOPILOT::PX4);
+		heartbeat.base_mode	= enum_value(mavlink::common::MAV_MODE_FLAG::TEST_ENABLED);
```
TSC21: Replace tab with whitespace?

mavros_extras/src/plugins/companion_process_status.cpp
```diff
@@ -0,0 +1,85 @@
...
+	ros::Subscriber status_sub;
+
+	/**
+	 * @brief Send companion process status to FCU
+	 *
+	 * Message specification: http://mavlink.org/messages/common#COMPANION_PROCESS_STATUS
+	 * @param req	received CompanionProcessStatus msg
+	 */
+
+
```
TSC21: Why these many whitespaces?

mavros_extras/src/plugins/companion_process_status.cpp
TSC21: Also another thing: from the description of , *0b00000010 system has a test mode enabled. This flag is intended for temporary system tests and should not be used for stable implementations*. I do think it's more reasonable to use `MAV_MODE_FLAG_CUSTOM_MODE_ENABLED` on this case.

mavros_extras/src/plugins/companion_process_status.cpp
```diff
@@ -0,0 +1,85 @@
...
+
```
TSC21: Also can you please add doxygen type docs for the functions? Example: https://github.com/mavlink/mavros/blob/master/mavros_extras/src/plugins/trajectory.cpp#L194-L199

mavros_extras/src/plugins/companion_process_status.cpp
```diff
@@ -0,0 +1,85 @@
...
+	void status_cb(const mavros_msgs::CompanionProcessStatus::ConstPtr &req)
+	{
+		mavlink::common::msg::HEARTBEAT heartbeat {};
+
+		heartbeat.type = enum_value(mavlink::common::MAV_TYPE::ONBOARD_CONTROLLER);
+		heartbeat.autopilot = enum_value(mavlink::common::MAV_AUTOPILOT::PX4);
+		heartbeat.base_mode	= enum_value(mavlink::common::MAV_MODE_FLAG::TEST_ENABLED);
+		heartbeat.system_status = req->state;	//enum="MAV_STATE" System status flag
+
+		ROS_DEBUG_STREAM_NAMED("companion_process_status", "Companion process status: " << utils::to_string_enum<MAV_STATE>(heartbeat.system_status)
```
TSC21: I think it's also reasonable to add the component ID of that process here. For that, we also need to add the component ID enum in https://github.com/mavlink/mavros/blob/master/mavros/src/lib/enum_to_string.cpp.

mavros_msgs/msg/CompanionProcessStatus.msg
```diff
@@ -0,0 +1,59 @@
+# Mavros message: COMPANIONPROCESSSTATUS
+
+std_msgs/Header header
+
+uint8 state        		# See enum COMPANION_PROCESS_STATE
```
TSC21: Can we match the ident here? Just to look cleaner :)

mavros_extras/src/plugins/companion_process_status.cpp
baumanta: @TSC21 I think in the plugin they already exist. where would you want those added?

mavros/src/lib/enum_to_string.cpp
```diff
@@ -213,6 +214,56 @@ std::string to_string(MAV_STATE e)
 
 	return mav_state_strings[idx];
 }
+
```
TSC21: @baumanta this is not right. The `to_string` function needs to be generated using `cog`. Check how the other enumerations are generated. You can also see with this add, you brake the checksum of line https://github.com/mavlink/mavros/pull/1118/files#diff-622b1c80b8e897ac3052c0d8b53ba2ddR267. You don't have to add this manually, you just need to let cog do it's job: https://github.com/mavlink/mavros/blob/master/CONTRIBUTING.md#cogpy-generators

mavros_extras/src/plugins/companion_process_status.cpp
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+private:
+	ros::NodeHandle status_nh;
+	ros::Subscriber status_sub;
+
+	/**
+	 * @brief Send companion process status to FCU over a heartbeat message
+	 *
+	 * Message specification: http://mavlink.org/messages/common#HEARTBEAT
```
hamishwillee: If this existed, the URL would be https://mavlink.io/en/messages/common.html#COMPANION_PROCESS_STATUS (ie not mavlink.org). But it does not exist.

baumanta: Hmm I had that first like that when we used our custom message. But when we decided to go for using the HEARTBEAT mavlink message instead of adding a custom one I thought it would be better to put that there instead. Is there a problem with that, should we change it?

TSC21: @baumanta the problem is not the msg name (which is ok). The problem is the link: it should be https://mavlink.io/en/messages/common.html#HEARTBEAT

hamishwillee: Yes, sorry, copy paste error. @TSC21 correct - mavlink.org is now defunct. Redirects were supposed to happen for messages but never did :-(

TSC21: @hamishwillee yeah this needs to be fixed, not only here, but in all other plugins with references to the message docs.

hamishwillee: Yes it sux. Probably not your highest priority to do globally, but worth fixing as we come to them.

baumanta: @hamishwillee @TSC21 Do I understand this right, we would just need to change all links to direct to mavlink.io instead of mavlink.org? But otherwise they are correct, no?

TSC21: Yes correct

hamishwillee: > But otherwise they are correct, no?

Yes, but note it isn't just a matter of changing the site address from .org to .io - there are other differences in the URI.

PR #1120, opened 2018-11-16
Updated ActuatorControl msg for direct motor control.

Pedro-Roque:
Adds direct control for the motors functionality, enabling direct PWM drive through mavros to actuators main outputs, by bypassing mixer settings. 

Related with: 
https://github.com/PX4/Firmware/pull/10863

### DIFF
--- a/mavros_msgs/msg/ActuatorControl.msg
+++ b/mavros_msgs/msg/ActuatorControl.msg
@@ -10,6 +10,7 @@ uint8 PX4_MIX_PAYLOAD = 2
 uint8 PX4_MIX_MANUAL_PASSTHROUGH = 3
 #uint8 PX4_MIX_FC_MC_VIRT = 4
 #uint8 PX4_MIX_FC_FW_VIRT = 5
+uint8 PX4_DIRECT_MOTOR_CONTROL = 99
 
 std_msgs/Header header
 uint8 group_mix


mavros_msgs/msg/ActuatorControl.msg
```diff
@@ -10,6 +10,7 @@ uint8 PX4_MIX_PAYLOAD = 2
 uint8 PX4_MIX_MANUAL_PASSTHROUGH = 3
 #uint8 PX4_MIX_FC_MC_VIRT = 4
 #uint8 PX4_MIX_FC_FW_VIRT = 5
+uint8 PX4_DIRECT_MOTOR_CONTROL = 99
```
vooon: Perhaps better to keep `PX4_MIX_` prefix.

```suggestion
uint8 PX4_MIX_DIRECT_MOTOR_CONTROL = 99
```

Pedro-Roque: Well it really doesn't go through the mixer, thats why I kept it without the _MIX_ . 

Yes, I also was puzzled with that, because on the mavlink its group_mlx as well.

Pedro-Roque:
@vooon is this good to merge or shall I do something else?

TSC21:
@Pedro-Roque this only makes sense to bring in if https://github.com/PX4/Firmware/pull/10863 is merged first.

Pedro-Roque:
> @Pedro-Roque this only makes sense to bring in if [PX4/Firmware#10863](https://github.com/PX4/Firmware/pull/10863) is merged first.

Yep makes sense. I'll join the dev call today to see if we get that going :)

willcbaker:
+1,  any progress?

Pedro-Roque:
@vooon @TSC21 I will close this PR as we are going through an easier, more modular implementation instead of relaying on the Actuator Control msg.

PR #1121, opened 2018-11-16
Added RPYrT and uncrustified.

Pedro-Roque:
Tested in SITL and flight arena with an Iris drone. 

Changes to PX4 Firmware were also needed, PR under way.

### DIFF
--- a/mavros/src/plugins/setpoint_raw.cpp
+++ b/mavros/src/plugins/setpoint_raw.cpp
@@ -56,9 +56,9 @@ public:
 	Subscriptions get_subscriptions()
 	{
 		return {
-			       make_handler(&SetpointRawPlugin::handle_position_target_local_ned),
-			       make_handler(&SetpointRawPlugin::handle_position_target_global_int),
-			       make_handler(&SetpointRawPlugin::handle_attitude_target),
+				make_handler(&SetpointRawPlugin::handle_position_target_local_ned),
+				make_handler(&SetpointRawPlugin::handle_position_target_global_int),
+				make_handler(&SetpointRawPlugin::handle_attitude_target),
 		};
 	}
 
@@ -216,26 +216,47 @@ private:
 
 	void attitude_cb(const mavros_msgs::AttitudeTarget::ConstPtr &req)
 	{
+		double thrust_scaling;
 		Eigen::Quaterniond desired_orientation;
 		Eigen::Vector3d baselink_angular_rate;
+		Eigen::Vector3d body_rate;
+		double thrust;
 
-		tf::quaternionMsgToEigen(req->orientation, desired_orientation);
+		// Set Thrust scaling in px4_config.yaml, setpoint_raw block.
+		// ignore thrust is false by default, unless no thrust scalling is set or thrust is zero
+		auto ignore_thrust = req->thrust != 0.0 && !sp_nh.getParam("thrust_scaling", thrust_scaling);
+
+		if (ignore_thrust) {
+			// I believe it's safer without sending zero thrust, but actually ignoring the actuation.
+			ROS_FATAL_THROTTLE_NAMED(5, "setpoint_raw", "Recieved thrust, but ignore_thrust is true: "
+				"the most likely cause of this is a failure to specify the thrust_scaling parameters "
+				"on px4/apm_config.yaml. Actuation will be ignored.");
+			return;
+		} else {
+			if (thrust_scaling == 0.0) {
+				ROS_WARN_THROTTLE_NAMED(5, "setpoint_raw", "thrust_scaling parameter is set to zero.");
+			}
+			thrust = std::min(1.0, std::max(0.0, req->thrust * thrust_scaling));
+		}
+
+		// Take care of attitude setpoint
+		desired_orientation = ftf::to_eigen(req->orientation);
 
 		// Transform desired orientation to represent aircraft->NED,
 		// MAVROS operates on orientation of base_link->ENU
 		auto ned_desired_orientation = ftf::transform_orientation_enu_ned(
-					ftf::transform_orientation_baselink_aircraft(desired_orientation));
+			ftf::transform_orientation_baselink_aircraft(desired_orientation));
 
-		auto body_rate = ftf::transform_frame_baselink_aircraft(baselink_angular_rate);
-
-		tf::vectorMsgToEigen(req->body_rate, body_rate);
+		body_rate = ftf::transform_frame_baselink_aircraft(
+			ftf::to_eigen(req->body_rate));
 
 		set_attitude_target(
 					req->header.stamp.toNSec() / 1000000,
 					req->type_mask,
 					ned_desired_orientation,
 					body_rate,
-					req->thrust);
+					thrust);
+
 	}
 };
 }	// namespace std_plugins
--- a/mavros/launch/px4_config.yaml
+++ b/mavros/launch/px4_config.yaml
@@ -98,6 +98,9 @@ setpoint_attitude:
     child_frame_id: "target_attitude"
     rate_limit: 50.0
 
+setpoint_raw:
+  thrust_scaling: 0.0       # used in rpyt callback
+
 # setpoint_position
 setpoint_position:
   tf:


vooon:
It's hard to see actual changes behind all that unneded whitespace changes.

Pedro-Roque:
> It's hard to see actual changes behind all that unneded whitespace changes.

Those white spaces were removed with the uncrustify. Should I put them back?

vooon:
Uncrustify someptimes does incorrect indentation, for example look at stair
of folded functions...
Better only include functional changes, just follow: tab indent, tab=8
spaces, mostly linux kernel style.

сб, 17 нояб. 2018 г. в 13:18, Pedro Roque <notifications@github.com>:

Pedro-Roque:
@vooon Check it now. The first 3 make_handler are already tabbed with 8 spaces, but git still says there is a change. Let me know if it needs fix or if we leave it like this... Cheers!

mavros/src/plugins/setpoint_raw.cpp
```diff
@@ -218,24 +222,44 @@ class SetpointRawPlugin : public plugin::PluginBase,
...
+		if (!sp_nh.getParam("thrust_scaling", thrust_scaling) && req->thrust != 0.0) {
+			ignore_thrust_message_ = true;
+		}
+
+		if (ignore_thrust_message_) {
+			ROS_FATAL("Recieved thrust, but ignore_thrust_message_ is true: "
+				"the most likely cause of this is a failure to specify the thrust_scaling parameters "
+				"on px4/apm_config.yaml. Thrust will be set to 0.");
+			thrust = 0.0;
+			return;
```
vooon: I'm not sure that you want just drop message (because you return before send).

mavros/src/plugins/setpoint_raw.cpp
```diff
@@ -218,24 +222,44 @@ class SetpointRawPlugin : public plugin::PluginBase,
...
 		// MAVROS operates on orientation of base_link->ENU
 		auto ned_desired_orientation = ftf::transform_orientation_enu_ned(
-					ftf::transform_orientation_baselink_aircraft(desired_orientation));
+			ftf::transform_orientation_baselink_aircraft(desired_orientation));
 
-		auto body_rate = ftf::transform_frame_baselink_aircraft(baselink_angular_rate);
-
-		tf::vectorMsgToEigen(req->body_rate, body_rate);
+		body_rate = ftf::transform_frame_baselink_aircraft(
+			ftf::to_eigen(req->body_rate));
```
vooon: Why not leave here `auto body_rate = ...`.
Change to `to_eigen` - ok.

mavros/src/plugins/setpoint_raw.cpp
Pedro-Roque: Yes, here I just want to print the warning and return without sending anything - for safety. If the user didnt specify the thrust scaling, then we don't want to do anything.

mavros/src/plugins/setpoint_raw.cpp
Pedro-Roque: It seemed to me more explicit to define it at the begginning, but I can change to auto if you think it's optimal. Let me know and I'll do it the soonest.

mavros/src/plugins/setpoint_raw.cpp
vooon: > Thrust will be set to 0.

Is confusing, since you just drop whole message.

mavros/src/plugins/setpoint_raw.cpp
```diff
@@ -218,24 +222,44 @@ class SetpointRawPlugin : public plugin::PluginBase,
 	{
 		Eigen::Quaterniond desired_orientation;
 		Eigen::Vector3d baselink_angular_rate;
-
+		Eigen::Vector3d body_rate;
+		double thrust;
+		// Set Thrust scaling in px4_config.yaml, setpoint_raw block.
+		if (!sp_nh.getParam("thrust_scaling", thrust_scaling) && req->thrust != 0.0) {
+			ignore_thrust_message_ = true;
+		}
```
vooon: One of the problems of getParam - it is not fast, but at least it have cache (in roscpp).
That style will look a bit better:
```cpp
auto ignore_thrust = req->thrust != 0.0 && !sp_nh.getParam(...)
```

mavros/src/plugins/setpoint_raw.cpp
```diff
@@ -71,6 +72,9 @@ class SetpointRawPlugin : public plugin::PluginBase,
 	ros::Subscriber local_sub, global_sub, attitude_sub;
 	ros::Publisher target_local_pub, target_global_pub, target_attitude_pub;
 
+	double thrust_scaling;
+	bool ignore_thrust_message_;
+
```
vooon: That variables is not used outside of handle func, no reason to make them private members.

mavros/src/plugins/setpoint_raw.cpp
```diff
@@ -218,24 +222,44 @@ class SetpointRawPlugin : public plugin::PluginBase,
...
+			return;
+		} else {
+			if (thrust_scaling == 0.0) {
+				ROS_WARN("thrust_scaling parameter is set to zero.");
+			}
+			thrust = std::min(1.0, std::max(0.0, req->thrust * thrust_scaling));
+		}
+
+		// Take care of attitude setpoint
 		tf::quaternionMsgToEigen(req->orientation, desired_orientation);
```
vooon: May be changed to to_eigen too.

mavros/src/plugins/setpoint_raw.cpp
```diff
@@ -218,24 +222,44 @@ class SetpointRawPlugin : public plugin::PluginBase,
 	{
 		Eigen::Quaterniond desired_orientation;
 		Eigen::Vector3d baselink_angular_rate;
-
+		Eigen::Vector3d body_rate;
+		double thrust;
+		// Set Thrust scaling in px4_config.yaml, setpoint_raw block.
```
vooon: Neet blank line between definition block and following code.

mavros/src/plugins/setpoint_raw.cpp
vooon: Usually we get parameter in `initialize()`, but your way allows to tune value.

mavros/src/plugins/setpoint_raw.cpp
Pedro-Roque: I don't really know how this came here

mavros/src/plugins/setpoint_raw.cpp
Pedro-Roque: Done.

mavros/src/plugins/setpoint_raw.cpp
Pedro-Roque: Done.

mavros/src/plugins/setpoint_raw.cpp
Pedro-Roque: Just checked this now. I will do the way you suggest for the definition, and keep the get_param inside the handle to allow re-tuning. But I agree with you, it's not optimal as it is slow, I might try to optimize it later on...

mavros/src/plugins/setpoint_raw.cpp
Pedro-Roque: Yep, I changed that comment and removed thrust zero-ing (it's not needed since we immediately return).

mavros/src/plugins/setpoint_raw.cpp
Pedro-Roque: Done.

mavros/src/plugins/setpoint_raw.cpp
```diff
@@ -216,26 +216,47 @@ class SetpointRawPlugin : public plugin::PluginBase,
...
+
+		if (ignore_thrust) {
+			// I believe it's safer without sending zero thrust, but actually ignoring the actuation.
+			ROS_FATAL("Recieved thrust, but ignore_thrust is true: "
+				"the most likely cause of this is a failure to specify the thrust_scaling parameters "
+				"on px4/apm_config.yaml. Actuation will be ignored.");
+			return;
+		} else {
+			if (thrust_scaling == 0.0) {
+				ROS_WARN("thrust_scaling parameter is set to zero.");
```
vooon: One more thing, i prefer ROS_xxx_NAMED() loggers in plugins.

Pedro-Roque: Done. Gave it the plugin name and set the frequency for once every 5 seconds.

PR #1122, opened 2018-11-20
apm_config: enable timesync and system for ardupilot

rmackay9:
ArduPilot has added support for handling of the TIMESYNC and SYSTEM_TIME messages and it seems to be working.

- documentation on the setup is [here on the AP developer wiki](ardupilot.org/dev/docs/ros-timesync.html) (should be visible about 10min after this PR was submitted)
- below is an image of the clocks syncronised between a TX2 companion computer and a Cube flight controller.

ros-time-sync-testing


rmackay9:
@vooon,

Thanks for merging.

You're right about the system-time rate being way too high from AP.  I'll look into that and also the mismatch in the TIMESYNC messages.

PR #1136, opened 2018-12-12
Request timed up parameters as soon as possible

angri:
Current version of PX4 has almost 700 parameters. Pulling all of those at once even through serial connection @ 921600 time after time leads to timeout for some parameters (usually 3-5, sometimes about 30, depending on what else is going on on the link). Current implementation of params plugin rerequests missing parameters, but does it only on timeout event. So after the first timeout has passed params plugin re requests the first missing parameter and resets the timeout. Then the second missing parameters gets processed only when timeout happens again. It takes over 30 seconds to pull 30 missing parameters. During that time the service ~param/pull doesn't return.

Proposed solution makes params plugin rerequest the next missing parameter as soon as the previous one was delivered.


angri:
Ok, so I found a bug in my code. The first timeout to happen was changing state to RXPARAM_TIMEDOUT, but that state was not checked for in the timeout callback itself. So now it's supposed to work this way: request parameter list and drop out of parameters_missing_idx all the parameters that were delivered. When first timeout happens plugin switches to RXPARAM_TIMEDOUT state and starts to request parameters one by one. Each of those still have 3 attempts (see line 492).

vooon:
Now looks better, LGTM. Thank you!

PR #1142, opened 2018-12-20
Handle LOCAL_POSITION_NED_COV, refactor local position plugin

okalachev:
Right now it's impossible to get flight controller's estimated covariances as well as estimated accelerations. I added handling `LOCAL_POSITION_NED_COV` message, which may contain the mentioned fields. Also I created a PX4 PR (https://github.com/PX4/Firmware/pull/11081) fixing some `LOCAL_POSITION_NED_COV` issues. But this message stream still can only be enabled only in the code.

I propose the next topics for `local_position` plugin:

* `pose` (`PoseStamped`)
* `pose_cov` (`PoseWithCovarianceStamped`)
* `veloicity` (`TwistStamped`)
* `velocity_cov` (`TwistWithCovarianceStamped`)
* `accel` (`AccelWithCovarianceStamped`)
* `odom` (`Odometry`)

I created fallback logic:

* If there is only `LOCAL_POSITION_NED` stream is available, then only `pose`, `velocity` and `odom` (without covariances) is published.
* If there is only `LOCAL_POSITION_NED_COV` stream is available then `pose`, `pose_cov`, `velocity`, `velocity_cov`, `accel` and `odom` is published.
* If the both messages stream are available then, `LOCAL_POSITION_NED` corresponds to `pose` and `velocity` topics, and `LOCAL_POSITION_NED_COV` corresponds to `pose_cov`, `velocity_cov`, `accel` and `odom` topics.

Also I fixed a little incorrectness in `velocity` topic: before it was published in `map` frame, but the linear velocities was set w. r. t. body frame, and it was shown incorrect in Rviz.

The basic ROS recommendation [[1]](http://wiki.ros.org/tf/Reviews/2010-03-12_API_Review) is to publish velocity message in body frame, so now all twist messages are in `base_link` frame.

Also I made some refactoring to simplify the code and decided to remove the weird `NED -> aircraft` transform as it's unconfigurable and also seems useless.

### DIFF
--- a/mavros/src/plugins/local_position.cpp
+++ b/mavros/src/plugins/local_position.cpp
@@ -19,8 +19,11 @@
 #include <mavros/mavros_plugin.h>
 #include <eigen_conversions/eigen_msg.h>
 
+#include <geometry_msgs/AccelWithCovarianceStamped.h>
 #include <geometry_msgs/PoseStamped.h>
+#include <geometry_msgs/PoseWithCovarianceStamped.h>
 #include <geometry_msgs/TwistStamped.h>
+#include <geometry_msgs/TwistWithCovarianceStamped.h>
 #include <geometry_msgs/TransformStamped.h>
 
 #include <nav_msgs/Odometry.h>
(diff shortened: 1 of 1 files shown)


okalachev:
PX4 denied to use `LOCAL_POSITION_NED_COV` message in favour of developed `ODOMETRY` message.

This PR still can make some sense, because:

1. `LOCAL_POSITION_NED_COV` may be use by other flight controllers.
2. It fixes velocity frame issue.
3. It simplifies and clarifies the code.

Then, the plugin should be extended for usage of `ODOMETRY` message.

TSC21:
I am also ok with this PR, even though PX4 dropped support.
The `odom` plugin in mavros_extras already supports ODOMETRY (sends, does not receive). I am working today on extending it to receive as well.
Merry Christmas everyone!

vooon:
I think i can merge it as is. And improvements later.

TSC21:
> Also I fixed a little incorrectness in `velocity` topic: before it was published in `map` frame, but the linear velocities was set w. r. t. body frame, and it was shown incorrect in Rviz.

@okalachev this is true for the `nav_msgs/Odometry` tropic, but I don't see in the spec why this has to be on the body frame for any `geometry_msgs/Twist` topic. This was reported as an issue in https://github.com/mavlink/mavros/issues/1159 and I will fix it in a PR.

okalachev:
I used [jsk_rviz_plugins](https://jsk-visualization.readthedocs.io/en/latest/jsk_rviz_plugins/plugins/twist_stamped.html) to visualize `TwistStamped` topic. And with this change both linear and angular velocities are visualised correctly.

Without it they weren't.

TSC21:
> I used [jsk_rviz_plugins](https://jsk-visualization.readthedocs.io/en/latest/jsk_rviz_plugins/plugins/twist_stamped.html) to visualize `TwistStamped` topic. And with this change both linear and angular velocities are visualised correctly.
>
>Without it they weren't.

There are projects, as you can see, that are dependent on the velocity on the local frame. We can't just drop that support. I added a new topic in https://github.com/mavlink/mavros/pull/1160

PR #1148, opened 2019-01-01
param plugin : add msg and publisher to catch latest param value

Kiwa21:
As requested by the goodrobots projects by @fnoop : https://github.com/goodrobots

I'm open to all suggestions.
Thanks,

### DIFF
--- a/mavros/src/plugins/param.cpp
+++ b/mavros/src/plugins/param.cpp
@@ -22,6 +22,7 @@
 #include <mavros_msgs/ParamGet.h>
 #include <mavros_msgs/ParamPull.h>
 #include <mavros_msgs/ParamPush.h>
+#include <mavros_msgs/Param.h>
 
 namespace mavros {
 namespace std_plugins {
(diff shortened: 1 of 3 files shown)


TSC21:
@Kiwa21 can you please explain the rationale behind adding this?

Kiwa21:
Goodrobots is developing a web GCS / web API.
They had the need to catch if a single parameter was changed by any other third means to be bale to update their own parameter table.
Maybe @fnoop can detail a bit more

mavros_msgs/msg/Param.msg
```diff
@@ -0,0 +1,7 @@
+# Parameter msg.
+
+string param_id
```
vooon: Here maybe better to add Header.
Just to carry receive stamp.

mavros/src/plugins/param.cpp
```diff
@@ -420,6 +425,13 @@ class ParamPlugin : public plugin::PluginBase {
 
 		auto param_id = mavlink::to_string(pmsg.param_id);
 
+		mavros_msgs::Param param_msg;
+		param_msg.param_id =  param_id;
+        param_msg.param_type = pmsg.param_type;
+        param_msg.param_count = pmsg.param_count;
+		param_msg.param_index = pmsg.param_index;
+	 	param_value_pub.publish(param_msg)
```
vooon: Where is filling of value? Indentation error.

Issue #370, opened 2015-08-10
Add support for requesting message interval


vooon:
Interface introduced in https://github.com/mavlink/mavlink/commit/25eccb1bfc5db3758d8336791641e06aa098cb3c

vooon:
Interface implemented in PX4. APM still uses per-stream messages.

Need to move message rate funcs to `sys_msgctl` plugin.

TSC21:
@vooon what's the state of this?

vooon:
Still not implemented on our side. But not so bad, because you still may configure PX4 via SD card.

rmackay9:
ArduPilot is just about to add support for the get/set-message-interval message.  It would certainly be nice to be able to use the mavros setrate command to be able to set a specific message's update rate.  I might try and do that myself although I certainly wouldn't complain if someone beat me to it.

vooon:
@rmackay9 i think best way to do it - just add `~set_message_interval` service to `sys_status` and don't change older set-rate (but mark it as deprecated).
Then add new subcommand to mavsys. Perhaps it would be great if new command can use message names or msgid.

But problem is that this info is hidden inside libmavconn.

vooon:
Oh, and another idea - add interface like param to set intervals on connection.

PR #1156, opened 2019-01-12
Add support for set message interval

rmackay9:
This PR attempts to close this issue: https://github.com/mavlink/mavros/issues/370

- new set_message_interval service added which accepts MessageInterval requests which provide a desired message id and rate (in hz) and then sends a COMMAND_LONG (using the CommandLong service) containing a MAV_CMD_SET_MESSAGE_INTERVAL.

   - The only slight complexity is that the request specifies a rate while the mavlink message provides an interval in us (I figured a rate was easier to use but it's possible this logic should have been put in mavsys instead of within the service).
   - If the rate is set to 0 or -1 set the outgoing message-interval is also set to 0 or -1 which sets the rate to the default or turns off the message.  Perhaps it would be more logical to have the caller specify "rate=0" to turning off the message and "rate=-1" for default rate.

- mavsys python script enhanced to allow using the above service from the command line

I've tested this on an AION robotics rover with an Nvidia tx2 running ROS and a Cube running ArduPilot Rover-3.5.0-dev and it worked correctly.  Some of the tests performed:

- set the RAW_IMU rate to 5hz and confirmed the update rate to the ground station (MP) changed correctly
- set the RAW_IMU rate to 0 to confirm the rate changed to the default (which was 3hz)
- set the RAW_IMU rate to -1 to confirm the message stopped being sent by the flight controller

Below are some screen shots of it in action.

mavros-rawimu-update-cmdline
mavros-mp-rawimu-update-rate

### DIFF
--- a/mavros/scripts/mavsys
+++ b/mavros/scripts/mavsys
@@ -16,7 +16,7 @@ import rospy
 import mavros
 from mavros.utils import *
 from mavros_msgs.msg import State
-from mavros_msgs.srv import SetMode, StreamRate, StreamRateRequest
+from mavros_msgs.srv import SetMode, StreamRate, StreamRateRequest, MessageInterval
 
 def do_mode(args):
     base_mode = 0
@@ -78,6 +78,24 @@ def do_rate(args):
 
 
 
+def do_message_interval(args):
+
+    if args.id is None:
+        fault("id not specified")
+        return
+
+    if args.rate is None:
+        fault("rate not specified")
+        return
+
+    try:
+        set_message_interval = rospy.ServiceProxy(mavros.get_topic('set_message_interval'), MessageInterval)
+        set_message_interval(message_id=args.id, message_rate=args.rate)
+    except rospy.ServiceException as ex:
+        fault(ex)
+
+
+
 def main():
     parser = argparse.ArgumentParser(description="Change mode and rate on MAVLink device.")
     parser.add_argument('-n', '--mavros-ns', help="ROS node namespace", default=mavros.DEFAULT_NAMESPACE)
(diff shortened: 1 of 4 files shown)


mavros/scripts/mavsys
```diff
@@ -78,6 +78,24 @@ def do_rate(args):
 
 
 
+def do_message_interval(args):
+
+    if args.id is None:
+        fault("id not specified")
+        return
```
vooon: Perhaps better to add ability to specify mavlink message name in mavsys.
That can be done by using pymavlink.

Idea:
```python
try:
    from pymavlink.dialects.v20 import common as ml_common
except ImportError:
    ml_common = None


def find_id(name):
    rev = {v.name: k for k, v in ml_common.mavlink_map.items()}
    return rev[name.upper()]

File: mavros_msgs/srv/MessageInterval.srv [hunk 2]
```diff
@@ -0,0 +1,7 @@
+# sets message interval
+# See MAV_CMD_SET_MESSAGE_INTERVAL
+
+uint16 message_id
```
vooon: That is wrong, because v2 ids is 24 bit. You should use uint32 here.

vooon:
I'll fix that small issue myself, thank you!

shubham-shahh:
Hi @rmackay9, @vooon thanks for this PR, when I use this method, the rate goes up for a few secs, then it comes back to 4Hz

for example, I used 

`rosrun mavros mavsys message_interval --id 32 --rate 10`

to increase the rate of local_position/odom (LOCAL_POSITION_NED)

that rate reaches to 10Hz and the in sometime, it falls back to 4Hz

shubham-shahh:
> Hi @rmackay9, @vooon thanks for this PR, when I use this method, the rate goes up for a few secs, then it comes back to 4Hz
> 
> for example, I used
> 
> `rosrun mavros mavsys message_interval --id 32 --rate 10`
> 
> to increase the rate of local_position/odom (LOCAL_POSITION_NED)
> 
> I am using Ardupilot 4.1.5
> 
> that rate reaches to 10Hz and then in some time, it falls back to 4Hz

A few resources for people facing this issue
[issue](https://discuss.ardupilot.org/t/set-message-interval-ignored-by-sitl-after-a-few-seconds/69411)
[docs](https://ardupilot.org/dev/docs/mavlink-requesting-data.html)

Issue #1159, opened 2019-01-23
0.28.0 regression: topic /mavros/local_position/velocity now in drone rotation frame


jkflying:
### Issue details

With PX4/Avoidance running in SITL, the drone velocity is reported in the local co-ordinate frame (ie. rotated to the drone). We see this causing multiple issues with trying to follow a position setpoint from the avoidance, as we have a bias East no matter what direction we are flying.

Previously in 0.27.0 the velocity at /mavros/local_position/velocity was in the local reference frame, not the drone reference frame.

### MAVROS version and platform

Mavros: 0.28.0
ROS: Kinetic
Ubuntu: 16.04

### Autopilot type and version

SITL / PX4

### Node logs

```
SUMMARY
========

CLEAR PARAMETERS
 * /mavros/

PARAMETERS
 * /factor_close_to_goal_start_speed_limitation: 3.0
 * /factor_close_to_goal_stop_speed_limitation: 4.0
 * /goal_acceptance_radius_in: 0.5
 * /goal_acceptance_radius_out: 1.5
 * /local_planner_node/goal_x_param: 17
 * /local_planner_node/goal_y_param: 15
 * /local_planner_node/goal_z_param: 3
 * /local_planner_node/pointcloud_topics: ['/camera_front/d...
 * /local_planner_node/world_name: /home/julian/catk...
 * /mavros/cmd/use_comp_id_system_control: False
 * /mavros/conn/heartbeat_rate: 1.0
 * /mavros/conn/system_time_rate: 1.0
 * /mavros/conn/timeout: 10.0
 * /mavros/conn/timesync_rate: 10.0
 * /mavros/distance_sensor/hrlv_ez4_pub/field_of_view: 0.0
 * /mavros/distance_sensor/hrlv_ez4_pub/frame_id: hrlv_ez4_sonar
 * /mavros/distance_sensor/hrlv_ez4_pub/id: 0
 * /mavros/distance_sensor/hrlv_ez4_pub/orientation: ROLL_180
 * /mavros/distance_sensor/hrlv_ez4_pub/send_tf: True
 * /mavros/distance_sensor/hrlv_ez4_pub/sensor_position/x: 0.0
 * /mavros/distance_sensor/hrlv_ez4_pub/sensor_position/y: 0.0
 * /mavros/distance_sensor/hrlv_ez4_pub/sensor_position/z: -0.1
 * /mavros/distance_sensor/laser_1_sub/id: 3
[...]

PR #1160, opened 2019-01-23
local_position: add an aditional topic for velocity on the local frame

TSC21:
Fixes https://github.com/mavlink/mavros/issues/1159. We now have a `local_position/velocity_body`, `local_position/velocity_local` and `local_position/velocity_body_cov` (since only odometry messages contain covariance information, and the velocity on these is in the body frame).

@jkflying can you give a try to this? Thanks.

### DIFF
--- a/mavros/src/plugins/local_position.cpp
+++ b/mavros/src/plugins/local_position.cpp
@@ -57,16 +57,17 @@ public:
 
 		local_position = lp_nh.advertise<geometry_msgs::PoseStamped>("pose", 10);
 		local_position_cov = lp_nh.advertise<geometry_msgs::PoseWithCovarianceStamped>("pose_cov", 10);
-		local_velocity = lp_nh.advertise<geometry_msgs::TwistStamped>("velocity", 10);
-		local_velocity_cov = lp_nh.advertise<geometry_msgs::TwistWithCovarianceStamped>("velocity_cov", 10);
+		local_velocity_local = lp_nh.advertise<geometry_msgs::TwistStamped>("velocity_local", 10);
+		local_velocity_body = lp_nh.advertise<geometry_msgs::TwistStamped>("velocity_body", 10);
+		local_velocity_cov = lp_nh.advertise<geometry_msgs::TwistWithCovarianceStamped>("velocity_body_cov", 10);
 		local_accel = lp_nh.advertise<geometry_msgs::AccelWithCovarianceStamped>("accel", 10);
 		local_odom = lp_nh.advertise<nav_msgs::Odometry>("odom",10);
 	}
 
 	Subscriptions get_subscriptions() {
 		return {
-		       make_handler(&LocalPositionPlugin::handle_local_position_ned),
-		       make_handler(&LocalPositionPlugin::handle_local_position_ned_cov)
+			       make_handler(&LocalPositionPlugin::handle_local_position_ned),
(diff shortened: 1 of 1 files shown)


jkflying:
On this version I get a continuous stream of the following printed warnings, I don't see them on master:

```
Error:   TF_DENORMALIZED_QUATERNION: Ignoring transform for child_frame_id "fcu" from authority "unknown_publisher" because of an invalid quaternion in the transform (0.000000 0.000000 0.000000 0.000000)
         at line 257 in /tmp/binarydeb/ros-kinetic-tf2-0.5.20/src/buffer_core.cpp
[ERROR] [1548258028.028077195, 11.976000000]: Ignoring transform for child_frame_id "fcu" from authority "unknown_publisher" because of an invalid quaternion in the transform (0.000000 0.000000 0.000000 0.000000)
Error:   TF_DENORMALIZED_QUATERNION: Ignoring transform for child_frame_id "fcu" from authority "unknown_publisher" because of an invalid quaternion in the transform (0.000000 0.000000 0.000000 0.000000)
         at line 257 in /tmp/binarydeb/ros-kinetic-tf2-0.5.20/src/buffer_core.cpp
[ERROR] [1548258028.061005889, 12.012000000]: Ignoring transform for child_frame_id "fcu" from authority "unknown_publisher" because of an invalid quaternion in the transform (0.000000 0.000000 0.000000 0.000000)
[ERROR] [1548258028.061048745, 12.012000000]: Ignoring transform for child_frame_id "fcu" from authority "unknown_publisher" because of an invalid quaternion in the transform (0.000000 0.000000 0.000000 0.000000)

```

jkflying:
@vooon Do you have any rough timeline for the release of 0.29.0? I need to update PX4/Avoidance (https://github.com/PX4/avoidance/pull/197), and I'd prefer to point at a real release rather than a random hash.

okalachev:
@TSC21, why rename `velocity` to `velocity_local`? This has broken everything for us.

Aren't we concerned about backwards compatibility in this project?

TSC21:
> @TSC21, why rename `velocity` to `velocity_local`? This has broken everything for us.
> 
> Aren't we concerned about backwards compatibility in this project?

@okalachev actually your PR initially broke the compatibility for the Avoidance team projects. The only way was to provide both velocity in local and body frame. Renaming it to velocity_local is the normal procedure so to make sure we keep the body and the local velocities properly identified.

okalachev:
I believe back then we reverted back what was breaking the compatibility.

Why not to use `velocity` for local frame (as it used to be) and `velocity_body` for body frame?

TSC21:
We try to keep the backward compatibility the max possible but there are maintenance decisions that need to be taken into account. Your change on the local_position plugin for example took repercussions on other projects and that had to be fixed as well. We always try yo keep the balance between the required changes and the repercussions they take.

TSC21:
> I believe back then we reverted back what was breaking the compatibility.
> 
> Why not to use `velocity` for local frame (as it used to be) and `velocity_body` for body frame?

No that was not the case. `velocity` was publishing body velocities, according to @jkflying and their team. So that had to be reverted.

jkflying:
@okalachev this was changed in your own PR, here: https://github.com/mavlink/mavros/pull/1142

We had to add a new `velocity_local` to get the old values back.

okalachev:
Isn't the new `velocity_local` the same as the old `velocity`?

PR #1161, opened 2019-01-25
px4_config: set the thrust_scaling to one by default

TSC21:
This should be set to unitary by default, as being 0 means there's no thrust.

@lamping7 FYI.

@vooon what's the possibility of having a 0.28.1 minor release to bring this? The reason is that this is currently breaking PX4 CI.


lamping7:
@vooon can you tag a release with this?

Issue #1174, opened 2019-02-07
mavros compilation error on arm64


acwatkins:
This is only bug and feature tracker, please use it
to report bugs or request features.

----

### Issue details

On amd64 platforms mavros builds fine, but on an arm64, specifically the NVidia TX2, there is a timing error between mavros_msgs and mavros, causing a compile error with a missing include. This is remedied by building twice.

**compile error:**
```
In file included from /ros_catkin_ws/src/mavros/mavros/include/mavros/utils.h:22:0,
                 from /ros_catkin_ws/src/mavros/mavros/include/mavros/mavros_uas.h:29,
                 from /ros_catkin_ws/src/mavros/mavros/src/lib/uas_data.cpp:17:
/ros_catkin_ws/src/mavros/mavros_msgs/include/mavros_msgs/mavlink_convert.h:17:33: fatal error: mavros_msgs/Mavlink.h: No such file or directory
compilation terminated.
mavros/mavros/CMakeFiles/mavros.dir/build.make:62: recipe for target 'mavros/mavros/CMakeFiles/mavros.dir/src/lib/uas_data.cpp.o' failed
```

### MAVROS version and platform

Mavros: 0.29 (master)
ROS: Kinetic
Ubuntu: 16.04

KiloNovemberDelta:
Hello, 
I have succeed to install Mavros on the Jetson TX2 through the apt-get command : 
``install ros-kinetic-mavros ros-kinetic-mavros-extras``

Or you really want to build it from source and catkin_ws ?

PR #1175, opened 2019-02-07
Issue #1174 Added dependency for mavros_msgs and mavros

acwatkins:


vooon:
That should be already done by [mavros/package.xml L53](https://github.com/mavlink/mavros/blob/master/mavros/package.xml#L53).
Cross-package dependency in cmake is bad idea.

acwatkins:
I normally do this, for projects that generate messages:
```
add_dependencies(your_program ${catkin_EXPORTED_TARGETS})
```
I believe the above command makes it wait for everything, not just what is used by that section. It seems the above is more accepted though.

http://docs.ros.org/jade/api/catkin/html/howto/format2/cpp_msg_dependencies.html

Some discussion on the topic:
https://answers.ros.org/question/52744/how-to-specify-dependencies-with-foo_msgs-catkin-packages/

vooon:
But mavros uses format 2, so http://docs.ros.org/kinetic/api/catkin/html/howto/format2/cpp_msg_dependencies.html

Can you do that for mavros lib and extras?

vooon:
I will do it myself.

PR #1176, opened 2019-02-10
mavros_msgs: Fix line endings for OpticalFlowRad message

sfalexrog:
As of the time of this pull request, [OpticalFlowRad.msg](https://github.com/mavlink/mavros/blob/5ae211f2843c3442169cc9dbd537dabab82d95bf/mavros_msgs/msg/OpticalFlowRad.msg) uses CRLF sequence for line breaks. This upsets some message parsers (like the one used by [rqt_multiplot](https://github.com/ANYbotics/rqt_multiplot_plugin)), preventing them from displaying the message structure properly.

This P/R addresses that, converting line ends to UNIX ones.


mhkabir:
Can you do a pass and check the other messages and services as well?

sfalexrog:
Hi! Yes, all other .msg and .srv files use UNIX line endings.

Of all other files, it seems that [mavros_extras/src/copter_visualization.cpp](https://github.com/mavlink/mavros/blob/master/mavros_extras/src/copter_visualization.cpp) and [mavros_extras/src/plugins/px4flow.cpp](https://github.com/mavlink/mavros/blob/master/mavros_extras/src/plugins/px4flow.cpp) use CRLF line endings. Should I convert them to UNIX as well?

mhkabir:
Thanks!

vooon:
That is not necessary to update C++ files, but it is not a problem.
LGTM, Thank you!

PR #1178, opened 2019-02-12
Update apm_config.yaml - Adding thrust_scaling to 1.0 in setpoint_raw msg

KiloNovemberDelta:
Setting thrust_scaling in the setpoint_raw message (in my case, to use /mavros/setpoint_raw/attitude)
Without it, when using Gazebo, get the following problem
"Recieved thrust, but ignore_thrust is true: the most likely cause of this is a failure to specify the thrust_scaling parameters on px4/apm_config.yaml. Actuation will be ignored." from the function void attitude_cb in setpoint_raw.cpp (http://docs.ros.org/kinetic/api/mavros/html/setpoint__raw_8cpp_source.html)

### DIFF
--- a/mavros/launch/apm_config.yaml
+++ b/mavros/launch/apm_config.yaml
@@ -98,6 +98,10 @@ setpoint_attitude:
     frame_id: "map"
     child_frame_id: "target_attitude"
     rate_limit: 50.0
+    
+ # setpoint_raw
+ setpoint_raw:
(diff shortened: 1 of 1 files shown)


vooon:
@TSC21 what is your opinion?

TSC21:
I had to do that for PX4 for the same reason. So this is ok, as long as APM is working correctly with this.

PR #1181, opened 2019-02-18
distance_sensor: Allow configurable variance window size

sfalexrog:
Currently ```distance_sensor``` plugin from ```mavros_extras``` uses a sample window with a hardcoded size to calculate distance sensor variance that is sent to the autopilot. This results in incorrectly large variances for low-frequency sensors (especially "smart" sensors that calculate variance themselves), especially during takeoff and landing. To compensate for that I suggest adding a parameter to ```distance_sensor``` with the desired window size.

This P/R introduces a new parameter, ```variance_window_size```, to the ```distance_sensor``` plugin. It controls how many samples will be used in the (co)variance calculation, and defaults to 50 (the old, hardcoded size). If this parameter is set to 0 or a negative value, the variance calculations are disabled, and the distance sensor (co)variance is set to 0.

### DIFF
--- a/mavros_extras/src/plugins/distance_sensor.cpp
+++ b/mavros_extras/src/plugins/distance_sensor.cpp
@@ -280,7 +280,7 @@ void DistanceSensorItem::range_cb(const sensor_msgs::Range::ConstPtr &msg)
 	uint8_t covariance_ = 0;
 
 	if (covariance > 0) covariance_ = covariance;
-	else if (variance_window_size <= 0) covariance_ = 0;
+	else if (variance_window_size <= 0) covariance_ = 255; // Assume unknown variance
 	else covariance_ = uint8_t(calculate_variance(msg->range) * 1E2);	// in cm
 
 	/** @todo Propose changing covarince from uint8_t to float */
--- a/mavros_extras/src/plugins/distance_sensor.cpp
+++ b/mavros_extras/src/plugins/distance_sensor.cpp
@@ -280,7 +280,7 @@ void DistanceSensorItem::range_cb(const sensor_msgs::Range::ConstPtr &msg)
 	uint8_t covariance_ = 0;
 
 	if (covariance > 0) covariance_ = covariance;
-	else if (variance_window_size <= 0) covariance_ = 255; // Assume unknown variance
+	else if (variance_window_size <= 0) covariance_ = UINT8_MAX; // Assume unknown variance
 	else covariance_ = uint8_t(calculate_variance(msg->range) * 1E2);	// in cm
 
 	/** @todo Propose changing covarince from uint8_t to float */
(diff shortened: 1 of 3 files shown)


TSC21:
An uknown variance should have a value of 256

TSC21:
https://mavlink.io/en/messages/common.html#DISTANCE_SENSOR

sfalexrog:
@TSC21 But AFAIK uint8_t is [0..255], so how can it be set to 256?

TSC21:
Sorry. I meant 255. There seems to be a typo on the Mavlink docs. @hamishwillee

TSC21:
@sfalexrog please set it to 255.

mavros_extras/src/plugins/distance_sensor.cpp
```diff
@@ -280,7 +280,7 @@ void DistanceSensorItem::range_cb(const sensor_msgs::Range::ConstPtr &msg)
 	uint8_t covariance_ = 0;
 
 	if (covariance > 0) covariance_ = covariance;
-	else if (variance_window_size <= 0) covariance_ = 0;
+	else if (variance_window_size <= 0) covariance_ = 255; // Assume unknown variance
```
TSC21: `UINT8_MAX`

sfalexrog: Changed that, thanks!

sfalexrog:
@TSC21 Could you elaborate on comparison to -1? Do you mean writing something like

```C++
else if (variance_window_size < 1) covariance_ = UINT8_MAX; // Assume unknown variance
```
instead of the current 

```C++
else if (variance_window_size <= 0) covariance_ = UINT8_MAX; // Assume unknown variance
```

TSC21:
What I ask is why a check for a value less or equal than 0? Why not just 0 or why no just -1?

mavros_extras/src/plugins/distance_sensor.cpp
```diff
@@ -279,6 +280,7 @@ void DistanceSensorItem::range_cb(const sensor_msgs::Range::ConstPtr &msg)
 	uint8_t covariance_ = 0;
 
 	if (covariance > 0) covariance_ = covariance;
+	else if (variance_window_size <= 0) covariance_ = UINT8_MAX; // Assume unknown variance
```
vooon: Just check is zero, because negative window size is impossible.

mavros_extras/src/plugins/distance_sensor.cpp
```diff
@@ -330,6 +332,9 @@ DistanceSensorItem::Ptr DistanceSensorItem::create_item(DistanceSensorPlugin *ow
 		// lookup for numeric value
 		p->orientation = utils::sensor_orientation_from_str(orientation_str);
 
+	// Variance window
+	pnh.param("variance_window_size", p->variance_window_size, DEFAULT_WINDOW_SIZE);
```
vooon: Note: because of `size_t` change, you should provide type to param template, something like `pnh.param<unsigned int>(...)`.

TSC21:
@sfalexrog updates?

TSC21:
Closing as stale. @sfalexrog please reopen if you want to address @vooon request.

PR #1184, opened 2019-02-25
setpoint_position: accept set-position-target-global-int messages

rmackay9:
This PR attempts to allow mavros to accept [SET_POSITION_TARGET_GLOBAL_INT](https://mavlink.io/en/messages/common.html#SET_POSITION_TARGET_GLOBAL_INT) messages from the flight controller or GCS.  This is useful because it can allow a ground station operator to pass the position target to ROS (so that it can do the path planning) without using rviz.

Below is an image of a test using Mission Planner attached to an ArduPilot Rover with ROS running on a tx2.  I was able to select ROS using MP's upper-left vehicle selector and the right-mouse-button-click and select, "Fly To Here".  When mavros consumes the message it converts the latitude, longitude, altitude into ENU and publishes it to /mavros/setpoint_position/cmd_pos topic

setpoint-mp

Mavros's topic can be remapped to [ROS's base local planner's input](http://wiki.ros.org/base_local_planner) ([setup is described here](http://ardupilot.org/dev/docs/ros-object-avoidance.html)) so that it can accept the target.
rviz-navigation-path

While testing I found two issues which I'm hoping the reviewers (@vooon?) can give me a hand in resolving:

- something seems incorrect with the conversion from lat/lon to ENU.  The north-south values seem to be reversed.  If I click to the north of the vehicle (in mission planner) the target (as viewed in rviz) appears to the south of the vehicle (or more accurately to the right of the vehicle which is facing east).  The East-West calculations seem ok.
- I had to force the z-axis output to zero (see comment, "force z-axis to zero") because I'm using a rover and ROS's navigation didn't seem happy unless the target has position of zero.

Anyway, all feedback is welcome, thanks!

### DIFF
--- a/mavros/src/plugins/setpoint_position.cpp
+++ b/mavros/src/plugins/setpoint_position.cpp
@@ -19,9 +19,11 @@
 #include <eigen_conversions/eigen_msg.h>
 
 #include <geometry_msgs/PoseStamped.h>
+#include <geographic_msgs/GeoPointStamped.h>
 
 #include <mavros_msgs/SetMavFrame.h>
 #include <mavros_msgs/GlobalPositionTarget.h>
+#include <mavros_msgs/PositionTarget.h>
 
 #include <GeographicLib/Geocentric.hpp>
 //#include <GeographicLib/Geoid.hpp>
@@ -32,7 +34,7 @@ using mavlink::common::MAV_FRAME;
 /**
  * @brief Setpoint position plugin
  *
- * Send setpoint positions to FCU controller.
+ * Send and receive setpoint positions from FCU controller.
  */
 class SetpointPositionPlugin : public plugin::PluginBase,
 	private plugin::SetPositionTargetLocalNEDMixin<SetpointPositionPlugin>,
@@ -44,7 +46,8 @@ public:
 		sp_nh("~setpoint_position"),
 		spg_nh("~"),
 		tf_rate(50.0),
-		tf_listen(false)
+		tf_listen(false),
+		is_map_init(false)
 	{ }
 
 	void initialize(UAS &uas_)
@@ -71,6 +74,9 @@ public:
 			gps_sub = spg_nh.subscribe("global_position/global", 10, &SetpointPositionPlugin::gps_cb, this);
 			// Subscribe for current local ENU pose.
 			local_sub = spg_nh.subscribe("local_position/pose", 10, &SetpointPositionPlugin::local_cb, this);
+
+			// subscriber for global origin (aka map origin)
+			gp_origin_sub = spg_nh.subscribe("global_position/gp_origin", 10, &SetpointPositionPlugin::gp_origin_cb, this);
 		}
 		mav_frame_srv = sp_nh.advertiseService("mav_frame", &SetpointPositionPlugin::set_mav_frame_cb, this);
 
@@ -81,11 +87,16 @@ public:
 		} else {
 			mav_frame = utils::mav_frame_from_str(mav_frame_str);
 		}
+
+		// publish targets received from FCU
+		setpointg_pub = sp_nh.advertise<geometry_msgs::PoseStamped>("cmd_pos", 10);
 	}
 
 	Subscriptions get_subscriptions()
 	{
-		return { /* Rx disabled */ };
+		return {
+			make_handler(&SetpointPositionPlugin::handle_set_position_target_global_int),
+		};
 	}
 
 private:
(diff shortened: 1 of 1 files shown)


mavros/src/plugins/setpoint_position.cpp
```diff
@@ -258,6 +282,52 @@ class SetpointPositionPlugin : public plugin::PluginBase,
 		res.success = true;
 		return true;
 	}
+
+	/* -*- rx handler -*- */
+
+	/**
+	 * @brief handle SET_POSITION_TARGET_GLOBAL_INT mavlink msg
```
TSC21: `SET_` prefixes are usually used for messages you send to the FCU, not something you receive from. I understand that what you are doing is basically expect that the data coming from the GCS gets parsed here, since what MAVROS does is nothing more than serve as a proxy to forward messages, but we should not mess how we process data that comes from the FCU and the data that comes from the GCS. This last one should be handled separately.

If the data comes from the FCU, it needs to come in [`POSITION_TARGET_GLOBAL_INT`](https://mavlink.io/en/messages/common.html#POSITION_TARGET_GLOBAL_INT) messages.

TSC21: Btw, for specific GCS data:
- https://github.com/mavlink/mavros/projects/2: still empty but can this feature can be added as an extension to it;
- https://github.com/mavlink/mavros/pull/1117: gives a good example on how to separate streams.

rmackay9: @TSC21, thanks for the feedback and links.  I see what you're saying about keeping the handling of messages from the GCS separate from the vehicle.  I was planning to actually make it possible for the vehicle to send set-position-target-local-ned messages so that ROS's navigation can help the vehicle get around some obstacles close to the vehicle.

In any case, I'll think about it a bit more and I'm still trying to find the time to resolve the frame conversion issue in this PR.

rmackay9:
I think this PR is working now and I've done a [blog on ardupilot.org](https://discuss.ardupilot.org/t/rover-object-avoidance-using-ros-for-path-planning/39553) including a couple of videos ([auto mode](https://www.youtube.com/watch?v=u99qwQSl9Z4), [guided mode](https://www.youtube.com/watch?v=tqVH4lWk51Y)) showing how ROS's path planner can be used as an external navigation source for object avoidance.

@vooon, any chance this can be merged?

I'm happy to make changes but I probably need some more specific feedback on what needs changing.

I understand @TSC21's comments that we need to be careful of mixing up commands received from a GCS and those received from a vehicle but with [these enhancements to ArduPilot Rover](https://github.com/ArduPilot/ardupilot/pull/10775), the position target is now coming from the vehicle.

At the risk of overblowing this minor changes, I think it a nice enhancement that allows the flight code to remain in control and only use ROS as a navigation tool when it wants to.

TSC21:
@rmackay9 just because something is working for a particular case doesn't mean that we are doing things thr correct way. There is an API and an implementation logic that should always be respected, unless there's nothing we can do besides messing with the API logic.
As you can read on my review, I didn't just referenced the mixing of GCS and FCU messages. I also said that `SET_` prefix messages are sent to the flight controller, **not** coming from it. You can use the counter-part message to get the data from the FCU. Using the `SET_...` messages here is, as I said, breaking the API and the implementation logic of this package. Doesn't make sense to have a group of messages which usage is respecting the API implementation and then have a case where this is change with no aparent reason.

TSC21:
> I'm happy to make changes but I probably need some more specific feedback on what needs changing.

As a reviewer, I gave a specific feedback on what needs to be changed :) I am happy to help you make this implementation correct (not make it work, cause it seems to be).

rmackay9:
@TSC21, thanks again for the feedback.  So the objection is with the SET_TARGET_POSITION_GLOBAL_INT messages coming from the FCU?  I think that has only been the case because we haven't previously tried to make the flight controller the "master".  I.e. the flight controller has always been the "leaf".. it's been pushed around by upper level controllers within ROS.  I don't really see why there would be a restriction on a flight controller providing a position target to ROS's navigation controller... and if we accept that it is OK for a flight controller to provide a position target to ROS, I don't see how it would get it there unless it used the SET_TARGET_POSITION_GLOBAL_INT message (or the very similar SET_TARGET_POSITION_LOCAL_NED)..

TSC21:
@rmackay9 this is not a matter of what has been tried or not. Is what the API is set to do and changing that paradigm first needs a discussion on the Mavlink API. We can't just assume we can use a message in a certain way just becaude it works. Then we would be using all the messages the way we wanted.
You can actually read on the message description the following:
>"Used by an external controller to command the vehicle (manual controller or other system)."
If one wants to update the Mavlink protocol API, then one should suggest that same change first, and not apply the message in a certain way that contradicts the API.

vooon:
Also perhaps that should be in extras, as it have very specific usage.

sergioma295:
I am having a big problem. I have created a new issue. Can somebody help me please? #1248

PR #1195, opened 2019-03-06
Get topic local_position/velocity back from local_position/velocity_local 

Jaeyoung-Lim:
I have realized that the topic `local_position/velocity` has been replaced by `local_position/velocity_local`. 
This seems to have been triggered first by #1142 and a fix for this changed the topic name in #1160 

It took me some time to find this and fix it in a [package](https://github.com/Jaeyoung-Lim/mavros_controllers) I am using, and I believe this would be true for a lot of other packages that depend on mavros.

I don't understand what is the reasoning behind the change of the topic name to `velocity_local` rather than keeping it  as `velocity` while introducing the new body frame velocity as `velocity_body`. The change is not even documented in the [mavros wiki](http://wiki.ros.org/mavros), which most people go for reference on the list of topics.

### DIFF
--- a/mavros/src/plugins/local_position.cpp
+++ b/mavros/src/plugins/local_position.cpp
@@ -57,7 +57,7 @@ public:
 
 		local_position = lp_nh.advertise<geometry_msgs::PoseStamped>("pose", 10);
 		local_position_cov = lp_nh.advertise<geometry_msgs::PoseWithCovarianceStamped>("pose_cov", 10);
-		local_velocity_local = lp_nh.advertise<geometry_msgs::TwistStamped>("velocity_local", 10);
+		local_velocity_local = lp_nh.advertise<geometry_msgs::TwistStamped>("velocity", 10);
 		local_velocity_body = lp_nh.advertise<geometry_msgs::TwistStamped>("velocity_body", 10);
 		local_velocity_cov = lp_nh.advertise<geometry_msgs::TwistWithCovarianceStamped>("velocity_body_cov", 10);
 		local_accel = lp_nh.advertise<geometry_msgs::AccelWithCovarianceStamped>("accel", 10);


TSC21:
The change was made on purpose to differentiate the topics. What has to be done is address it on the docs.

Jaeyoung-Lim:
@TSC21  I understand the decision was made to make the package names more clear, but changing the topic name breaks backward compatibility of mavros.

TSC21:
This was a maintenance decision. It doesn't break compatibility because ROS developers are always able to remap the topics on ROS without having to change the source code.

Jaeyoung-Lim:
@TSC21 I don't agree that is enough.

The packages using the old topic will not give out any errors but simply have the velocity information empty in the package using the topic. (Which happened to me)

Would there be anyway to prevent this from happening? or is this simply something that the user is expected to debug?

TSC21:
> @TSC21 I don't agree that is enough.
> 
> The packages using the old topic will not give out any errors but simply have the velocity information empty in the package using the topic. (Which happened to me)
> 
> Would there be anyway to prevent this from happening? or is this simply something that the user is expected to debug?

This got messy with #1142 when the topic was replaced with body velocity instead of local velocity, which ended up being something worse in the end result, since you were expecting local velocity and instead you were getting body velocity. #1160 fixed this and made sure we have a differentiation of both velocities on the topic name. Again, a maintenance decision. So in the end, if you don't get velocity data, and as @vooon suggested, you need to verify why you are not getting the data. In this case, it's an API change, and we tend to reference API changes on the releases, even though we not always update it on the documentation.

Jaeyoung-Lim:
@TSC21 Thank you for the explanation. That works for me. Just as a user relying on the API, expected the topic names to be more stable.

PR #1207, opened 2019-03-26
TypoFix : Update apm_config.yaml

KiloNovemberDelta:
Seems to be a typo to fix. There is a space to remove for the setpoint_raw. Without it, apm.launch is not working.


KiloNovemberDelta:
It is already done at : https://github.com/mavlink/mavros/pull/1206. 
I will remove this PR.

PR #1218, opened 2019-04-13
Obstacle_distance plugin: NaN values reported as 0 distance

jmachuca77:
Fixed an issue when the laser scan topic contains NaN values they where being sent as 0 distances.


TSC21:
@mrivi can you review?

PR #1222, opened 2019-04-19
mavros/src/plugins/command.cpp: log if command's wait ack timeout

szobov:
It's confusing now, when you receive an answer: 4 (`MAV_RESULT_FAILED`) on command long, because after you dig into command processing side, in my case it's px4 firmware, you'll see that there is no possible answer `MAV_RESULT_FAILED` on the command.

This changes make it more explicit.

Thanks!

### DIFF
--- a/mavros/src/plugins/command.cpp
+++ b/mavros/src/plugins/command.cpp
@@ -42,6 +42,7 @@ public:
 	explicit CommandTransaction(uint16_t command) :
 		ack(),
 		expected_command(command),
+		// Default result if wait ack timeout
 		result(enum_value(mavlink::common::MAV_RESULT::FAILED))
 	{ }
 };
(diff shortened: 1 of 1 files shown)


mavros/src/plugins/command.cpp
```diff
@@ -180,6 +181,9 @@ class CommandPlugin : public plugin::PluginBase {
 			bool is_not_timeout = wait_ack_for(*ack_it);
 			lock.lock();
 
+			if (!is_not_timeout) {
+				ROS_WARN_NAMED("cmd", "CMD: Command %u -- wait ack timeout", command);
+			}
```
vooon: Probably better put that to `wait_ack_for()`, where you have command number in passed `CommandTransaction`.
That way you support both COMMAND_LONG and COMMAND_INT.

szobov: I can put it to `wait_ack_for()`, but actually this function is used only in `send_command_long_and_wait()`, so even if I put the changes, it will work only for `COMMAND_LONG`.
Should I do it?

vooon: I forgot that INT don't ACK... But i still think that better to warn in timeout func.

```cpp
if (tr.ack.wait(...) == timeout) {
    WARN(...);
    return false;
}
else {
    return true;
}

szobov: fixed

szobov: no, I should fix error, sorry

szobov: now it's ok

PR #1233, opened 2019-05-13
specialize transform_frame_ned_enu and transform_frame_enu_ned 

mrivi:
This PR makes possible the transformation between ENU and NED frames of vectors with NAN.

### DIFF
--- a/mavros/src/lib/ftf_frame_conversions.cpp
+++ b/mavros/src/lib/ftf_frame_conversions.cpp
@@ -57,6 +57,14 @@ static const auto NED_ENU_R = NED_ENU_Q.normalized().toRotationMatrix();
 static const auto AIRCRAFT_BASELINK_R = AIRCRAFT_BASELINK_Q.normalized().toRotationMatrix();
 
 /**
+ * @brief Use reflections instead of rotations for NED <-> ENU transformation
+ * to avoid NaN/Inf floating point pollution across different axes
+ * since in NED <-> ENU the axes are perfectly aligned.
+ */
+static const Eigen::PermutationMatrix<3> NED_ENU_REFLECTION_XY(Eigen::Vector3i(1,0,2));
+static const Eigen::DiagonalMatrix<double,3> NED_ENU_REFLECTION_Z(1,1,-1);
+
+/**
  * @brief Auxiliar matrices to Covariance transforms
  */
 using Matrix6d = Eigen::Matrix<double, 6, 6>;
@@ -79,12 +87,13 @@ Eigen::Quaterniond transform_orientation(const Eigen::Quaterniond &q, const Stat
 	}
 }
 
+
 Eigen::Vector3d transform_static_frame(const Eigen::Vector3d &vec, const StaticTF transform)
 {
 	switch (transform) {
 	case StaticTF::NED_TO_ENU:
 	case StaticTF::ENU_TO_NED:
-		return Eigen::Vector3d(vec.y(), vec.x(), -vec.z());
+		return NED_ENU_REFLECTION_XY * (NED_ENU_REFLECTION_Z * vec);
 
 	case StaticTF::AIRCRAFT_TO_BASELINK:
 	case StaticTF::BASELINK_TO_AIRCRAFT:
@@ -101,7 +110,8 @@ Covariance3d transform_static_frame(const Covariance3d &cov, const StaticTF tran
 	switch (transform) {
 	case StaticTF::NED_TO_ENU:
 	case StaticTF::ENU_TO_NED:
-		cov_out = cov_in * NED_ENU_Q;
+		cov_out = NED_ENU_REFLECTION_XY * (NED_ENU_REFLECTION_Z * cov_in * NED_ENU_REFLECTION_Z) *
+		          NED_ENU_REFLECTION_XY.transpose();
 		return cov_out_;
 
 	case StaticTF::AIRCRAFT_TO_BASELINK:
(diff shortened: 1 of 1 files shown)


mavros/src/lib/ftf_frame_conversions.cpp
```diff
@@ -84,7 +84,7 @@ Eigen::Vector3d transform_static_frame(const Eigen::Vector3d &vec, const StaticT
 	switch (transform) {
 	case StaticTF::NED_TO_ENU:
 	case StaticTF::ENU_TO_NED:
-		return NED_ENU_AFFINE * vec;
+		return Eigen::Vector3d(vec.y(), vec.x(), -vec.z());
```
TSC21: Add a check for NaNs here and apply the respective return depending on the result.

jkflying: Why do extra checks? This gives the same result, but without the floating point issues.

jkflying:
Specifically for NED <-> ENU we don't need to use a rotation, since the rotation matrix is:

| 0 | 1 | 0 |
| 1 | 0 | 0 |
| 0 | 0 |-1 |

Which results in simply a swap of x/y and negation of z.

(In practice there are some very slight rounding issues, due to the PI/2 -> quaternion -> affine conversion which can't be exact due to floating point. That doesn't change things for NaN, but exact values would at least fix Inf.)

This can also be derived, simply from the system definitions, since: 
NED x1=N, y1=E, z1=D
ENU x2=E, y2=N, z2=U
and U=-D
So, x1=y2, y1=x2, z1=-z2.
So it is fairly clear that to transform we need to swap x and y, and negate z.

So, this PR just gets rid of all of the multiplications with 0. Note that it is just specialized for **only** operating on vectors, and **only** for NED<->ENU. 

In summary, I don't see where this is incorrect, or where it might introduce issues. 


PS. I guess this fix will at some point be needed for covariances as well, since it is quite easy to have Inf on yaw there - no heading info from mag etc. - but we aren't passing covariances around for our work, so it doesn't affect us yet.

jkflying:
Another option would be to encode this in 2 reflections, using something like this:
```C++
/**
 * @brief Use reflections instead of rotations for NED <-> ENU transformation
 * to avoid NaN/Inf floating point pollution across different axes
 * since in NED <-> ENU the axes are perfectly aligned.
 */
static const Eigen::PermutationMatrix<3> NED_ENU_REFLECTION_1(Eigen::Vector3i(1,0,2));
static const Eigen::DiagonalMatrix<double,3> NED_ENU_REFLECTION_2(1,1,-1);

Eigen::Vector3d transform_static_frame(const Eigen::Vector3d &vec, const StaticTF transform)
{
	switch (transform) {
	case StaticTF::NED_TO_ENU:
	case StaticTF::ENU_TO_NED:
		return  NED_ENU_REFLECTION_1 * (NED_ENU_REFLECTION_2 * vec);

	case StaticTF::AIRCRAFT_TO_BASELINK:
	case StaticTF::BASELINK_TO_AIRCRAFT:
		return AIRCRAFT_BASELINK_AFFINE * vec;
	}
}
```

This this could be used quite easily with the covariance as well, in a symmetrical manner:
```C++

Covariance3d transform_static_frame(const Covariance3d &cov, const StaticTF transform)
{
	Covariance3d cov_out_;
	EigenMapConstCovariance3d cov_in(cov.data());
	EigenMapCovariance3d cov_out(cov_out_.data());

	switch (transform) {
	case StaticTF::NED_TO_ENU:
	case StaticTF::ENU_TO_NED:
		cov_out = NED_ENU_REFLECTION_1 * (NED_ENU_REFLECTION_2 * cov_in *
		                                  NED_ENU_REFLECTION_2.transpose()) *
		          NED_ENU_REFLECTION_1.transpose();
		return cov_out_;

	case StaticTF::AIRCRAFT_TO_BASELINK:
	case StaticTF::BASELINK_TO_AIRCRAFT:
		cov_out = cov_in * AIRCRAFT_BASELINK_Q;
		return cov_out_;
	}
}
```

Etc, for the 6d/9d covariances.

@TSC21 thoughts?

jkflying:
Ok, I updated the code to use the 2-reflection method for the Vector3d, and Covariance3/6/9d. I'm running Kinetic, and the tests seem to be good.
```
[ RUN      ] FRAME_TF.transform_static_frame__enu_to_ned_123
[       OK ] FRAME_TF.transform_static_frame__enu_to_ned_123 (0 ms)
[ RUN      ] FRAME_TF.transform_static_frame__ned_to_enu_123
[       OK ] FRAME_TF.transform_static_frame__ned_to_enu_123 (0 ms)
```
@TSC21 do you have any way of running this on a real system?

@vooon how is this solution? It seems a bit cleaner than doing direct component assignments, and scales properly to the other vector types.

mhkabir:
I like this solution! :+1: I can run on a real system, but will need some time. 

Vote to merge in the meanwhile - code looks good.

mhkabir:
Should we implement this for the aircraft-baselink (body frame) transformation as well?

TSC21:
Sames as @mhkabir. So merging.

TSC21:
> Should we implement this for the aircraft-baselink (body frame) transformation as well?

I don't think so since we just rotate one axis. So I don't think we get the issues as on the other rotation.

jkflying:
> I don't think so since we just rotate one axis. So I don't think we get the issues as on the other rotation.

I'm not sure - anything using a rotation matrix (or quaternion) would do this I think, if one of the vector values was strange. Floating points numbers are just a bit leaky of things like NaN and Inf, since they round stuff to the nearest ulp and thus the axes aren't 100% aligned.

I hadn't looked at the aircraft-baselink transform, but it could be done similarly, in fact it can be done with a single diagonal matrix since just 2 axes are made negative. I think maybe even just swapping all the uses of`AIRCRAFT_BASELINK_Q`/`AIRCRAFT_BASELINK_AFFINE` for uses of

`Eigen::DiagonalMatrix<double, 3> AIRCRAFT_BASELINK_M(AIRCRAFT_BASELINK_Q.toRotationMatrix().diagonal())`

should be sufficient. But this all needs some more testing, I'll try running with SITL tomorrow. We had it running fine before the reflection refactor, and tests are passing, but I don't have anything that uses the covariances, and maybe we should add tests with Inf/Nan in them.

PR #1252, opened 2019-06-17
Fix tf loop

Lukepahill:
There's a tf tree loop that's cased by broadcasting two static transforms between the same frames. Most ROS code ignores this but some nodes get very upset when they see a loop.

### DIFF
--- a/mavros/src/lib/uas_data.cpp
+++ b/mavros/src/lib/uas_data.cpp
@@ -54,8 +54,6 @@ UAS::UAS() :
 	// send static transform from "local_origin" (ENU) to "local_origin_ned" (NED)
 	publish_static_transform("local_origin", "local_origin_ned", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, M_PI_2)));
 	publish_static_transform("fcu", "fcu_frd", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, 0)));
-	// for being able to apply the transforms at the data coming from the FCU, the opposite transform is also required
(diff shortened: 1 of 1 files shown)


mhkabir:
Um, if you want to do the other transform, then just look up the inverse transform (switch target_frame and source_frame in the lookup call). I don't see any reason to have the loop in there.

TSC21:
> This shouldn't be necessary (TF should do it implicitly)

I don't know the reason, but I remember this was not working. But I know that's supposed to be true. Anyhow, this PR is not a solution if it doesn't handle the transform properly in the plugin.

Lukepahill:
Is it actually that the the odom pluggin needs a transform from local_origin_ned to fcu_frd? (https://github.com/mavlink/mavros/blob/master/mavros_extras/src/plugins/odom.cpp#L206)  Because currently MavROS does not provide this.  I can make that change, though I'm unsure how the frames are intended to be related.

jkflying:
This should be re-opened, it fixes TF cache issues.

TSC21:
@jkflying I merged it. Though, can you confirm that the `odom` plugin processes the transforms correctly when receiving ODOMETRY from the FCU? Thanks

jkflying:
@TSC21 I think we only do ODOMETRY in the other direction, camera -> ROS -> FCU. What's the use case for doing it in the other direction?

TSC21:
> @TSC21 I think we only do ODOMETRY in the other direction, camera -> ROS -> FCU. What's the use case for doing it in the other direction?

Actually if you check the firmware side you will see thaf we have ODOMETRY in the other direction as well. I have a use case in a project where is actually used in a SLAM pipeline, where the odometry input comes from the flight controller. I merged this as a short fix but I do think this breaks the `odom` plugin usage.

PR #1257, opened 2019-06-22
Add Mount Control plugin to control mount continuously

Jaeyoung-Lim:
I have been trying to control the gimbal continously while the vehicle is constantly being commanded with continuous setpoints in offboard mode. This can be used for image based object / human tracking control. 

I have been trying to use `actuator_setpoints` with mixer group = 2 but this led to some [problems](https://discuss.px4.io/t/on-realtime-control-of-gimbal-from-companion-computer/11468). However, to use actuator control setpoints, `vmount` needs to be disabled.

A better alternative is to use `MAV_CMD_DO_MOUNT_CONTROL` message as this utilizes the `vmount` running on the drone. However, the `command` plugin in mavros only has a service interface, which is not suited for controlling the gimbal continuously.

**Solution**
This PR adds a MountControl plugin, which streams a `MAV_CMD_DO_MOUNT_CONTROL` mavlink stream by subscribing to a topic `~mount_control/command`. A custom `MountControl.msg` was defined for this interafce.

**Test data / coverage**
The vehicle is flown in offboard mode to follow a circular trajectory while being streamed mount control setpoints. 
- log : https://review.px4.io/plot_app?log=b83e0ef5-3429-4a92-a062-3cd496808c03
- video: https://youtu.be/k99cxt65rx8

**Additional context**
This provides a alternative to the issues with controlling the gimbal with `actuator_control` messages as in https://github.com/PX4/Firmware/issues/12029
The video above is comparable to the video below where the gimbal is controlled by actuator setpoints: https://youtu.be/vEbUaGOuFjs

### DIFF
--- a/mavros_extras/src/plugins/mount_control.cpp
+++ b/mavros_extras/src/plugins/mount_control.cpp
@@ -1,0 +1,73 @@
+/**
+ * @brief Mouny Control plugin
+ * @file mount_control.cpp
+ * @author Jaeyoung Lim <jaeyoung@auterion.com>
+ *
+ * @addtogroup plugin
+ * @{
+ */
+/*
+ * Copyright 2018 Jaeyoung Lim.
+ *
+ * This file is part of the mavros package and subject to the license terms
+ * in the top-level LICENSE file of the mavros repository.
+ * https://github.com/mavlink/mavros/tree/master/LICENSE.md
+ */
+
+#include <mavros/mavros_plugin.h>
+
+#include <mavros_msgs/CompanionProcessStatus.h>
+
+namespace mavros {
+namespace extra_plugins {
+
+//! Mavlink enumerations
+using mavlink::common::MAV_TYPE;
+using mavlink::common::MAV_STATE;
+using mavlink::common::MAV_COMPONENT;
+using mavlink::common::MAV_MOUNT_MODE;
+using utils::enum_value;
+
+/**
+ * @brief Mount Control plugin
+ *
+ * Publishes Mission commands to control the camera or antenna mount.
+ * @see status_cb()
+ */
+class MountControlPlugin : public plugin::PluginBase {
+public:
+	MountControlPlugin() : PluginBase(),
+	status_nh("~mount_control")
+	{ }
+
+	void initialize(UAS &uas_)
+	{
+		PluginBase::initialize(uas_);
+
+		status_sub = status_nh.subscribe("command", 10, &MountControlPlugin::status_cb, this);
+	}
+
+	Subscriptions get_subscriptions()
+	{
+		return { /* Rx disabled */ };
+	}
+
+private:
+	ros::NodeHandle status_nh;
+	ros::Subscriber status_sub;
+
+	/**
+	 * @brief Send mount control commands to vehicle
+	 *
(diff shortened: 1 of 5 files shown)


mavros_extras/src/plugins/mount_control.cpp
```diff
@@ -0,0 +1,87 @@
...
+	 * @param req	received MountControl msg
+	 */
+	void command_cb(const mavros_msgs::MountControl::ConstPtr &req)
+	{
+		mavlink::common::msg::COMMAND_LONG cmd {};
+
+		const uint8_t tgt_sys_id = m_uas->get_tgt_system();
+		const uint8_t tgt_comp_id = m_uas->get_tgt_component();
+		cmd.target_system = tgt_sys_id;
+		cmd.target_component = tgt_comp_id;
```
vooon: You don't need intermediate `tgt_*` variables.

Jaeyoung-Lim: I have addressed the comments :)

mavros_extras/src/plugins/mount_control.cpp
```diff
@@ -0,0 +1,85 @@
...
+		cmd.target_system = m_uas->get_tgt_system();
+		cmd.target_component = m_uas->get_tgt_component();
+		cmd.command = enum_value(MAV_CMD::DO_MOUNT_CONTROL);
+		cmd.param1 = req->pitch;
+		cmd.param2 = req->roll;
+		cmd.param3 = req->yaw;
+		cmd.param4 = req->altitude; // 
+		cmd.param5 = req->latitude; // lattitude in degrees * 1E7
+		cmd.param6 = req->longitude; // longitude in degrees * 1E7
+		cmd.param7 = req->mode; // MAV_MOUNT_MODE
```
TSC21: Is this something you change on the fly, or rather a fixed mode from start ton beginning?

Jaeyoung-Lim: @TSC21 I guess this could depend on the application, but the intention was to define the mavros_msgs as close as to the real mavlink message. There is a separate mavlink for `DO_MOUNT_CONFIGURE` which is not implemented in this plugin

TSC21: Can you extend it already? Would make it more complete.

mavros_extras/src/plugins/mount_control.cpp
```diff
@@ -0,0 +1,85 @@
...
+//! Mavlink enumerations
+using mavlink::common::MAV_MOUNT_MODE;
+using mavlink::common::MAV_CMD;
+using utils::enum_value;
+
+/**
+ * @brief Mount Control plugin
+ *
+ * Publishes Mission commands to control the camera or antenna mount.
+ * @see status_cb()
```
TSC21: `command_cb()`

mavros_extras/src/plugins/mount_control.cpp
Jaeyoung-Lim: @TSC21 This would be possible already with the services using the command plugin. Do you think defining a new custom service in this plugin would be better approach

TSC21: @Jaeyoung-Lim yes that could work yes :)

Issue #1213, opened 2019-04-08
/mavros/global_position/local not always start at 0,0,0


elgarbe:
Hi, is global_position/local an odomotry topic or is it a ECEF position of the vehicle?
Some time that topic start from 0,0,0 like an odomety, but some times is start at ECEF coordinates of my vheicle. I'm using px4 in SITL.
What about /local_position/pose? what is the frame of this message?

----

### Issue details

_Please describe the problem, or desired feature_

### MAVROS version and platform

Mavros: ?0.18.4?
ROS: ?Kinetic?
Ubuntu: ?16.04?

### Autopilot type and version

[  ] ArduPilot
[  ] PX4

Version: ?3.7.1?

### Node logs

```
copy output of mavros_node. Usually console where you run roslaunch
```

### Diagnostics

```
place here result of:
rostopic echo -n1 /diagnostics
```

### Check ID

```
rosrun mavros checkid
```

TSC21:
> Hi, is global_position/local an odomotry topic or is it a ECEF position of the vehicle?

That topic is an odometry of the vehicle but computed from an ECEF origin, so the position is in the ECEF cartesian local frame.

> What about /local_position/pose? what is the frame of this message?

It's the local NED frame.

ericjohnson97:
I too have noticed that global_position/local topic sometimes starts at 0 0 0, and some times starts as ECEF. It seems very random to me which frame global_position/local will be as I will launch my gazebo + Arducopter SITL setup and get the 0 0 0 the first time, and the ECEF version the next. Its super strange to me, as I have been testing this for the last few hours and the reference frame can change with no modifications made to my setup.  I would appreciate any clues as to why this would happen. Thanks in advance!

TSC21:
@ericjohnson97 please provide the steps to reproduce it and an example output of what you are seeing so this can be debugged and fixed. Thank you

ericjohnson97:
give me a little bit. I still can't reliably produce either scenario. I am going to continue to try and track this one down. I will let you know what I find. Best!

elgarbe:
I couldn't find a way to get always 0,0,0 ir always ECEF. I launch iris/mavros and echo the message, if it is wrong I kill simulation and start over. Some times it works for several hours, but sometimes it fails a couple of times.
Let me know if you found a solution.

ericjohnson97:
[...]

PR #1269, opened 2019-07-09
fix for global_position/local sometimes having incorrect position

ericjohnson97:
This check makes sure the gp_origin is not incorrectly set when mavros is launch before aircraft has completed its boot up and obtained a gps fix.  The issue is described in this issue here https://github.com/mavlink/mavros/issues/1213

### DIFF
--- a/mavros/src/plugins/global_position.cpp
+++ b/mavros/src/plugins/global_position.cpp
@@ -325,7 +325,7 @@ private:
 						map_point.x(), map_point.y(), map_point.z());
 
 			// Set the current fix as the "map" origin if it's not set
-			if (!is_map_init) {
+			if (!is_map_init && fix->status.status == 0) {
 				map_origin.x() = fix->latitude;
 				map_origin.y() = fix->longitude;
 				map_origin.z() = fix->altitude;
--- a/mavros/src/plugins/global_position.cpp
+++ b/mavros/src/plugins/global_position.cpp
@@ -325,7 +325,7 @@ private:
 						map_point.x(), map_point.y(), map_point.z());
 
 			// Set the current fix as the "map" origin if it's not set
-			if (!is_map_init && fix->status.status == 0) {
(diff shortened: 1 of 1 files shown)


mavros/src/plugins/global_position.cpp
```diff
@@ -325,7 +325,7 @@ class GlobalPositionPlugin : public plugin::PluginBase {
 						map_point.x(), map_point.y(), map_point.z());
 
 			// Set the current fix as the "map" origin if it's not set
-			if (!is_map_init) {
+			if (!is_map_init && fix->status.status == 0) {
```
vooon: Are you sure? 0 means no fix, and from your description you need 3D fix at least.

ericjohnson97: So I've thought about this a little bit more, and there seems to be 2 different properties to denote gps fix. there is gps fix status which I believe denotes if the drone has any GPS signal. And then there is gps fix_type which denotes which kind of fix it is. If we use fix_type, the value would need to > 2. You are right my commit above works in sim testing, but is not robust. if there is a condition where the initial GPS coordinates are erroneous this would not work properly, but it would still be better than the current version, with no check. That being said, making the check rely on fix_type is the better route and I will rework it. It will probably be at least a week since I just went on vacation and I don't have the time or the proper setup to test a new version of the fix. 
Thanks!

vooon: No, code is fine, i just forgot that NavSatFix aready evaluate fix_type > 2 (meaning 3D FIX or augmented).
But it does not check is that augmented fix or not. So eq to 0 is good, but better to change it to:
```
&& fix->status.status >= sensor_msgs::NavSatStatus::STATUS_FIX
```

See: https://github.com/mavlink/mavros/blob/master/mavros/src/plugins/global_position.cpp#L163

vooon:
LGTM, Thanks!

PR #1287, opened 2019-07-31
libmavconn: simplify parse_buffer, and fix dropped_packets and parse_error counters

### LINKED ISSUE #1285: packet_rx_drop_count & parse_error are always 0 under /diagnostics
Seems like mavros tries to report parse_error and packet_rx_drop_count under /diagnostics (mavlink_diag.cpp), but their values are never incremented.

The problem seem to be in the way MAVConnInterface::parse_buffer tries to count those. It increments the m_status.parse_error, but m_status.parse_error is set to 0 in every call to mavlink_frame_char_buffer, and therefore when mavlink_diag calls get_status(), it always sees these values as 0.

Mavros: ?0.31.0?
ROS: ?Melodic?
Ubuntu: ?18.04?

[  ] ArduPilot
[ X ] PX4

kobytg17:
Currently the dropped_packets & parse_error counters are always published as 0 in mavros_diag.cpp.
This seems to be caused by using the wrong status struct.

Seems like mavros was editing m_status after mavlink_frame_char_buffer. This struct
seems to be the parsing state, and looks like it shouldn't be modified by the caller (for example
status->parse_error is zeroed out in the end of mavlink_frame_char_buffer).
Also, the crc & signature checks done in mavros seems redundant.

r_mavlink_status seems to be the struct that holds the mavlink connection information, therefore I
changed get_status to return it instead.

This fixes #1285.

### DIFF
--- a/libmavconn/include/mavconn/interface.h
+++ b/libmavconn/include/mavconn/interface.h
@@ -255,7 +255,7 @@ protected:
 	size_t conn_id;
 
 	inline mavlink::mavlink_status_t *get_status_p() {
-		return &m_status;
+		return &m_parse_status;
 	}
 
 	inline mavlink::mavlink_message_t *get_buffer_p() {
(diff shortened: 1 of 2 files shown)


TSC21:
@vooon let's bring this in.

vooon:
@TSC21 , @kobytg17 sorry, i had no time to review.
I'm not sure that separate state struct is needed, but i don't looked into mavlink parser for a year.

PR #1288, opened 2019-08-02
Fix bugs in output odometry pose transforms

nsteele:
- fixed transform of position data from parent frame to local frame (it was backwards)
- fixed body orientation w/respect to local frame
- fixed transform of velocity data from child frame to body frame (it was backwards)

builds fine and `catkin_make run_tests` is all passes (I saw ROS_ERRORs printed related to a sensor_orientation test but I don't think it's related to this code, and it didn't cause any test failures)


TSC21:
Can you explain the rationale here? Now I am sure that velocity tfs weren't tested (since PC4 doesn't fuse it). But I know that position and orientatiom were.

PR #1289, opened 2019-08-03
extras/obstacle_distance: add rx handler

jonathan7:
This makes the obstacle_distance plugin bidirectional by publishing obstacle_distance messages over a ros LaserScan topic.

### DIFF
--- a/mavros_extras/src/plugins/obstacle_distance.cpp
+++ b/mavros_extras/src/plugins/obstacle_distance.cpp
@@ -22,6 +22,8 @@ namespace mavros {
 namespace extra_plugins {
 //! Radians to degrees
 static constexpr double RAD_TO_DEG = 180.0 / M_PI;
+//! Degrees tp radians
+static constexpr double DEG_TO_RAD = M_PI / 180.0;
 //! Mavlink MAV_DISTANCE_SENSOR enumeration
 using mavlink::common::MAV_DISTANCE_SENSOR;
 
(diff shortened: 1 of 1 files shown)


mavros_extras/src/plugins/obstacle_distance.cpp
```diff
@@ -99,6 +109,44 @@ class ObstacleDistancePlugin : public plugin::PluginBase {
...
+		laser_scan->angle_max = laser_scan->angle_min + laser_scan->angle_increment * obstacle.distances.size();
+		uint16_t dist;
+		for (int i=obstacle.distances.size(); i>0; i--) {	// TODO better memory map?
+			dist = obstacle.distances[i-1];
+			if (dist == UINT16_MAX){
+				laser_scan->ranges.push_back(INFINITY); // Discard all 'unknown/invalid' ranges
+			} else {
+				laser_scan->ranges.push_back(dist * 1E-2);
+			}
+		}
```
vooon: I think something like that will be a bit better:
```cpp
laser_scan->ranges.reserve(obstacle.distances.size());
for (auto dist = obstacle.distabces.rbegin(); dist != obstacle.distances.rend(); dist++) {
    laser_scan->ranges.push_back((*dist == UINT16_MAX) ? INFINITY : *dist * 1E-2);
}

Comment (mhkabir) [reply mavros_extras/src/plugins/obstacle_distance.cpp]: It's also missing the transformation from local NED to local ENU.

Comment (vooon) [reply mavros_extras/src/plugins/obstacle_distance.cpp]: @mhkabir  We don't do any transforms on "send", so i don't think that we should transform anython on "receive" (and should remove transforms upper).

Comment (mhkabir) [reply mavros_extras/src/plugins/obstacle_distance.cpp]: Hmm, then that's wrong. We're kind of abusing the laserscan message in that case.

File: mavros_extras/src/plugins/obstacle_distance.cpp [hunk 2]
```diff
@@ -99,6 +109,39 @@ class ObstacleDistancePlugin : public plugin::PluginBase {
...
+			laser_scan->angle_increment = obstacle.increment_f * DEG_TO_RAD;
+		} else {
+			laser_scan->angle_increment = obstacle.increment * DEG_TO_RAD;
+		}
+
+		// Transform FRD to ENU (90 degrees left, then flip axes)
+		laser_scan->angle_min = (int(obstacle.angle_offset+270) % 360) * -DEG_TO_RAD;
+		laser_scan->angle_max = laser_scan->angle_min + laser_scan->angle_increment * obstacle.distances.size();
+		laser_scan->ranges.reserve(obstacle.distances.size());
+		for (auto dist = obstacle.distances.rbegin(); dist != obstacle.distances.rend(); dist++) {
```
TSC21: @jonathan7 please fix the indentation here.

TSC21:
Closing as stale.

Issue #766, opened 2017-08-08
Getting visible satellites count


okalachev:
### Issue details

Had not found a topic or something, where I can get ``satellites_visible`` value from ``GPS_RAW_INT`` message. Analysing the code did not help.
Is there a possibility to get it?

### MAVROS version and platform

Mavros: 0.18.7
ROS: Kinetic
Ubuntu: 16.04

### Autopilot type and version

[X] ArduPilot
[X] PX4

Version: Any

TSC21:
`rostopic echo -n1 /diagnostics`

vooon:
No such topic, because standard `sensor_msgs/NavSatFix` does not have a field for sat count.
Only one place where sat count exposed is diagnostics.

okalachev:
What is the standard way to take out the information from /diagnostic (for example, in Python)?
To iterate over DiagnosticArray, and find the value by the ``key`` field? Or there is any more straight solutions?

TSC21:
> By the way, it seems to me, that satellites count is an important thing, don't you consider to move it to a separate topic, for example ~global_position/raw/satellites_count?

Unless it is indispensable, we are not going to create a new msg just for that. `/diagnostics` topic publishes msgs of type `diagnostic_msgs/DiagnosticArray` which contains an array of `diagnostic_msgs/DiagnosticStatus`. So you just have to find the `KeyValue[]` (`value.key == Satellites visible`) and get the `value` to that key. That would be on DiagnosticStatus Message with name `mavros: GPS`.

okalachev:
I get it, generally.

> Unless it is indispensable, we are not going to create a new msg just for that.
Could it be of type ``std_msgs/UInt8``?

It seems a little bit strange, that the user has to find the data by textual description of it (``Satellites visible``), not by an identifier.

TSC21:
One thing you can do is propose adding that field to NavSatFix msg on the ROS repo. That way we could support it also.

TSC21:
@okalachev if you plan to propose a change for ROS msgs, probably you would want to propose the field to [`sensor_msgs/NavSatStatus`](http://docs.ros.org/lunar/api/sensor_msgs/html/msg/NavSatStatus.html) instead.

okalachev:
[...]

PR #1295, opened 2019-08-14
added amount of satellites to global_position/raw/

dayjaby:
This provides a simple solution for #766 : I simply publish the satellite count as ~global_position/raw/satellites with msg type std_msgs.UInt32

### DIFF
--- a/mavros/src/plugins/global_position.cpp
+++ b/mavros/src/plugins/global_position.cpp
@@ -159,9 +159,6 @@ private:
 	void handle_gps_raw_int(const mavlink::mavlink_message_t *msg, mavlink::common::msg::GPS_RAW_INT &raw_gps)
 	{
 		auto fix = boost::make_shared<sensor_msgs::NavSatFix>();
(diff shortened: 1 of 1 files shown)


mavros/src/plugins/global_position.cpp
```diff
@@ -156,6 +159,9 @@ class GlobalPositionPlugin : public plugin::PluginBase {
 	void handle_gps_raw_int(const mavlink::mavlink_message_t *msg, mavlink::common::msg::GPS_RAW_INT &raw_gps)
 	{
 		auto fix = boost::make_shared<sensor_msgs::NavSatFix>();
+		auto sat_cnt = boost::make_shared<std_msgs::UInt32>();
+		sat_cnt->data = raw_gps.satellites_visible;
+		raw_sat_pub.publish(sat_cnt);
```
vooon: I'd prefer to place that new code to the tail of the function.

dayjaby: Sounds reasonable! See latest commit.

PR #1299, opened 2019-08-19
fix corner case in obstacle distance plugin

baumanta:
The plugin left room to put in data which was too big for uint16_t, which would result in wrapped data.  Now each value is checked and replaced with UINT16_MAX if it was too big, also NANs are handled properly.

### DIFF
--- a/mavros_extras/src/plugins/obstacle_distance.cpp
+++ b/mavros_extras/src/plugins/obstacle_distance.cpp
@@ -67,10 +67,11 @@ private:
 		if (req->ranges.size() <= obstacle.distances.size()) {
 			// all distances from sensor will fit in obstacle distance message
 			for (int i = 0; i < req->ranges.size(); i++) {
-				if (std::isnan(req->ranges[i]) || req->ranges[i] >= (UINT16_MAX / 1e2)) {
+				float distance_cm = req->ranges[i] * 1e2;
+				if (std::isnan(distance_cm) || distance_cm >= UINT16_MAX || distance_cm < 0) {
 					obstacle.distances[i] = UINT16_MAX;
 				} else {
-					obstacle.distances[i] = req->ranges[i] * 1e2;		//!< [centimeters]
+					obstacle.distances[i] = static_cast<uint16_t>(distance_cm);
 				}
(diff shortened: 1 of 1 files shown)


baumanta:
FYI @mrivi @jkflying

mavros_extras/src/plugins/obstacle_distance.cpp
```diff
@@ -66,10 +66,14 @@ class ObstacleDistancePlugin : public plugin::PluginBase {
 
 		if (req->ranges.size() <= obstacle.distances.size()) {
 			// all distances from sensor will fit in obstacle distance message
-			Eigen::Map<Eigen::Matrix<uint16_t, Eigen::Dynamic, 1> > map_distances(obstacle.distances.data(), req->ranges.size());
-			auto cm_ranges = Eigen::Map<const Eigen::VectorXf>(req->ranges.data(), req->ranges.size()) * 1e2;
-			map_distances = cm_ranges.cast<uint16_t>();							//!< [centimeters]
-			std::fill(obstacle.distances.begin() + req->ranges.size(), obstacle.distances.end(), UINT16_MAX);    //!< fill the rest of the array values as "Unknown"
+			for (int i = 0; i < req->ranges.size(); i++) {
```
jkflying: I think the below would better read as something like:
```c++
float distance_cm = req->ranges[i] * 1e2;
if (std::isnan(distance_cm) || distance_cm >= UINT16_MAX || distance_cm < 0) {
    obstacle.distances[i] = UINT16_MAX;
} else {
    obstacle.distances[i] = static_cast<uint16_t>(distance_cm);
}
```

mavros_extras/src/plugins/obstacle_distance.cpp
```diff
@@ -81,12 +85,12 @@ class ObstacleDistancePlugin : public plugin::PluginBase {
 					if (req_index < req->ranges.size()) {
 						const float dist_m = req->ranges[req_index];
 						if (!std::isnan(dist_m)) {
```
jkflying: I think similar logic from above would make this easier to read

PR #1300, opened 2019-08-21
TF tree and odometry plugin

### LINKED ISSUE #1293: SITL error: unconnected trees
Hey,

with actual PX4 master and mavros i am getting following error

`ODOM: Ex: Could not find a connection between 'local_origin_ned' and 'fcu' because they are not part of the same tree.Tf has two or more unconnected trees.`

[frames.pdf](https://github.com/mavlink/mavros/files/3500358/frames.pdf)

kamilritz:
With this PR a proper tf tree is built and the odometry plugin simplified. This should simplify providing the FCU with external position/odometry information.

Fixes #1293

The suggested changes include:
- odom plugin is not publishing tfs anymore.
- multiple static tfs are published in uas_data.cpp. (includes a fix proposed by @mhkabir )
- transform between local frame and body frame is published by local_position. 
  (needed to change the frame_id in px4_config.yaml)
- change default frame_id in px4_config.yaml for all plugin.  (to be discussed)
- simplify the odom plugin to support less frame options.
- fixed inconsistent usage and naming of transformations.

UPDATED: tf tree (including frames that are published by Realsense T265 VIO sensor):
Screenshot from 2019-09-20 10-22-03

The send odometry messages, both to and from FCU, have been tested. If you want to test it your self, you have to wait for the Firmware PR that will solve some confusion in mavlink_receiver.cpp and mavlink_messages.cpp

@TSC21 @mhkabir

### DIFF
--- a/mavros/include/mavros/mavros_uas.h
+++ b/mavros/include/mavros/mavros_uas.h
@@ -272,6 +272,16 @@ public:
 	tf2_ros::StaticTransformBroadcaster tf2_static_broadcaster;
 
 	/**
+	 * @brief Add static transform. To publish all static transforms at once, we stack them in a std::vector.
+	 *
+	 * @param frame_id    parent frame for transform
+	 * @param child_id    child frame for transform
+	 * @param tr          transform
+	 * @param vector      vector of transforms
+	 */
+	void add_static_transform(const std::string &frame_id, const std::string &child_id, const Eigen::Affine3d &tr, std::vector<geometry_msgs::TransformStamped>& vector);
+
+	/**
 	 * @brief Publishes static transform.
 	 *
 	 * @param frame_id    parent frame for transform
--- a/mavros/launch/px4_config.yaml
+++ b/mavros/launch/px4_config.yaml
@@ -213,19 +213,9 @@ mocap:
 
 # odom
 odometry:
-  in:
-    frame_id: "odom"
-    child_frame_id: "base_link"
-    frame_tf:
-      local_frame: "local_origin_ned"
-      body_frame_orientation: "flu"
-  out:
-    frame_tf:
-      # available: check MAV_FRAME odometry local frames in
-      # https://mavlink.io/en/messages/common.html
-      local_frame: "vision_ned"
-      # available: ned, frd or flu (though only the tf to frd is supported)
-      body_frame_orientation: "frd"
+  fcu:
+    odom_parent_id_des: "map"    # desired parent frame rotation of the FCU's odometry
+    odom_child_id_des: "base_link"    # desired child frame rotation of the FCU's odometry
 
 # px4flow
 px4flow:
@@ -238,7 +228,7 @@ px4flow:
 vision_pose:
   tf:
     listen: false           # enable tf listener (disable topic subscribers)
-    frame_id: "map"
+    frame_id: "odom"
     child_frame_id: "vision_estimate"
     rate_limit: 10.0
 
@@ -259,12 +249,12 @@ wheel_odometry:
   wheel1: {x: 0.0, y:  0.15, radius: 0.05} # x-, y-offset (m,NED) and radius (m)
   send_raw: true              # send wheel's RPM and cumulative distance (~/wheel_odometry/rpm, ~/wheel_odometry/distance)
   send_twist: false           # send geometry_msgs/TwistWithCovarianceStamped instead of nav_msgs/Odometry
-  frame_id: "map"             # origin frame
+  frame_id: "odom"             # origin frame
   child_frame_id: "base_link" # body-fixed frame
   vel_error: 0.1              # wheel velocity measurement error 1-std (m/s)
   tf:
-    send: true
-    frame_id: "map"
+    send: false
+    frame_id: "odom"
     child_frame_id: "base_link"
 
 # vim:set ts=2 sw=2 et:
--- a/mavros/src/lib/uas_data.cpp
+++ b/mavros/src/lib/uas_data.cpp
@@ -52,12 +52,12 @@ UAS::UAS() :
 	}
 
 	// Publish helper TFs used for frame transformation in the odometry plugin
-	std::vector<geometry_msgs::TransformStamped> transformVector;
-	stack_static_transform("map", "map_ned", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, M_PI_2)),transformVector);
-	stack_static_transform("odom", "odom_ned", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, M_PI_2)),transformVector);
-	stack_static_transform("base_link", "base_link_frd", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, 0)),transformVector);
+	std::vector<geometry_msgs::TransformStamped> transform_vector;
+	add_static_transform("map", "map_ned", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, M_PI_2)),transform_vector);
+	add_static_transform("odom", "odom_ned", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, M_PI_2)),transform_vector);
+	add_static_transform("base_link", "base_link_frd", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, 0)),transform_vector);
 
-	tf2_static_broadcaster.sendTransform(transformVector);
+	tf2_static_broadcaster.sendTransform(transform_vector);
 }
 
 /* -*- heartbeat handlers -*- */
--- a/mavros/src/lib/uas_data.cpp
+++ b/mavros/src/lib/uas_data.cpp
@@ -51,9 +51,13 @@ UAS::UAS() :
 		ros::shutdown();
 	}
 
-	// send static transform from "local_origin" (ENU) to "local_origin_ned" (NED)
-	publish_static_transform("local_origin", "local_origin_ned", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, M_PI_2)));
-	publish_static_transform("fcu", "fcu_frd", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, 0)));
+	// Publish helper TFs used for frame transformation in the odometry plugin
+	std::vector<geometry_msgs::TransformStamped> transformVector;
+	stack_static_transform("map", "map_ned", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, M_PI_2)),transformVector);
+	stack_static_transform("odom", "odom_ned", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, M_PI_2)),transformVector);
+	stack_static_transform("base_link", "base_link_frd", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, 0)),transformVector);
+
+	tf2_static_broadcaster.sendTransform(transformVector);
 }
 
 /* -*- heartbeat handlers -*- */
@@ -221,6 +225,19 @@ sensor_msgs::NavSatFix::Ptr UAS::get_gps_fix()
 }
 
 /* -*- transform -*- */
+
+//! Stack static transform into vector
+void UAS::add_static_transform(const std::string &frame_id, const std::string &child_id, const Eigen::Affine3d &tr, std::vector<geometry_msgs::TransformStamped>& vector)
+{
+	geometry_msgs::TransformStamped static_transform;
+
+	static_transform.header.stamp = ros::Time::now();
+	static_transform.header.frame_id = frame_id;
+	static_transform.child_frame_id = child_id;
+	tf::transformEigenToMsg(tr, static_transform.transform);
+
+	vector.emplace_back(static_transform);
+}
 
 //! Publishes static transform
 void UAS::publish_static_transform(const std::string &frame_id, const std::string &child_id, const Eigen::Affine3d &tr)
--- a/mavros_extras/src/plugins/odom.cpp
+++ b/mavros_extras/src/plugins/odom.cpp
@@ -68,7 +68,7 @@ public:
 	Subscriptions get_subscriptions()
 	{
 		return {
-			       make_handler(&OdometryPlugin::handle_odom)
+			make_handler(&OdometryPlugin::handle_odom)
 		};
 	}
 
--- a/mavros_extras/src/plugins/odom.cpp
+++ b/mavros_extras/src/plugins/odom.cpp
@@ -31,103 +31,38 @@ using Matrix6d = Eigen::Matrix<double, 6, 6, Eigen::RowMajor>;
 /**
  * @brief Odometry plugin
  *
- * Sends odometry data to the FCU position and attitude estimators.
- * @see odom_cb()
+ * Sends odometry data to the FCU estimator and
+ * publishes odometry data that comes from FCU.
+ *
+ * This plugin is following ROS REP 147. Pose is expressed in parent frame. (Quaternion rotates from child to parent)
+ * The twist is expressed in the child frame.
+ *
+ * @see odom_cb()	transforming and sending odometry to fcu
+ * @see handle_odom()	receiving and transforming odometry from fcu
  */
 class OdometryPlugin : public plugin::PluginBase {
 public:
-	EIGEN_MAKE_ALIGNED_OPERATOR_NEW 	// XXX(vooon): added to try to fix #1223. Not sure that it is needed because class do not have Eigen:: fields.
+	EIGEN_MAKE_ALIGNED_OPERATOR_NEW		// XXX(vooon): added to try to fix #1223. Not sure that it is needed because class do not have Eigen:: fields.
 
 	OdometryPlugin() : PluginBase(),
 		odom_nh("~odometry"),
-		body_frame_orientation_in_desired("flu"),
-		body_frame_orientation_out_desired("frd"),
-		child_frame_id("base_link"),
-		frame_id("odom"),
-		local_frame_in("local_origin_ned"),
-		local_frame_out("vision_ned")
+		fcu_odom_parent_id_des("map"),
+		fcu_odom_child_id_des("base_link")
 	{ }
 
 	void initialize(UAS &uas_)
 	{
 		PluginBase::initialize(uas_);
 
-		odom_nh.param<std::string>("frame_id", frame_id, "odom");
-		odom_nh.param<std::string>("child_frame_id", child_frame_id, "base_link");
-
-		// frame tf params:
-		odom_nh.param<std::string>("in/frame_tf/local_frame", local_frame_in, "local_origin_ned");
-		odom_nh.param<std::string>("out/frame_tf/local_frame", local_frame_out, "vision_ned");
-		odom_nh.param<std::string>("in/frame_tf/body_frame_orientation", body_frame_orientation_in_desired, "flu");
-		odom_nh.param<std::string>("out/frame_tf/body_frame_orientation", body_frame_orientation_out_desired, "frd");
-
-		boost::algorithm::to_lower(local_frame_out);
-		boost::algorithm::to_lower(body_frame_orientation_out_desired);
+		// frame params:
+		odom_nh.param<std::string>("fcu/odom_parent_id_des", fcu_odom_parent_id_des, "map");
+		odom_nh.param<std::string>("fcu/odom_child_id_des", fcu_odom_child_id_des, "base_link");
 
 		// publishers
 		odom_pub = odom_nh.advertise<nav_msgs::Odometry>("in", 10);
 
 		// subscribers
 		odom_sub = odom_nh.subscribe("out", 10, &OdometryPlugin::odom_cb, this);
-
-		//! Map from param string to local MAV_FRAME
-		const std::unordered_map<std::string, MAV_FRAME> lf_map {
-			{ "vision_enu", MAV_FRAME::VISION_ENU },
-			{ "vision_ned", MAV_FRAME::VISION_NED },
-			{ "mocap_enu", MAV_FRAME::MOCAP_ENU },
-			{ "mocap_ned", MAV_FRAME::MOCAP_NED }
-		};
-
-		//! Map from param string to body MAV_FRAME
-		const std::unordered_map<std::string, MAV_FRAME> bf_map {
-			{ "ned", MAV_FRAME::BODY_NED },
-			{ "frd", MAV_FRAME::BODY_FRD },
-			{ "flu", MAV_FRAME::BODY_FLU }
-		};
-
-		// Determine input frame_id naming
-		if (local_frame_in == "local_origin_ned")
-			local_frame_orientation_in = local_frame_in;
-		else
-			ROS_FATAL_NAMED("odom", "ODOM: invalid input local frame \"%s\"", local_frame_in.c_str());
-
-		// Determine output frame_id naming - considering the ROS msg frame_id
-		// as "odom" by default
-		auto lf_it = lf_map.find(local_frame_out);
-		if (lf_it != lf_map.end()) {
-			lf_id = lf_it->second;
-			auto orient = local_frame_out.substr(local_frame_out.length() - 3);
-			if (orient != "enu") {
-				local_frame_orientation_out = "local_origin_" + orient;
-			} else {
-				local_frame_orientation_out = "local_origin";
-			}
-		}
-		else
-			ROS_FATAL_NAMED("odom", "ODOM: invalid ouput local frame \"%s\"", local_frame_out.c_str());
-
-		// Determine input child_frame_id naming
-		auto bf_it_in = bf_map.find(body_frame_orientation_in_desired);
-		if (bf_it_in != bf_map.end()) {
-			if (body_frame_orientation_in_desired != "flu")
-				body_frame_orientation_in_desired = "fcu_" + body_frame_orientation_in_desired;
-			else
-				body_frame_orientation_in_desired = "fcu";
-		}
-		else
-			ROS_FATAL_NAMED("odom", "ODOM: invalid input body frame orientation \"%s\"", body_frame_orientation_in_desired.c_str());
-
-		// Determine output child_frame_id naming
-		auto bf_it = bf_map.find(body_frame_orientation_out_desired);
-		if (bf_it != bf_map.end()) {
-			bf_id = bf_it->second;
-			if (body_frame_orientation_out_desired != "flu")
-				body_frame_orientation_out_desired = "fcu_" + body_frame_orientation_out_desired;
-			else
-				body_frame_orientation_in_desired = "fcu";
-		}
-		else
-			ROS_FATAL_NAMED("odom", "ODOM: invalid output body frame orientation \"%s\"", body_frame_orientation_out_desired.c_str());
 	}
 
 	Subscriptions get_subscriptions()
@@ -142,52 +77,22 @@ private:
 	ros::Publisher odom_pub;			//!< nav_msgs/Odometry publisher
 	ros::Subscriber odom_sub;			//!< nav_msgs/Odometry subscriber
 
-	std::string local_frame_in;			//!< orientation and source of the input local frame
-	std::string local_frame_out;			//!< orientation and source of the output local frame
-	std::string local_frame_orientation_in;		//!< orientation of the local frame (input data)
-	std::string local_frame_orientation_out;	//!< orientation of the local frame (output data)
-	std::string body_frame_orientation_in_desired;	//!< orientation of the body frame (input data)
-	std::string body_frame_orientation_out_desired;	//!< orientation of the body frame (output data)
-	std::string frame_id;				//!< parent frame identifier
-	std::string child_frame_id;			//!< child frame identifier
-
-	MAV_FRAME lf_id;				//!< local frame (pose) ID
-	MAV_FRAME bf_id;				//!< body frame (pose) ID
+	std::string fcu_odom_parent_id_des;			//!< desired orientation of the fcu odometry message's parent frame
+	std::string fcu_odom_child_id_des;			//!< desired orientation of the fcu odometry message's child frame
 
 	/**
-	 * @brief Lookup transforms
-	 * @todo Implement in a more general fashion in the API IOT apply frame transforms
-	 * This should also run in parallel on a thread
-	 * @param[in] &frame_id The parent frame of reference
-	 * @param[in] &child_frame_id The child frame of reference
-	 * @param[in] &local_frame_orientation The desired local frame orientation
-	 * @param[in] &body_frame_orientation The desired body frame orientation
-	 * @param[in,out] &tf_parent2local The affine transform from the parent frame to the local frame
-	 * @param[in,out] &tf_child2local The affine transform from the child frame to the local frame
-	 * @param[in,out] &tf_parent2body The affine transform from the parent frame to the body frame
-	 * @param[in,out] &tf_child2body The affine transform from the child frame to the body frame
+	 * @brief Lookup static transform with error handling
+	 * @param[in] &target The parent frame of the transformation you want to get
+	 * @param[in] &source The child frame of the transformation you want to get
+	 * @param[in,out] &tf_source2target The affine transform from the source to target
 	 */
-	void transform_lookup(const std::string &frame_id, const std::string &child_frame_id,
-		const std::string &local_frame_orientation, const std::string &body_frame_orientation,
-		Eigen::Affine3d &tf_parent2local, Eigen::Affine3d &tf_child2local,
-		Eigen::Affine3d &tf_parent2body, Eigen::Affine3d &tf_child2body)
+	void lookup_static_transform(const std::string &target, const std::string &source,
+				Eigen::Affine3d &tf_source2target)
 	{
 		try {
-			// transform lookup WRT local frame
-			tf_parent2local = tf2::transformToEigen(m_uas->tf2_buffer.lookupTransform(
-				frame_id, local_frame_orientation,
-				ros::Time(0)));
-			tf_child2local = tf2::transformToEigen(m_uas->tf2_buffer.lookupTransform(
-				child_frame_id, local_frame_orientation,
-				ros::Time(0)));
-
-			// transform lookup WRT body frame
-			tf_parent2body = tf2::transformToEigen(m_uas->tf2_buffer.lookupTransform(
-				frame_id, body_frame_orientation,
-				ros::Time(0)));
-			tf_child2body = tf2::transformToEigen(m_uas->tf2_buffer.lookupTransform(
-				child_frame_id, body_frame_orientation,
-				ros::Time(0)));
+			// transform lookup at current time.
+			tf_source2target = tf2::transformToEigen(m_uas->tf2_buffer.lookupTransform(
+							target, source, ros::Time(0)));
 		} catch (tf2::TransformException &ex) {
 			ROS_ERROR_THROTTLE_NAMED(1, "odom", "ODOM: Ex: %s", ex.what());
 			return;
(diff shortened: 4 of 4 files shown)


mavros_extras/src/plugins/odom.cpp
```diff
@@ -390,13 +265,6 @@ class OdometryPlugin : public plugin::PluginBase {
 		/* -*- ODOMETRY msg parser -*- */
 		msg.time_usec = odom->header.stamp.toNSec() / 1e3;
 
-		// [[[cog:
```
TSC21: Do not remove the cog code.

kamilritz: Added it back

kamilritz:
@TSC21 @mhkabir can we get this in

mavros/launch/px4_config.yaml
```diff
@@ -44,20 +44,20 @@ cmd:
 
 # global_position
 global_position:
-  frame_id: "map"             # origin frame
-  child_frame_id: "base_link" # body-fixed frame
+  frame_id: "local_origin"  # origin frame
+  child_frame_id: "fcu_utm" # body-fixed frame
```
TSC21: Why `fcu_utm`?

mavros/launch/px4_config.yaml
```diff
@@ -44,20 +44,20 @@ cmd:
...
-    frame_id: "map"  # TF frame_id
+    frame_id: "local_origin"  # TF frame_id
     global_frame_id: "earth"  # TF earth frame_id
-    child_frame_id: "base_link" # TF child_frame_id
+    child_frame_id: "fcu_utm" # TF child_frame_id
 
 # imu_pub
 imu:
-  frame_id: "base_link"
+  frame_id: "fcu"
```
TSC21: Why are you changing this frames? We should use the frame naming according to [ROS REP 105](https://www.ros.org/reps/rep-0105.html).

mavros/launch/px4_config.yaml
```diff
@@ -214,18 +214,12 @@ mocap:
...
   in:
-    frame_id: "odom"
-    child_frame_id: "base_link"
-    frame_tf:
-      local_frame: "local_origin_ned"
-      body_frame_orientation: "flu"
+    fcu_odom_parent_id_des: "local_origin"    # frame in which pose is expressed 
+    fcu_odom_child_id_des: "fcu"              # frame in which twist is expressed
   out:
-    frame_tf:
```
TSC21: Keep the `tf` namespace.

mavros/src/lib/uas_data.cpp
```diff
@@ -47,13 +47,19 @@ UAS::UAS() :
...
+			" | Run install_geographiclib_dataset.sh script in order to install Geoid Model dataset!");
 		ros::shutdown();
 	}
 
-	// send static transform from "local_origin" (ENU) to "local_origin_ned" (NED)
-	publish_static_transform("local_origin", "local_origin_ned", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, M_PI_2)));
-	publish_static_transform("fcu", "fcu_frd", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, 0)));
+	// send static transform that are part of the tf tree as a vector
+	std::vector<geometry_msgs::TransformStamped> transformVector;
+	stack_static_transform("local_origin", "local_origin_ned", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, M_PI_2)),transformVector);
```
TSC21: Let's make this `odom` and `odom_ned` and add also `stack_static_transform("map", "map_ned", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, M_PI_2)),transformVector);`.

mavros/src/lib/uas_data.cpp
```diff
@@ -47,13 +47,19 @@ UAS::UAS() :
...
+	stack_static_transform("vision_enu", "vision_ned", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, M_PI_2)),transformVector);
+	stack_static_transform("mocap_enu", "mocap_ned", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, M_PI_2)),transformVector);
+	stack_static_transform("fcu", "fcu_frd", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, 0)),transformVector);
```
TSC21: Let's make this `base_link` and `base_link_frd`.

mavros_extras/src/plugins/odom.cpp
```diff
@@ -31,103 +31,42 @@ using Matrix6d = Eigen::Matrix<double, 6, 6, Eigen::RowMajor>;
...
  *
- * Sends odometry data to the FCU position and attitude estimators.
- * @see odom_cb()
+ * Sends odometry data to the FCU position and attitude estimators and
+ * publishes odometry data that comes from FCU.
+ *
+ * This plugin is following ROS REP 147. Pose is expressed in parent frame. (Quaternion rotates from child to parent)
+ * The twist is expressed in the child frame.
+ *
+ * @see odom_cb()               transforming and sending odometry to fcu                -> IN  direction
```
TSC21: Indentation

mavros/launch/px4_config.yaml
kamilritz: can not remember any more. Any suggestions? it can not be fcu, since this would break the tree.

TSC21: `fcu_global`

mavros/launch/px4_config.yaml
kamilritz: Ok, I can change it back. Should we keep the frame_id that are not related to the tf tree as they were? But then they don't match the tf tree anymore.

TSC21: Let's build the TF tree in a way that we are sure that the naming matches the ROS REP 105. Everything that goes out of the "normal" frame convention should at least have the REP 105 frame naming as a prefix. Ex: `base_link` and `base_link_frd` or `odom` and `odom_ned`.

mavros/src/lib/uas_data.cpp
kamilritz: How should odom be connected to map?

TSC21: They won't be connected. `odom` has an arbitrary origin and can drift over time while `map` has a fixed known origin and should not have drift.

TSC21: Actually, according to the REP 105, `map` can be the parent of `odom`. Just consider that everything that comes from the PX4 side in the `local_position` plugin should be set with `map` as the local frame and `base_link` as the body frame. What comes from the `odom` plugin should have `odom` as the local frame. Regarding the body frame, I guess we need to make clear that this body frame is "different" from the body frame on the data coming from the `local_position` plugin. So maybe `base_link_odom`?

mavros_extras/src/plugins/odom.cpp
```diff
@@ -37,8 +37,8 @@ using Matrix6d = Eigen::Matrix<double, 6, 6, Eigen::RowMajor>;
  * This plugin is following ROS REP 147. Pose is expressed in parent frame. (Quaternion rotates from child to parent)
  * The twist is expressed in the child frame.
  *
- * @see odom_cb()               transforming and sending odometry to fcu                -> IN  direction
- * @see handle_odom()	receiving and transforming odometry from fcu    -> OUT direction
+ * @see odom_cb()		transforming and sending odometry to fcu		-> IN  direction
```
TSC21: this is still not correct

mavros/src/lib/uas_data.cpp
kamilritz: The body frames can be the same as long their is no tf transformation between odom and base_link

mavros_extras/src/plugins/odom.cpp
kamilritz: now it should

mavros/launch/px4_config.yaml
kamilritz: These frames are not used to produce tf transforms. They are just parent and child frames of the odom messages.

kamilritz:
@TSC21  The naming of the frames is now updated. Have a look at the tf tree in the PR description.

Providing logs that show that the whole pipeline is coherent and all the transformation are being right is not easy. I have checked the correctness of messages at all the points in the pipeline by moving the drone in all linear and angular direction and looking at the message at this points. Together with the merged [Firmware-PR](https://github.com/PX4/Firmware/pull/12793) the odom plugin provided correct odometry information (orientation, position and linear velocity are actively used in ECL) and also output the EKF odometry information accurately.

A lot of successful test flights have been done during the last several week with this odom plugin. You can find a log [here](https://review.px4.io/plot_app?log=bd994025-2312-4ee9-a7b2-f6ab2d65c6b8). I also redid some flight tests with the newest name changes in this PR.

mavros/include/mavros/mavros_uas.h
```diff
@@ -271,6 +271,16 @@ class UAS {
 	tf2_ros::TransformBroadcaster tf2_broadcaster;
 	tf2_ros::StaticTransformBroadcaster tf2_static_broadcaster;
 
+	/**
+	 * @brief Stack static transform.
```
mhkabir: Improve description

mavros/include/mavros/mavros_uas.h
```diff
@@ -271,6 +271,16 @@ class UAS {
...
+	 *
+	 * @param frame_id    parent frame for transform
+	 * @param child_id    child frame for transform
+	 * @param tr          transform
+	 * @param vector      vector of transforms
+	 */
+	void stack_static_transform(const std::string &frame_id, const std::string &child_id, const Eigen::Affine3d &tr, std::vector<geometry_msgs::TransformStamped>& vector);
```
mhkabir:
```suggestion
	void add_static_transform(const std::string &frame_id, const std::string &child_id, const Eigen::Affine3d &tr, std::vector<geometry_msgs::TransformStamped>& vector);
```

mavros/src/lib/uas_data.cpp
```diff
@@ -51,9 +51,14 @@ UAS::UAS() :
...
-	publish_static_transform("local_origin", "local_origin_ned", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, M_PI_2)));
-	publish_static_transform("fcu", "fcu_frd", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, 0)));
+	// send static transform that are part of the tf tree as a vector
+	std::vector<geometry_msgs::TransformStamped> transformVector;
+	stack_static_transform("map", "map_ned", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, M_PI_2)),transformVector);
+	stack_static_transform("odom", "odom_ned", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, M_PI_2)),transformVector);
+	stack_static_transform("base_link", "base_link_frd", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, 0)),transformVector);
+
+	tf2_static_broadcaster.sendTransform(transformVector);
+	ROS_WARN_STREAM("Publish static");
```
mhkabir:
```suggestion

```

mavros/src/lib/uas_data.cpp
```diff
@@ -51,9 +51,14 @@ UAS::UAS() :
 		ros::shutdown();
 	}
 
-	// send static transform from "local_origin" (ENU) to "local_origin_ned" (NED)
-	publish_static_transform("local_origin", "local_origin_ned", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, M_PI_2)));
-	publish_static_transform("fcu", "fcu_frd", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, 0)));
+	// send static transform that are part of the tf tree as a vector
```
mhkabir:
```suggestion
	// Publish helper TFs used for frame transformation in the odometry plugin
```

mavros/src/lib/uas_data.cpp
```diff
@@ -51,9 +51,14 @@ UAS::UAS() :
...
+	std::vector<geometry_msgs::TransformStamped> transformVector;
```
mhkabir:
```suggestion
	std::vector<geometry_msgs::TransformStamped> transform_vector;
```

mavros/src/lib/uas_data.cpp
```diff
@@ -222,6 +227,19 @@ sensor_msgs::NavSatFix::Ptr UAS::get_gps_fix()
 
 /* -*- transform -*- */
 
+//! Stack static transform into vector
+void UAS::stack_static_transform(const std::string &frame_id, const std::string &child_id, const Eigen::Affine3d &tr, std::vector<geometry_msgs::TransformStamped>& vector)
+{
+	geometry_msgs::TransformStamped static_transformStamped;
```
mhkabir:
```suggestion
	geometry_msgs::TransformStamped static_transform;
```

mavros_extras/src/plugins/odom.cpp
```diff
@@ -31,103 +31,42 @@ using Matrix6d = Eigen::Matrix<double, 6, 6, Eigen::RowMajor>;
...
 
 	OdometryPlugin() : PluginBase(),
 		odom_nh("~odometry"),
-		body_frame_orientation_in_desired("flu"),
-		body_frame_orientation_out_desired("frd"),
-		child_frame_id("base_link"),
-		frame_id("odom"),
-		local_frame_in("local_origin_ned"),
-		local_frame_out("vision_ned")
+		fcu_odom_parent_id_des("map"),
```
mhkabir: Why the `_des`?

mavros_extras/src/plugins/odom.cpp
```diff
@@ -31,103 +31,42 @@ using Matrix6d = Eigen::Matrix<double, 6, 6, Eigen::RowMajor>;
...
-		odom_nh.param<std::string>("in/frame_tf/body_frame_orientation", body_frame_orientation_in_desired, "flu");
-		odom_nh.param<std::string>("out/frame_tf/body_frame_orientation", body_frame_orientation_out_desired, "frd");
-
-		boost::algorithm::to_lower(local_frame_out);
-		boost::algorithm::to_lower(body_frame_orientation_out_desired);
+		// frame params:
+		odom_nh.param<std::string>("in/fcu_odom_parent_id_des", fcu_odom_parent_id_des, "map");
+		odom_nh.param<std::string>("out/fcu_odom_child_id_des", fcu_odom_child_id_des, "base_link");
+		odom_nh.param<std::string>("in/ext_odom_parent_id", ext_odom_parent_id, "odom");
+		odom_nh.param<std::string>("out/ext_odom_child_id", ext_odom_child_id, "base_link");
```
mhkabir: Why is this in the out namespace? Can you explain these frames quickly? We can put that in the mavros wiki.

mhkabir: It seems to me that FCU ones should be `out` and the Ext Odom ones should be `in`

mavros_extras/src/plugins/odom.cpp
```diff
@@ -142,52 +81,25 @@ class OdometryPlugin : public plugin::PluginBase {
...
-	std::string body_frame_orientation_out_desired;	//!< orientation of the body frame (output data)
-	std::string frame_id;				//!< parent frame identifier
-	std::string child_frame_id;			//!< child frame identifier
-
-	MAV_FRAME lf_id;				//!< local frame (pose) ID
-	MAV_FRAME bf_id;				//!< body frame (pose) ID
+	std::string fcu_odom_parent_id_des;			//!< desired orientation of the child frame (input data)
+	std::string fcu_odom_child_id_des;			//!< desired orientation of the body frame (input data)
+	std::string ext_odom_parent_id;				//!< parent frame of the ext odometry msg
+	std::string ext_odom_child_id;				//!< child frame of the ext odometry msg
```
mhkabir: This information is already in the Odometry message. Why do we need to set the frame manually again?

mavros_extras/src/plugins/odom.cpp
```diff
@@ -142,52 +81,25 @@ class OdometryPlugin : public plugin::PluginBase {
...
-			// transform lookup WRT body frame
-			tf_parent2body = tf2::transformToEigen(m_uas->tf2_buffer.lookupTransform(
-				frame_id, body_frame_orientation,
-				ros::Time(0)));
-			tf_child2body = tf2::transformToEigen(m_uas->tf2_buffer.lookupTransform(
-				child_frame_id, body_frame_orientation,
-				ros::Time(0)));
+			tf_source2target = tf2::transformToEigen(m_uas->tf2_buffer.lookupTransform(
+							target, source,
+							ros::Time(0)));
```
mhkabir: Is this only ever used for static lookups? If not, then **correct timestamps** should be used, not latest.

mavros_extras/src/plugins/odom.cpp
```diff
@@ -198,35 +110,26 @@ class OdometryPlugin : public plugin::PluginBase {
...
+	 *
+	 * Callback to odometry that is sent from FCU by mavlink_messages. It expects
+	 * the parent frame of the msg to be "local_origin_ned" and the child frame to be "fcu_frd".
+	 * The odometry message gets rotated such that parent frame and child frame are the desired
+	 * frames set in the rosparams "odometry/in/fcu_odom_*_id_des" (set in px4_config.yaml). If these frames are not part
+	 * of the usual tf tree, the user is supposed to add these frames to the existing tf tree.
+	 *
 	 * @param msg	Received Mavlink msg
-	 * @param att	ATTITUDE msg
+	 * @param att	ODOMETRY msg
```
mhkabir:
```suggestion
	 * @param odom_msg	ODOMETRY msg
```

mavros_extras/src/plugins/odom.cpp
```diff
@@ -142,52 +81,25 @@ class OdometryPlugin : public plugin::PluginBase {
...
+	 * @brief Lookup transform with error handling
+	 * @param[in] &target The parent frame of the transformation you want to get
+	 * @param[in] &source The child frame of the transformation you want to get
+	 * @param[in,out] &tf_source2target The affine transform from the source to target
 	 */
-	void transform_lookup(const std::string &frame_id, const std::string &child_frame_id,
-		const std::string &local_frame_orientation, const std::string &body_frame_orientation,
-		Eigen::Affine3d &tf_parent2local, Eigen::Affine3d &tf_child2local,
-		Eigen::Affine3d &tf_parent2body, Eigen::Affine3d &tf_child2body)
+	void transform_lookup(const std::string &target, const std::string &source,
```
mhkabir:
```suggestion
	void lookup_transform(const std::string &target, const std::string &source,
```

mavros_extras/src/plugins/odom.cpp
```diff
@@ -198,35 +110,26 @@ class OdometryPlugin : public plugin::PluginBase {
 	 * @brief Handle ODOMETRY MAVlink message.
 	 *
 	 * Message specification: https://mavlink.io/en/messages/common.html#ODOMETRY
+	 *
+	 * Callback to odometry that is sent from FCU by mavlink_messages. It expects
+	 * the parent frame of the msg to be "local_origin_ned" and the child frame to be "fcu_frd".
+	 * The odometry message gets rotated such that parent frame and child frame are the desired
+	 * frames set in the rosparams "odometry/in/fcu_odom_*_id_des" (set in px4_config.yaml). If these frames are not part
+	 * of the usual tf tree, the user is supposed to add these frames to the existing tf tree.
```
mhkabir: "If these frames are not part of the usual tf tree, the user is supposed to add these frames to the existing tf tree."

This is unclear. I don't see why it needs to be added to the tree manually, and how this is supposed to work.

mavros_extras/src/plugins/odom.cpp
```diff
@@ -198,35 +110,26 @@ class OdometryPlugin : public plugin::PluginBase {
 	 * @brief Handle ODOMETRY MAVlink message.
 	 *
 	 * Message specification: https://mavlink.io/en/messages/common.html#ODOMETRY
+	 *
+	 * Callback to odometry that is sent from FCU by mavlink_messages. It expects
+	 * the parent frame of the msg to be "local_origin_ned" and the child frame to be "fcu_frd".
```
mhkabir:
```suggestion
	 * the parent frame of the mavlink msg to be "local_origin_ned" and the child frame to be "fcu_frd".
```

Please confirm^

mavros_extras/src/plugins/odom.cpp
```diff
@@ -198,35 +110,26 @@ class OdometryPlugin : public plugin::PluginBase {
 	 * @brief Handle ODOMETRY MAVlink message.
 	 *
 	 * Message specification: https://mavlink.io/en/messages/common.html#ODOMETRY
+	 *
+	 * Callback to odometry that is sent from FCU by mavlink_messages. It expects
```
mhkabir:
```suggestion
	 * Callback for mavlink ODOMETRY messages sent from the FCU. It expects
```

mavros_extras/src/plugins/odom.cpp
```diff
@@ -198,35 +110,26 @@ class OdometryPlugin : public plugin::PluginBase {
...
-
 		/**
-		 * Required affine rotations to apply transforms
+		 * Required rotations to transform the FCU's odometry msg tto desired parent and child frame
 		 */
-		Eigen::Affine3d tf_parent2local;
-		Eigen::Affine3d tf_child2local;
-		Eigen::Affine3d tf_parent2body;
-		Eigen::Affine3d tf_child2body;
+		Eigen::Affine3d tf_parent2parentDes;
```
mhkabir:
```suggestion
		Eigen::Affine3d tf_parent2parent_des;
```

mavros_extras/src/plugins/odom.cpp
```diff
@@ -198,35 +110,26 @@ class OdometryPlugin : public plugin::PluginBase {
...
+		Eigen::Affine3d tf_child2childDes;
```
mhkabir:
```suggestion
		Eigen::Affine3d tf_child2child_des;
```

mavros_extras/src/plugins/odom.cpp
```diff
@@ -198,35 +110,26 @@ class OdometryPlugin : public plugin::PluginBase {
...
 
-		//! Lookup to the required trans
-		transform_lookup("local_origin_ned", "fcu_frd",
-			local_frame_orientation_in, body_frame_orientation_in_desired,
-			tf_parent2local, tf_child2local, tf_parent2body, tf_child2body);
+		transform_lookup( fcu_odom_parent_id_des, "map_ned", tf_parent2parentDes );
+		transform_lookup( fcu_odom_child_id_des,  "base_link_frd",          tf_child2childDes );
```
mhkabir: Formatting is odd, please run the linter.

mavros_extras/src/plugins/odom.cpp
```diff
@@ -312,15 +207,11 @@ class OdometryPlugin : public plugin::PluginBase {
...
-		Eigen::Affine3d tf_child2body;
+		Eigen::Affine3d tf_parent2parentDes;
+		Eigen::Affine3d tf_child2childDes;
 
-		//! Build 6x6 pose covariance matrix to be transformed and sent
-		transform_lookup(odom->header.frame_id, odom->child_frame_id,
-			local_frame_orientation_out, body_frame_orientation_out_desired,
-			tf_parent2local, tf_child2local, tf_parent2body, tf_child2body);
+		transform_lookup( "odom_ned", ext_odom_parent_id, tf_parent2parentDes );
+		transform_lookup( "base_link_frd",    ext_odom_child_id,  tf_child2childDes );
```
mhkabir: Lint

mavros_extras/src/plugins/odom.cpp
```diff
@@ -341,46 +232,30 @@ class OdometryPlugin : public plugin::PluginBase {
 		Matrix6d r_vel = Matrix6d::Zero();	//!< Zero initialized velocity 6-D Covariance matrix. WRT child_frame_id
 
 		mavlink::common::msg::ODOMETRY msg {};
+		msg.frame_id = utils::enum_value(MAV_FRAME::VISION_NED);
```
mhkabir: We deprecated this. https://mavlink.io/en/messages/common.html#MAV_FRAME_VISION_NED

mavros_extras/src/plugins/odom.cpp
```diff
@@ -341,46 +232,30 @@ class OdometryPlugin : public plugin::PluginBase {
...
+		msg.child_frame_id = utils::enum_value(MAV_FRAME::BODY_FRD);
 
 		/**
-		 * Position and orientation are in the same frame as frame_id.
-		 * For a matter of simplicity, and given the existent MAV_FRAME
-		 * enum values, the default frame_id will be a local frame of
-		 * reference, so the pose is WRT a local frame.
+		 * Position parsing from odometry's parent frame to "vision_ned" frame.
```
mhkabir: Comments lies, see above. Let's use LOCAL_FRD instead of vision_ned to be consistent

mavros_extras/src/plugins/odom.cpp
```diff
@@ -341,46 +232,30 @@ class OdometryPlugin : public plugin::PluginBase {
...
-		if (odom->child_frame_id == "world" || odom->child_frame_id == "odom") {
-			// the child_frame_id would be the same reference frame as frame_id
-			set_tf(tf_child2local, lf_id);
-		}
-		else {
-			// the child_frame_id would be the WRT a body frame reference
-			set_tf(tf_child2body, bf_id);
-		}
+		/**
+		 * Linear and angular velocities has to be fcu_frd frame
```
mhkabir: LOCAL_FRD?

mavros_extras/src/plugins/odom.cpp
kamilritz: We receive odometry message with a parent and a child frame. These child frames are most likely not the ones that we want to send to fcu or to ROS. So will rotate the odometry message components such that they are in the right frame. Notice here that we only apply the rotation and do not take into account any translation. This is handled in the ECL/EKF

mavros_extras/src/plugins/odom.cpp
kamilritz: You are right. The in and out directions are not always used properly. Thank you very much for noticing this.

kamilritz: Would you suggest to remove the in and out namespace? I think the parameter's name are clear enough.

mavros_extras/src/plugins/odom.cpp
kamilritz: But the frame names differs between PX4 and ROS side. e.g. we receive odom message from fcu. parent frame will be local_origin(_ned) and we want to look up transform from map_ned to map.

mavros_extras/src/plugins/odom.cpp
kamilritz: The external odometry sensor or mocap system will most likely introduce two frames. A reference frame with a for us unknown orientation. And the same for the link on the body. The user of the plugin has to specifiy how these frames are related to the odom reference frame as well as to the base_link. With this information we will be able to roatete the odometry message in the frames we will expect on the px4 side.

mavros_extras/src/plugins/odom.cpp
mhkabir: > These child frames are most likely not the ones that we want to send to fcu or to ROS.

mhkabir: > These child frames are most likely not the ones that we want to send to fcu or to ROS.

Why not? For example, my VIO system publishes an odometry message with parent_frame_id = mission, child_frame_id = base_link. I want to send this to the FCU. What is the use-case for not wanting to send this?

>  Notice here that we only apply the rotation and do not take into account any translation. This is handled in the ECL/EKF

Also this sounds wrong. If you're going to use the TF lookup, you should be applying the full transform, not just the rotation. Can you explain the rationale here?

mavros_extras/src/plugins/odom.cpp
mhkabir: I think the in/out is still good to keep. Everything is quite confusing as-is, and over-clarifying does not hurt in this case.

mavros_extras/src/plugins/odom.cpp
mhkabir: Yes, but I think we don't need it for the ext odom one, because the frame information already comes in the message, and that is what we should send to FCU.

mavros_extras/src/plugins/odom.cpp
kamilritz: @mhkabir  When using a new version of uncrustify I get the following error that is related with the uncrustify config:
mavros/tools/uncrustify-cpp.cfg:38 Unknown symbol 'align_number_left'
The inconvenient thing is that on Ubuntu 18.04 you have to install uncrustify from old binaries to get it to work.

mavros_extras/src/plugins/odom.cpp
kamilritz: > Also this sounds wrong. If you're going to use the TF lookup, you should be applying the full transform, not just the rotation. Can you explain the rationale here?

The linear offset of the sensor is compensated in the EKF2 by the EKF_EV_POS_* params. Don't think about these rotation as transforming the odometry message, but rather as changing the frame orientation that is set by the creator of the VIO system. E.g. you could design your VIO system to have a world z axis that is not aligned with gravity, why ever one want to do so.

mavros_extras/src/plugins/odom.cpp
kamilritz: Yes, you are right. This is ros side only

mavros_extras/src/plugins/odom.cpp
kamilritz: @mhkabir How do I get the LOCAL_FRD frame? Do I have to add it manually in enum_to_string.cpp?

mavros_extras/src/plugins/odom.cpp
kamilritz: Velocities are expressed in body frame. So the right frame would be base_link_frd

kamilritz:
@mhkabir I put your feedback in. Please have a look and tell me if you are okay with this.

mavros/src/lib/uas_data.cpp
```diff
@@ -51,9 +51,14 @@ UAS::UAS() :
...
-	publish_static_transform("local_origin", "local_origin_ned", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, M_PI_2)));
-	publish_static_transform("fcu", "fcu_frd", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, 0)));
+	// send static transform that are part of the tf tree as a vector
+	std::vector<geometry_msgs::TransformStamped> transform_vector;
+	add_static_transform("map", "map_ned", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, M_PI_2)),transform_vector);
+	add_static_transform("odom", "odom_ned", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, M_PI_2)),transform_vector);
+	add_static_transform("base_link", "base_link_frd", Eigen::Affine3d(ftf::quaternion_from_rpy(M_PI, 0, 0)),transform_vector);
+
+	tf2_static_broadcaster.sendTransform(transform_vector);
+	ROS_WARN_STREAM("Publish static");
```
TSC21: @kamilritz the ROS_WARN_STREAM is still appearing

mavros_extras/src/plugins/odom.cpp
TSC21: @kamilritz please fix this as well

mavros_extras/src/plugins/odom.cpp
```diff
@@ -198,35 +104,24 @@ class OdometryPlugin : public plugin::PluginBase {
...
-		Eigen::Affine3d tf_child2body;
+		Eigen::Affine3d tf_parent2parent_des;
+		Eigen::Affine3d tf_child2child_des;
 
-		//! Lookup to the required trans
-		transform_lookup("local_origin_ned", "fcu_frd",
-			local_frame_orientation_in, body_frame_orientation_in_desired,
-			tf_parent2local, tf_child2local, tf_parent2body, tf_child2body);
+		lookup_static_transform( fcu_odom_parent_id_des, "map_ned", tf_parent2parent_des );
+		lookup_static_transform( fcu_odom_child_id_des,  "base_link_frd",          tf_child2child_des );
```
TSC21: please fix this indentation

mavros/src/lib/uas_data.cpp
kamilritz: thanks, it is removed now

mavros_extras/src/plugins/odom.cpp
TSC21: also, why the need for a whitespace before and after the arguments?

kamilritz:
@TSC21 LGTM

mhkabir:
Please don't merge yet. I want to review again - not fully clear on all points yet.

mavros/launch/px4_config.yaml
```diff
@@ -68,7 +68,7 @@ imu:
 local_position:
   frame_id: "map"
   tf:
-    send: false
+    send: true
```
mhkabir: I don't support this being enabled by default.

mhkabir: This can have unexpected repercussions for systems where the map transform is already published by something other than mavros. Please revert, and we can put in a doc note.

mhkabir: I don't support this being enabled by default.

This can have unexpected repercussions for systems where the map transform is already published by something other than mavros. Please revert, and we can put in a doc note.

mavros/src/lib/uas_data.cpp
mhkabir: Not resolved.

mavros/src/lib/uas_data.cpp
```diff
@@ -222,6 +226,19 @@ sensor_msgs::NavSatFix::Ptr UAS::get_gps_fix()
...
+void UAS::add_static_transform(const std::string &frame_id, const std::string &child_id, const Eigen::Affine3d &tr, std::vector<geometry_msgs::TransformStamped>& vector)
+{
+	geometry_msgs::TransformStamped static_transform;
+
+	static_transform.header.stamp = ros::Time::now();
+	static_transform.header.frame_id = frame_id;
+	static_transform.child_frame_id = child_id;
+	tf::transformEigenToMsg(tr, static_transform.transform);
+
+	vector.push_back(static_transform);
```
mhkabir:
```suggestion
	vector.emplace_back(static_transform);
```

mavros_extras/src/plugins/odom.cpp
mhkabir: > E.g. you could design your VIO system to have a world z axis that is not aligned with gravity, why ever one want to do so.

This is not something the plugin should be dealing with, in my opinion. This should be the bare minimum needed to translate between ROS and the mavlink message. If you system has such a weird convention of not using a gravity-aligned reference frame, I would expect you to pre-rotate the measurement before sending it into the plugin.

mavros_extras/src/plugins/odom.cpp
```diff
@@ -31,103 +31,38 @@ using Matrix6d = Eigen::Matrix<double, 6, 6, Eigen::RowMajor>;
...
-		odom_nh.param<std::string>("in/frame_tf/local_frame", local_frame_in, "local_origin_ned");
-		odom_nh.param<std::string>("out/frame_tf/local_frame", local_frame_out, "vision_ned");
-		odom_nh.param<std::string>("in/frame_tf/body_frame_orientation", body_frame_orientation_in_desired, "flu");
-		odom_nh.param<std::string>("out/frame_tf/body_frame_orientation", body_frame_orientation_out_desired, "frd");
-
-		boost::algorithm::to_lower(local_frame_out);
-		boost::algorithm::to_lower(body_frame_orientation_out_desired);
+		// frame params:
+		odom_nh.param<std::string>("fcu/odom_parent_id_des", fcu_odom_parent_id_des, "map");
+		odom_nh.param<std::string>("fcu/odom_child_id_des", fcu_odom_child_id_des, "base_link");
```
mhkabir: Note to self: these are the ROS frames the mavlink ODOMETRY data from the FCU is published in.

mavros_extras/src/plugins/odom.cpp
```diff
@@ -142,52 +77,22 @@ class OdometryPlugin : public plugin::PluginBase {
...
-	std::string local_frame_orientation_in;		//!< orientation of the local frame (input data)
-	std::string local_frame_orientation_out;	//!< orientation of the local frame (output data)
-	std::string body_frame_orientation_in_desired;	//!< orientation of the body frame (input data)
-	std::string body_frame_orientation_out_desired;	//!< orientation of the body frame (output data)
-	std::string frame_id;				//!< parent frame identifier
-	std::string child_frame_id;			//!< child frame identifier
-
-	MAV_FRAME lf_id;				//!< local frame (pose) ID
-	MAV_FRAME bf_id;				//!< body frame (pose) ID
+	std::string fcu_odom_parent_id_des;			//!< desired orientation of the child frame (input data)
```
mhkabir: Comment is wrong? This should say "output data"

mavros_extras/src/plugins/odom.cpp
```diff
@@ -142,52 +77,22 @@ class OdometryPlugin : public plugin::PluginBase {
...
+	std::string fcu_odom_child_id_des;			//!< desired orientation of the body frame (input data)
```
mhkabir: Same as above

mavros_extras/src/plugins/odom.cpp
```diff
@@ -198,35 +103,24 @@ class OdometryPlugin : public plugin::PluginBase {
 	 * @brief Handle ODOMETRY MAVlink message.
 	 *
 	 * Message specification: https://mavlink.io/en/messages/common.html#ODOMETRY
+	 *
+	 * Callback for mavlink ODOMETRY messages sent from the FCU. It will be received with parent_frame
+	 * "local_origin_ned" (aka "map_ned") and child_frame "fcu_frd" (aka "base_link_frd"). We will publish it
+	 * with the parent and child frame being set in the rosparams "odometryfcu/odom_*_id_des" (set in px4_config.yaml).
```
mhkabir: Comment seems to be outdated. I propose:


Callback for mavlink ODOMETRY messages sent from the FCU. According to the mavlink specification, all quantities are for the child frame (fcu_frd), expressed in the parent frame (local_origin_ned). To be compliant with ROS REP 147 for the published nav_msgs/Odometry, the data will be appropriately transformed and published. Frames for the publish message should be specified in specified in the rosparams "odometry/fcu/odom_*_id_des" (set in px4_config.yaml).

mhkabir:
I was just re-thinking this whole transformation business, and I think we don't need to involve this ROS TF tree, given we know that all the quantities in the mavlink message always need to be expressed in LOCAL_FRD/NED. This PR, while a massive improvement over what we had, still feels overcomplicated.

I'll raise a PR with my proposed approach (based off this one).

TSC21:
> I was just re-thinking this whole transformation business, and I think we don't need to involve this ROS TF tree, given we know that all the quantities in the mavlink message always need to be expressed in LOCAL_FRD/NED. This PR, while a massive improvement over what we had, still feels overcomplicated.
> 
> I'll raise a PR with my proposed approach (based off this one).

@mhkabir I would keep the TF approach for now. I actually this that it's reasonable to consider using TF instead of Eigen matrices to compute frame transforms throughout the MAVROS plugins. We just need to evaluate the performance of such approach.

mhkabir:
You cannot rotate matrices without Eigen. Right now, TF is not doing much other than getting the fixed transform from UAS.cpp to the odometry plugin, which seems to be pointless.

TSC21:
> You cannot rotate matrices without Eigen. Right now, TF is not doing much other than getting the fixed transform from UAS.cpp to the odometry plugin, which seems to be pointless.

Yes you are right. Though, again, I would suggest we merge this (with the respective changes). And then you can propose a change over @kamilritz changes.

mavros/src/lib/uas_data.cpp
kamilritz: Are you sure? It is in the newest version for me. Also in this PR.

mavros/launch/px4_config.yaml
kamilritz: Agree. I set it to false again.

kamilritz:
> I was just re-thinking this whole transformation business, and I think we don't need to involve this ROS TF tree, given we know that all the quantities in the mavlink message always need to be expressed in LOCAL_FRD/NED. 

@mhkabir I do not fully agree with you. I find the approach that the user of the odom-plugin has to publish only two static tf-transforms in their launch files very elegant. Using the tf-tree together with rviz gives you also an easy way of checking your frames visually. 
How would you like to get the two needed transformations from the user? IMHO moving it to the px4.yaml file is less convenient for the user.

TSC21:
@kamilritz please rebase so we can merge.

TSC21:
@vooon after this gets merged, we need a release to fix the `ODOM: Ex: Could not find a connection between 'local_origin_ned' and 'fcu' because they are not part of the same tree.Tf has two or more unconnected trees.` issue.

PR #1301, opened 2019-08-24
Setpoint trajectory plugin using MultiDOFJointTrajectory message

Jaeyoung-Lim:
As MAVs are dynamically constrained, the capability to track trajectories are important for high performance maneuvers. As the flight controller do not provide trajectory tracking capabilities, current implementation of trajectory planners usually send setpoints to the flight controller in order of time. This trivial task has been implemented on the planner side outside of mavros.

This PR adds a plugin to mavros for setting a trajectory as a setpoint using the [MultiDOFJointTrajectory](http://docs.ros.org/melodic/api/trajectory_msgs/html/msg/MultiDOFJointTrajectory.html) message.

The plugin receives a trajectory, and sends the reference with the [SET_POSITION_TARGET_LOCAL_NED](https://mavlink.io/en/messages/common.html#SET_POSITION_TARGET_LOCAL_NED) message to the flight controller by comparing the time in the trajectory message.

It also publishes the desired trajectory as a `nav_msgs/Path` message so that it can be visualized.
@mhkabir

### DIFF
--- a/mavros/src/plugins/setpoint_trajectory.cpp
+++ b/mavros/src/plugins/setpoint_trajectory.cpp
@@ -87,6 +87,7 @@ private:
 
 	std::string frame_id;
 	MAV_FRAME mav_frame;
+	ftf::StaticTF transform;
 
 	void reset_timer(ros::Duration duration){
 		sp_timer.stop();
@@ -117,6 +118,12 @@ private:
 
 	void local_cb(const trajectory_msgs::MultiDOFJointTrajectory::ConstPtr &req)
 	{
+		if(static_cast<MAV_FRAME>(mav_frame) == MAV_FRAME::BODY_NED || static_cast<MAV_FRAME>(mav_frame) == MAV_FRAME::BODY_OFFSET_NED){
+			transform = ftf::StaticTF::BASELINK_TO_AIRCRAFT;
+		} else {
+			transform = ftf::StaticTF::ENU_TO_NED;
+		}
+
 		trajectory_target_msg = req;
 
 		// Read first duration of the setpoint and set the timer
@@ -137,21 +144,8 @@ private:
 		float yaw, yaw_rate;
 		uint16_t type_mask;		
 		if(!setpoint_target->transforms.empty()){
-			position = [&] () {
-				if (static_cast<MAV_FRAME>(mav_frame) == MAV_FRAME::BODY_NED || static_cast<MAV_FRAME>(mav_frame) == MAV_FRAME::BODY_OFFSET_NED) {
-					return ftf::transform_frame_baselink_aircraft(ftf::to_eigen(setpoint_target->transforms[0].translation));
-				} else {
-					return ftf::transform_frame_enu_ned(ftf::to_eigen(setpoint_target->transforms[0].translation));
-				}
-			} ();
-			attitude = [&] () {
-				if (mav_frame == MAV_FRAME::BODY_NED || mav_frame == MAV_FRAME::BODY_OFFSET_NED) {
-					return ftf::transform_orientation_baselink_aircraft(ftf::to_eigen(setpoint_target->transforms[0].rotation));
-				} else {
-					return ftf::transform_orientation_enu_ned(
-						ftf::transform_orientation_baselink_aircraft(ftf::to_eigen(setpoint_target->transforms[0].rotation)));
-				}
-			} ();
+			position = ftf::detail::transform_static_frame(ftf::to_eigen(setpoint_target->transforms[0].translation), transform);
+			attitude = ftf::detail::transform_orientation(ftf::to_eigen(setpoint_target->transforms[0].rotation), transform);
 
 		} else {
 			type_mask = type_mask | uint16_t(POSITION_TARGET_TYPEMASK::X_IGNORE)
@@ -162,13 +156,7 @@ private:
 		}
 
 		if(!setpoint_target->velocities.empty()){
-			velocity = [&] () {
-				if (static_cast<MAV_FRAME>(mav_frame) == MAV_FRAME::BODY_NED || static_cast<MAV_FRAME>(mav_frame) == MAV_FRAME::BODY_OFFSET_NED) {
-					return ftf::transform_frame_baselink_aircraft(ftf::to_eigen(setpoint_target->velocities[0].linear));
-				} else {
-					return ftf::transform_frame_enu_ned(ftf::to_eigen(setpoint_target->velocities[0].linear));
-				}
-			} ();
+			velocity = ftf::detail::transform_static_frame(ftf::to_eigen(setpoint_target->velocities[0].linear), transform);
 			yaw_rate = setpoint_target->velocities[0].angular.z;
 
 		} else {
@@ -180,13 +168,7 @@ private:
 		}
 		
 		if(!setpoint_target->accelerations.empty()){
-			af = [&] () {
-				if (static_cast<MAV_FRAME>(mav_frame) == MAV_FRAME::BODY_NED || static_cast<MAV_FRAME>(mav_frame) == MAV_FRAME::BODY_OFFSET_NED) {
-					return ftf::transform_frame_baselink_aircraft(ftf::to_eigen(setpoint_target->accelerations[0].linear));
-				} else {
-					return ftf::transform_frame_enu_ned(ftf::to_eigen(setpoint_target->accelerations[0].linear));
-				}
-			} ();
+			af = ftf::detail::transform_static_frame(ftf::to_eigen(setpoint_target->accelerations[0].linear), transform);
 
 		} else {
 			type_mask = type_mask | uint16_t(POSITION_TARGET_TYPEMASK::AX_IGNORE) 
--- a/mavros/src/plugins/setpoint_trajectory.cpp
+++ b/mavros/src/plugins/setpoint_trajectory.cpp
@@ -20,12 +20,15 @@
 
 #include <nav_msgs/Path.h>
 #include <mavros_msgs/PositionTarget.h>
+#include <mavros_msgs/SetMavFrame.h>
 #include <trajectory_msgs/MultiDOFJointTrajectory.h>
 #include <trajectory_msgs/MultiDOFJointTrajectoryPoint.h>
 #include <std_srvs/Trigger.h>
 
 namespace mavros {
 namespace std_plugins {
+using mavlink::common::MAV_FRAME;
+
 /**
  * @brief Setpoint TRAJECTORY plugin
  *
@@ -48,8 +51,17 @@ public:
 		desired_pub = sp_nh.advertise<nav_msgs::Path>("desired", 10);
 
 		trajectory_reset_srv = sp_nh.advertiseService("reset", &SetpointTrajectoryPlugin::reset_cb, this);
-
+		mav_frame_srv = sp_nh.advertiseService("mav_frame", &SetpointTrajectoryPlugin::set_mav_frame_cb, this);
+		
 		sp_timer = sp_nh.createTimer(ros::Duration(0.01), &SetpointTrajectoryPlugin::reference_cb, this, true);
+
+		// mav_frame
+		std::string mav_frame_str;
+		if (!sp_nh.getParam("mav_frame", mav_frame_str)) {
+			mav_frame = MAV_FRAME::LOCAL_NED;
+		} else {
+			mav_frame = utils::mav_frame_from_str(mav_frame_str);
+		}
 	}
 
 	Subscriptions get_subscriptions()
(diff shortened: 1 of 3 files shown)


mavros/src/plugins/setpoint_trajectory.cpp
```diff
@@ -0,0 +1,162 @@
...
+		
+			for(size_t i = 0; i < trajectory_target_msg->points.size(); i++){
+
+				Eigen::Vector3d position, velocity, af;
+				Eigen::Quaterniond attitude;
+				float yaw, yaw_rate;
+				trajectory_msgs::MultiDOFJointTrajectoryPoint pt = trajectory_target_msg->points[i];
+				uint16_t type_mask;
+
+				if(pt.time_from_start.toSec() >= curr_time_from_start.toSec() ) { //TODO: Better logic to handle this case?
```
vooon: You should remember trajectory item that you already sent.

mavros/src/plugins/setpoint_trajectory.cpp
```diff
@@ -0,0 +1,162 @@
...
+					set_position_target_local_ned(
+								trajectory_target_msg->header.stamp.toNSec() / 1000000,
+								1,
+								0,
+								position,
+								velocity,
+								af,
+								yaw, 0);
+					if(i == trajectory_target_msg->points.size()-1)  trajectory_target_msg.reset(); //End of trajectory
+					break;
```
vooon: I think here better to use iterators and compare against `v::cend()`.

Jaeyoung-Lim: I decided it is better to return the function if the trajectory is published and delete the message when it makes it to the end of the iterator

mavros/src/plugins/setpoint_trajectory.cpp
Jaeyoung-Lim: @vooon Could you elaborate on this comment? I tried not remembering the item, since the trajectory_msg might have changed(most likely) between the two times

Jaeyoung-Lim:
@Voon Thanks for the inputs!

For the oneshot timer, are you suggesting that the timer is triggered when the next item in the trajectory is valid?

vooon:
@Jaeyoung-Lim yes, i suggest that you need to send new SP on required moment.
Or you have an idea to implement repeating SP's to keep required message rate by this timer?

TSC21:
@Jaeyoung-Lim are you planning to finish this soon? We should make a release of MAVROS soon and would be cool to bring this in.

Jaeyoung-Lim:
@TSC21 I just need a bit more testing. I will try to finish this by one or two days

TSC21:
> @TSC21 I just need a bit more testing. I will try to finish this by one or two days. Would this be enough for the release?

Thanks! We should be able to do the release in the next days. The larger the feature set we bring in the better.

Jaeyoung-Lim:
@TSC21 Cool. I noticed that we also need a service to cancel the trajectory, if we decide that the trajectory is not good during following the trajectory. This I will do together with more testing.

TSC21:
> I noticed that we also need a service to cancel the trajectory, if we decide that the trajectory is not good during following the trajectory.

That's reasonable yes!

mavros/src/plugins/setpoint_trajectory.cpp
```diff
@@ -0,0 +1,165 @@
...
+		float yaw, yaw_rate;
+		uint16_t type_mask;		
+		if(!setpoint_target->transforms.empty()){
+			position = ftf::to_eigen(setpoint_target->transforms[0].translation);
+			tf::quaternionMsgToEigen(setpoint_target->transforms[0].rotation, attitude);
+		} else {
+			type_mask = type_mask | uint16_t(POSITION_TARGET_TYPEMASK::X_IGNORE) | uint16_t(POSITION_TARGET_TYPEMASK::Y_IGNORE) | uint16_t(POSITION_TARGET_TYPEMASK::Z_IGNORE);
+		}
+
+		if(!setpoint_target->velocities.empty()) velocity = ftf::to_eigen(setpoint_target->velocities[0].linear);
```
TSC21: can you break the lines on the if else conditions? it becomes much more readable that way

mavros/src/plugins/setpoint_trajectory.cpp
```diff
@@ -0,0 +1,165 @@
...
+		else type_mask = type_mask | uint16_t(POSITION_TARGET_TYPEMASK::VX_IGNORE) | uint16_t(POSITION_TARGET_TYPEMASK::VY_IGNORE) | uint16_t(POSITION_TARGET_TYPEMASK::VZ_IGNORE);
+		
+		if(!setpoint_target->accelerations.empty()) af = ftf::to_eigen(setpoint_target->accelerations[0].linear);
+		else type_mask = type_mask | uint16_t(POSITION_TARGET_TYPEMASK::AX_IGNORE) | uint16_t(POSITION_TARGET_TYPEMASK::AY_IGNORE) | uint16_t(POSITION_TARGET_TYPEMASK::AZ_IGNORE);
+
+		// Transform frame ENU->NED
+		position = ftf::transform_frame_enu_ned(position);
```
TSC21: Just bring these directly bellow as `set_position_target_local_ned`. No need for auxiliar variables.

mavros/src/plugins/setpoint_trajectory.cpp
```diff
@@ -0,0 +1,165 @@
...
+public:
+	SetpointTrajectoryPlugin() : PluginBase(),
+		sp_nh("~setpoint_trajectory")
+	{ }
+
+	void initialize(UAS &uas_)
+	{
+		PluginBase::initialize(uas_);
+
+		sp_nh.param<std::string>("frame_id", frame_id, "map");
```
TSC21: Do we want this to be `map` or `odom` by default?

Jaeyoung-Lim: This I don't have a good insight. What would be a better way to handle this?

mavros/src/plugins/setpoint_trajectory.cpp
```diff
@@ -0,0 +1,165 @@
...
+		uint16_t type_mask;		
+		if(!setpoint_target->transforms.empty()){
+			position = ftf::to_eigen(setpoint_target->transforms[0].translation);
+			tf::quaternionMsgToEigen(setpoint_target->transforms[0].rotation, attitude);
+		} else {
+			type_mask = type_mask | uint16_t(POSITION_TARGET_TYPEMASK::X_IGNORE) | uint16_t(POSITION_TARGET_TYPEMASK::Y_IGNORE) | uint16_t(POSITION_TARGET_TYPEMASK::Z_IGNORE);
+		}
+
+		if(!setpoint_target->velocities.empty()) velocity = ftf::to_eigen(setpoint_target->velocities[0].linear);
+		else type_mask = type_mask | uint16_t(POSITION_TARGET_TYPEMASK::VX_IGNORE) | uint16_t(POSITION_TARGET_TYPEMASK::VY_IGNORE) | uint16_t(POSITION_TARGET_TYPEMASK::VZ_IGNORE);
```
TSC21: Also please do something similar to https://github.com/mavlink/mavros/blob/master/mavros/src/plugins/setpoint_position.cpp#L193-L198

mavros/src/plugins/setpoint_trajectory.cpp
TSC21: Depends. Are we navigating within an arbitrary navigation frame or a well defined global frame (resulting from GPS coordinates for example)?

mavros/src/plugins/setpoint_trajectory.cpp
```diff
@@ -0,0 +1,180 @@
...
+	void reference_cb(const ros::TimerEvent &event)
+	{
+		using mavlink::common::POSITION_TARGET_TYPEMASK;
+
+		if(!trajectory_target_msg)
+			return;
+
+		Eigen::Vector3d position, velocity, af;
+		Eigen::Quaterniond attitude;
+		float yaw, yaw_rate;
```
TSC21: What about the yaw rate?

Jaeyoung-Lim: Good point. Would it make sense to always set it to zero?

TSC21: Why? Can't we handle it as well? You have twist data that you can use

Jaeyoung-Lim: That would always enforce the angular rate velocities commands to be valid in the type mask. I am not sure if that is something we want

TSC21: I am not sure I get you: why can't we set the angular rate when the velocity setmask is set for example?

mavros/src/plugins/setpoint_trajectory.cpp
Jaeyoung-Lim: I guess it is safe to assume we are navigating in the local frame for now. Would `map` be an appropriate frame for that?

mavros/src/plugins/setpoint_trajectory.cpp
Jaeyoung-Lim: I just wasn't sure if we want to use the velocity information in the message. I guess it makes sense to use it actually

mavros/src/plugins/setpoint_trajectory.cpp
TSC21: Maybe this helps you take thag decision: https://www.ros.org/reps/rep-0105.html

TSC21:
About the reset: what if the setpoints were already sent to the flight controller?

mavros/src/plugins/setpoint_trajectory.cpp
TSC21: Also, if we are sending setpoints exclusively in the local frame, then that needs to be clarified as a comment. Though, I don't see a reason why we shouldn't send setpoints in the body frame as well, similar to what we do on the other setpoint plugins.

mavros/src/plugins/setpoint_trajectory.cpp
Jaeyoung-Lim: Done!

Jaeyoung-Lim:
@TSC21 It will try and follow the last setpoint that has been sent, but soon will go into hold mode after the timeout as there are no setpoints being sent to thfe flight controller.

mavros/src/plugins/setpoint_trajectory.cpp
Jaeyoung-Lim: How can we handle this with the multiDOFJointTrajectory message? Do you have any good pointers? For me it seems like you just need to make sure you set the correct frame_id

TSC21: The setpoint message has a frame field. You hard setting it to the local frame.

TSC21: https://mavlink.io/en/messages/common.html#SET_POSITION_TARGET_LOCAL_NED

Jaeyoung-Lim: Would setting the default to `MAV_FRAME_LOCAL_NED` make sense?

TSC21: No. Check what's being done here: https://github.com/mavlink/mavros/blob/master/mavros/src/plugins/setpoint_position.cpp#L139

TSC21: @Jaeyoung-Lim any news here?

Jaeyoung-Lim: Sorry for the delay. I was working on this and was experiencing some problems with the transforms. I will try and get this done this week.

mavros/src/plugins/setpoint_trajectory.cpp
```diff
@@ -123,10 +137,21 @@ class SetpointTrajectoryPlugin : public plugin::PluginBase,
...
-			tf::quaternionMsgToEigen(setpoint_target->transforms[0].rotation, attitude);
-			yaw = ftf::quaternion_get_yaw(ftf::transform_orientation_enu_ned(
-					ftf::transform_orientation_baselink_aircraft(attitude)));
+			position = [&] () {
+				if (static_cast<MAV_FRAME>(mav_frame) == MAV_FRAME::BODY_NED || static_cast<MAV_FRAME>(mav_frame) == MAV_FRAME::BODY_OFFSET_NED) {
+					return ftf::transform_frame_baselink_aircraft(ftf::to_eigen(setpoint_target->transforms[0].translation));
+				} else {
+					return ftf::transform_frame_enu_ned(ftf::to_eigen(setpoint_target->transforms[0].translation));
+				}
+			} ();
```
vooon: https://github.com/mavlink/mavros/blob/master/mavros/include/mavros/frame_tf.h#L180

I think better just select StaticTF once.

Jaeyoung-Lim: Once you mean when we receive the trajectory message?

vooon: I'm a bit aware where it's better to place, but in my opinion, you need only once select transform type (baselink-aircraft / enu-ned) and then just use `ftf::transform_frame()` / `ftf::transform_orientation()`.
That also will look much better than current lambdas.

Jaeyoung-Lim:
@vooon I am getting segfaults when using `ftf::detail::transform_orientation()` any ideas?

Jaeyoung-Lim:
@vooon Ah, there was a race condition. I added a mutex

Jaeyoung-Lim:
@Voon @TSC21 Would it be possible to get this in for v0.34 ?

TSC21:
> @vooon @TSC21 Would it be possible to get this in for v0.34 ?

Yes.

Jaeyoung-Lim:
@TSC21 Thanks!

Jaeyoung-Lim:
@vooon Is there no release happening today?

vooon:
Today probably only mavlink. There is no big changes in mavros. But i may do 0.33.1.

PR #1308, opened 2019-09-06
Mavros integration with moveit

saifullah3396:
This pull request makes the following contributions:
1. Added a simple launch file for visaulizing px4-iris with rviz. I saw there was a bit of confusion on how to launch it in rviz since rviz used urdf while px4 loads an sdf file in gazebo. So it just gives a simple solution to new users.

2. A new package named mavros-moveit was added that contains a moveit config, a moveit controller-manager, and an action server that is called by the controller manager. The mavros_moveit_config package defines the attachment of the 'map' frame with 'base_link' frame through a virtual joint and defines the controller configuration. The trajectories generated by moveit are sent to the vehicle through the joint control action server defined in the mavros_moveit_actions package. The action server receives the trajectory on execution callback, arms the robot, and sends it to OFFBOARD mode. After that, it sends the trajectory commands to the robot through the setpoint_position/local topic. Once the robot finishes following the trajectory commands, the server sends the robot to AUTO.LOITER mode and returns.


saifullah3396:
@TSC21 thank you for your insight. Now that you have mentioned, it makes sense to have mavros independent of additional packages. I will do exactly as you recommend so that this update can be matured into a standard package for usage with mavros/px4. I will also add a velocity controller to it as an option since it definitely makes more sense for tracking the desired trajectories.

TSC21:
@saifullah3396 please keep us updated on that. Also do bring that to PX4 Slack so everyone interested can participate as well. Thanks!

PR #1314, opened 2019-09-18
added manual flag to mavros/state

dayjaby:
We need to know whether the drone is in a manual state or not. In the HEARTBEAT mavlink message there is a flag for that. The alternative would be to see if /mavros/state/mode is any of the manual modes, but this PR would be more compliant with mavlink.

### DIFF
--- a/mavros/src/plugins/sys_status.cpp
+++ b/mavros/src/plugins/sys_status.cpp
@@ -728,6 +728,7 @@ private:
 		state_msg->connected = true;
 		state_msg->armed = !!(hb.base_mode & enum_value(MAV_MODE_FLAG::SAFETY_ARMED));
 		state_msg->guided = !!(hb.base_mode & enum_value(MAV_MODE_FLAG::GUIDED_ENABLED));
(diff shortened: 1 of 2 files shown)


mavros/src/plugins/sys_status.cpp
```diff
@@ -728,6 +728,7 @@ class SystemStatusPlugin : public plugin::PluginBase
 		state_msg->connected = true;
 		state_msg->armed = !!(hb.base_mode & enum_value(MAV_MODE_FLAG::SAFETY_ARMED));
 		state_msg->guided = !!(hb.base_mode & enum_value(MAV_MODE_FLAG::GUIDED_ENABLED));
+		state_msg->manual = !!(hb.base_mode & enum_value(MAV_MODE_FLAG::MANUAL_INPUT_ENABLED));
```
vooon: Probably better to keep "manual input" name.

dayjaby: Of course! Done in the latest commit.

PR #1315, opened 2019-09-20
plugin Command: add parameter `command_ack_timeout`

szobov:
Sometimes commands take more time than default 5 seconds. Due to a low bandwidth of UART and a high rate of some mavlink streams. To eliminate this problem it's better to provide the parameter to configure the command's ack timeout.

I've not found any place in the code. where I can fix documentation for this changes.

### DIFF
--- a/mavros/src/plugins/command.cpp
+++ b/mavros/src/plugins/command.cpp
@@ -64,7 +64,7 @@ public:
 		PluginBase::initialize(uas_);
 
 		cmd_nh.param("use_comp_id_system_control", use_comp_id_system_control, false);
-		int command_ack_timeout_ms;
+		double command_ack_timeout_ms;
 		cmd_nh.param("command_ack_timeout", command_ack_timeout_ms, ACK_TIMEOUT_MS_DEFAULT);
 		command_ack_timeout_dt = ros::Duration(command_ack_timeout_ms / 1000);
 
@@ -103,7 +103,7 @@ private:
 	bool use_comp_id_system_control;
 
 	L_CommandTransaction ack_waiting_list;
-	static constexpr int ACK_TIMEOUT_MS_DEFAULT = 5000;
+	static constexpr double ACK_TIMEOUT_MS_DEFAULT = 5000.0;
 	ros::Duration command_ack_timeout_dt;
 
 	/* -*- message handlers -*- */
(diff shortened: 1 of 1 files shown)


szobov:
@vooon ping

mavros/src/plugins/command.cpp
```diff
@@ -56,15 +56,17 @@ class CommandPlugin : public plugin::PluginBase {
...
 	{ }
 
 	void initialize(UAS &uas_)
 	{
 		PluginBase::initialize(uas_);
 
 		cmd_nh.param("use_comp_id_system_control", use_comp_id_system_control, false);
+		int command_ack_timeout_ms;
+		cmd_nh.param("command_ack_timeout", command_ack_timeout_ms, ACK_TIMEOUT_MS_DEFAULT);
+		command_ack_timeout_dt = ros::Duration(command_ack_timeout_ms / 1000);
```
vooon: Use `double` with seconds, e.g. https://github.com/mavlink/mavros/blob/master/mavros/src/plugins/sys_status.cpp#L443

szobov: fixed

szobov:
@vooon could it be merged?

mavros/src/plugins/command.cpp
```diff
@@ -64,9 +64,12 @@ class CommandPlugin : public plugin::PluginBase {
 		PluginBase::initialize(uas_);
 
 		cmd_nh.param("use_comp_id_system_control", use_comp_id_system_control, false);
-		double command_ack_timeout_ms;
-		cmd_nh.param("command_ack_timeout", command_ack_timeout_ms, ACK_TIMEOUT_MS_DEFAULT);
-		command_ack_timeout_dt = ros::Duration(command_ack_timeout_ms / 1000);
+
+		double command_ack_timeout;
```
vooon:
```
double foo;

nh.param("bar", &bar, 1);
nh.param("foo", &foo, 2);

baz = ros::Ducration(foo);
```

szobov: yes, thanks. Got and fixed it.

Btw, is here any style guides? It's a bit frustrating now, when I should look through another source files in order to understand style.

vooon: No, we don't have one. I tried to setup uncrustify to follow my style, and it's fine most time, but it can't reorganise code blocks like above.

szobov: The code style guide will work fine in this kind of situations.
I mean, it's really difficult to format this code automatically, but in the code style you can just describe it with the words and examples, so none can argue it.
Also, you can fill the code style guide step by step and you shouldn't reformat all existing code immediately. Just format the new code according to the guide and reformat the old code, if you are touching it. I can make a PR with a draft of code style guide, if you want.

Btw, thanks for merge.

PR #1321, opened 2019-09-26
Make local_position plugin work regardless of imu plugin

mortenfyhn:
If I run the local_position plugin without running the imu plugin, the orientation value of the local position is just the identity quaternion.  This seems to be because the local_position plugin calls `UAS::get_attitude_orientation_enu` to read the orientation, but it never sets the orientation by calling `UAS::update_attitude_imu_enu`. The imu plugin does call `UAS::update_attitude_imu_enu`, so if both are running at the same time, then everything works as expected.

My guess is that local_position is intended to be possible to run alone.

The easiest, but maybe not best solution would be to just mention in the documentation that local_position depends on imu also running. A better solution in my opinion would be to make local_position independent of imu. The draft code here works but shouldn't be merged the way it is.

### DIFF
--- a/mavros/src/plugins/local_position.cpp
+++ b/mavros/src/plugins/local_position.cpp
@@ -68,6 +68,7 @@ public:
 
 	Subscriptions get_subscriptions() {
 		return {
+			       make_handler(&LocalPositionPlugin::handle_attitude),
 			       make_handler(&LocalPositionPlugin::handle_local_position_ned),
 			       make_handler(&LocalPositionPlugin::handle_local_position_ned_cov)
 		};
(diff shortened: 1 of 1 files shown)


mortenfyhn:
That seems fair, in that case I would at least suggest to mention the dependency here: http://wiki.ros.org/mavros#mavros.2BAC8-Plugins.local_position

For someone not intimate with mavros, it's easy to assume that the imu plugin is only for raw IMU data, and that local_position which gives position/orientation is a separate thing. At least to me the dependency is not obvious.

TSC21:
> That seems fair, in that case I would at least suggest to mention the dependency here: http://wiki.ros.org/mavros#mavros.2BAC8-Plugins.local_position
> 
> For someone not intimate with mavros, it's easy to assume that the imu plugin is only for raw IMU data, and that local_position which gives position/orientation is a separate thing. At least to me the dependency is not obvious.

I agree and I know that the Wiki is far from being updated. But there's a lot of things going on and the time is short :) Closing this for now.

mortenfyhn:
I gotchu. I'll try to update the wiki myself.

Issue #1350, opened 2019-11-17
Build failure with boost >= 1.70


acxz:
### Issue details

libmavconn does not build with later boost versions. The following build error occurs:
```
Scanning dependencies of target mavconn
[ 16%] Building CXX object CMakeFiles/mavconn.dir/src/tcp.cpp.o
[ 66%] Building CXX object CMakeFiles/mavconn.dir/src/mavlink_helpers.cpp.o
[ 66%] Building CXX object CMakeFiles/mavconn.dir/src/serial.cpp.o
[ 66%] Building CXX object CMakeFiles/mavconn.dir/src/interface.cpp.o
[ 83%] Building CXX object CMakeFiles/mavconn.dir/src/udp.cpp.o
/tmp/yaourt-tmp-korny/aur-ros-melodic-libmavconn/src/mavros-release-release-melodic-libmavconn-0.26.3-0/src/tcp.cpp: In Elementfunktion »void mavconn::MAVConnTCPClient::client_connected(size_t)«:
/tmp/yaourt-tmp-korny/aur-ros-melodic-libmavconn/src/mavros-release-release-melodic-libmavconn-0.26.3-0/src/tcp.cpp:123:9: Fehler: »boost::asio::ip::tcp::socket« {aka »class boost::asio::basic_stream_socket<boost::asio::ip::tcp>«} hat kein Element namens »get_io_service«
  123 |  socket.get_io_service().post(std::bind(&MAVConnTCPClient::do_recv, shared_from_this()));
      |         ^~~~~~~~~~~~~~
/tmp/yaourt-tmp-korny/aur-ros-melodic-libmavconn/src/mavros-release-release-melodic-libmavconn-0.26.3-0/src/tcp.cpp: In Elementfunktion »virtual void mavconn::MAVConnTCPClient::send_bytes(const uint8_t*, size_t)«:
/tmp/yaourt-tmp-korny/aur-ros-melodic-libmavconn/src/mavros-release-release-melodic-libmavconn-0.26.3-0/src/tcp.cpp:169:9: Fehler: »boost::asio::ip::tcp::socket« {aka »class boost::asio::basic_stream_socket<boost::asio::ip::tcp>«} hat kein Element namens »get_io_service«
  169 |  socket.get_io_service().post(std::bind(&MAVConnTCPClient::do_send, shared_from_this(), true));
      |         ^~~~~~~~~~~~~~
[...]

PR #1351, opened 2019-11-17
add macro for get_io_service() to work with boost>1.70

acxz:
This PR addresses #1350 

To make sure libmavconn compiles with both boost versions >= 1.70 and below 1.70 a macro is created to ensure the correct API call is made for get_io_service().


vooon:
LGTM, Thanks!

acxz:
Thx for the quick response, really means a lot!

Issue #1082, opened 2018-09-07
mapping LaserScan to obstacle_distance mavlink message


rmackay9:
I'm attempting to use the [obstacle_distance](https://github.com/mavlink/mavros/blob/master/mavros_extras/src/plugins/obstacle_distance.cpp) plugin to send distances from a 360degree RPLidar range finder to ArduPilot.  The [OBSTACLE_DISTANCE](http://mavlink.org/messages/common#OBSTACLE_DISTANCE) mavlink messages are being sent to AP but only include the first 72 degrees of data from the range finder.

The messages from RPLidar are appear in /scan as [LaserScan messages](http://docs.ros.org/melodic/api/sensor_msgs/html/msg/LaserScan.html) and below is a screen shot of the output of, "rostopic echo /scan".
![scan-to-obstacle-distance](https://user-images.githubusercontent.com/1498098/45201751-72d02800-b2b1-11e8-899f-cb78b33296d3.png)

I've enabled the obstacle_distance plugin by adding "obstacle_distance" to the apm_pluginlist.yaml file and then in the node.launch file, I've added a remap like this, "\<remap from="/mavros/obstacle/send" to="/scan" />".

The problem is though that the LaserScan message's have 1degree increments and the OBSTACLE_DISTANCE message's distance array only has 72 slots.  Ideally I would like to combine the LaserScan data down into 45degree buckets and then send to AP via the OBSTACLE_DISTANCE message.  I know 45 degree buckets is coarse but it should be sufficient for the obstacle avoidance we want to use this data for.

vooon:
You need to add filter node that will reduce number of points.
I think that node should be already written, just need to google.
And that is not specific to mavros, it is more generic ros comms.

rmackay9:
OK, thanks for that.  I'll close this issue then 'cuz sounds like we've got a path to a solution and it hopefully/probably won't require a change to mavros.  Txs again.

rmackay9:
@vooon, we're making progress on getting the 360 lidar data into AP ([ArduPilot driver enhancement is here](https://github.com/ArduPilot/ardupilot/pull/9418)).

We've also hacked together a change to RPLidar's node.cpp that resizes the output (look for "Resize of data for mavros" in [this link here](https://www.dropbox.com/s/xj82jyugv9pwfw3/node.cpp?dl=0)).

[...]

PR #1370, opened 2019-12-31
mavros_extras/obstacle_distance: Change namespace, add angle_offset

sfalexrog:
While trying to set up collision prevention, I've found several issues with the `obstacle_distance` plugin:
* The parameter namespace for `obstacle_distance` is `obstacle`, which is not very discoverable;
* `angle_offset` is always zero in the MAVLink message.

This P/R changes the namespace to match the plugin name, adds an example parameter in `px4_config.yaml` (I'm not familiar with Ardupilot, not sure the plugin works with it), and uses `angle_min` from ROS `sensor_msgs::LaserScan` as `angle_offset`. This worked for me with a D435+depthimage_to_laserscan (and a simple node to "flip" the laserscan message).

### DIFF
--- a/mavros/launch/px4_config.yaml
+++ b/mavros/launch/px4_config.yaml
@@ -259,6 +259,6 @@ wheel_odometry:
 
 # obstacle_distance
 obstacle_distance:
-  mav_frame: "BODY_FRD"       # PX4 1.10 supports GLOBAL, LOCAL_NED, BODY_FRD
+  mav_frame: "GLOBAL"       # PX4 1.10 supports GLOBAL, LOCAL_NED, BODY_FRD
 
 # vim:set ts=2 sw=2 et:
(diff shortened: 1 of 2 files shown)


sfalexrog:
According to #1082, I should probably try and notify the avoidance team. @mrivi, @baumanta, @jkflying, @nicovanduijn, I'd really appreciate any input on your part!

mavros/launch/px4_config.yaml
```diff
@@ -257,4 +257,8 @@ wheel_odometry:
     frame_id: "odom"
     child_frame_id: "base_link"
 
+# obstacle_distance
+obstacle_distance:
+  mav_frame: "BODY_FRD"       # PX4 1.10 supports GLOBAL, LOCAL_NED, BODY_FRD
```
jkflying: Just FYI, this changes the default, which was global. I think it would be better to keep the default the same, and let people override if they need it.

sfalexrog: Makes sense, thanks!

vooon:
I added angle_offset, but change of namespace would break backward compatibility, which i do not want to see past 1.0.0...

PR #1373, opened 2020-01-08
bugfix: Save linear acceleration on RAW_IMU only on APM

zhouchengming99:
Previous code save linear acceleration on RAW_IMU for any type
autopilot, and set received_linear_accel to true, will cause to
publish wrong data on ~imu/data topic, like linear_acceleration
is zero but linear_acceleration_covariance[0] is NOT -1.

### DIFF
--- a/mavros/src/plugins/imu.cpp
+++ b/mavros/src/plugins/imu.cpp
@@ -238,10 +238,18 @@ private:
 		tf::vectorEigenToMsg(gyro_flu, imu_msg->angular_velocity);
 		tf::vectorEigenToMsg(accel_flu, imu_msg->linear_acceleration);
 
-		// Save readings
-		linear_accel_vec_flu = accel_flu;
-		linear_accel_vec_frd = accel_frd;
-		received_linear_accel = true;
+		if (!m_uas->is_ardupilotmega()) {
+			ROS_WARN_THROTTLE_NAMED(60, "imu", "IMU: linear acceleration on RAW_IMU known on APM only.");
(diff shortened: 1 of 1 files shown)


zhouchengming99:
Emm, Looks like the build robot failed because a mavconn testcase:
18:33:30 Error:   mavconn: tcp15: receive: End of file
18:33:30          at line 231 in /tmp/ws/src/mavros/libmavconn/src/tcp.cpp
Is this a known issue?

vooon:
What FCU aside APM sends IMU_RAW?

mavros/src/plugins/imu.cpp
```diff
@@ -238,10 +238,18 @@ class IMUPlugin : public plugin::PluginBase {
 		tf::vectorEigenToMsg(gyro_flu, imu_msg->angular_velocity);
 		tf::vectorEigenToMsg(accel_flu, imu_msg->linear_acceleration);
 
-		// Save readings
-		linear_accel_vec_flu = accel_flu;
-		linear_accel_vec_frd = accel_frd;
-		received_linear_accel = true;
+		if (!m_uas->is_ardupilotmega()) {
```
vooon: `~imu/data` should be consistent across different FCUs, so it's safer not to publish anything, than unscaled values.

zhouchengming99: Yes, ~imu/data_raw will publish unscaled values, and ~imu/data will keep default value {0};
m_uas->is_ardupilotmega() will become True actually latter, but after the heartbeat publish the correct pilot type.

PR #1380, opened 2020-01-29
Dispatch GCS IP address

mortenfyhn:
Hi all! Are you still open for PRs on the ROS 1 version of mavros?

I wanted to get the IP of the ground control station that mavros/mavconn connects to, so I made this change. I'm not sure if this is something you want in the upstream mavros, but I'm making this PR to find out.

It runs a 1 Hz timer in the main mavros loop that checks the ip stored in the MAVConnUDP instance (through a new public const method), and publishes it on a latched topic `gcp_ip`. It only does this once.

This implementation works, but it feels like it could be integrated in a nicer way. All suggestions appreciated.

### DIFF
--- a/libmavconn/include/mavconn/udp.h
+++ b/libmavconn/include/mavconn/udp.h
@@ -60,6 +60,8 @@ public:
 		return socket.is_open();
 	}
 
+	std::string get_remote_endpoint() const;
+
 private:
 	boost::asio::io_service io_service;
 	std::unique_ptr<boost::asio::io_service::work> io_work;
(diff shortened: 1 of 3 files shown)


mortenfyhn:
Any feedback on this?

vooon:
Hello, Sorry, i must be missed that PR.
Solution a bit hacky, but i think it's ok. At least i have no suggestions now.

mortenfyhn:
Thanks! I'm open to editing it in a new pr if you or anyone else have ideas for a more elegant implementation.

PR #1384, opened 2020-02-17
Visualization of the direction of the device

Alamoris:
In the current implementation, it is completely unclear where the "nose" of the device is directed.
In order not to overload the scene with elements, I propose adding a parameter that will tint the front and rear propellers in different colors, red and white, respectively. This logic used on DJI propellers on the f450 and f500 frame.

### DIFF
--- a/mavros_extras/src/visualization.cpp
+++ b/mavros_extras/src/visualization.cpp
@@ -225,7 +225,7 @@ static void create_vehicle_markers( int num_rotors, float arm_len, float body_wi
 				rotor.color.r = 1.0;
 				rotor.color.g = 0;
 				rotor.color.b = 0;
-				rotor.color.a = 1.0;
+				rotor.color.a = 0.8;
 			}
 		}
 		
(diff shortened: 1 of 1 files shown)


vooon:
Could you please add screenshot from rviz to see the difference?

Alamoris:
Here are a few screenshots with different numbers of device rotors. In some screenshots, added a direction arrow, for greater visibility
Screenshot from 2020-02-18 11-09-31
Screenshot from 2020-02-18 11-09-24
Screenshot from 2020-02-18 11-07-40
Screenshot from 2020-02-18 10-58-34
Screenshot from 2020-02-18 10-51-58
Screenshot from 2020-02-18 10-51-28

okalachev:
@Alamoris, @vooon, I would prefer the rear propellers to be not so bright. Can we make the color adjustable?

`front_props` and `rear_props`, for example.

vooon:
@okalachev , well, i think better to merge as is, and then you could add parameters.

PR #1386, opened 2020-02-19
Include trajectory_msgs in CMakeLists.txt

agoeckner:
This allows build to complete successfully. The `trajectory_msgs` package is a requirement and is specified in the package.xml, but not in CMakeLists.txt.


vooon:
Thanks!

PR #1402, opened 2020-03-20
distance_sensor: Fill horizontal_fov, vertical_fov, quaternion

sfalexrog:
Currently, the `distance_sensor` plugin does not fill the `horizontal_fov`, `vertical_fov` and `quaternion` fields of the [`DISTANCE_SENSOR`](https://mavlink.io/en/messages/common.html#DISTANCE_SENSOR) message. This results in uninitialized values being sent to the flight controller.

This P/R fills in the `horizontal_fov` and `vertical_fov` fields from the `field_of_view` field of the `sensor_msgs/Range` message. It also sets the quaternion to zero (it is not used unless the orientation is set to `MAV_SENSOR_ROTATION_CUSTOM`, and the message description says: "Set it to 0 if invalid.").

### DIFF
--- a/mavros_extras/src/plugins/distance_sensor.cpp
+++ b/mavros_extras/src/plugins/distance_sensor.cpp
@@ -163,7 +163,8 @@ private:
 				uint32_t max_distance,
 				uint32_t current_distance,
 				uint8_t type, uint8_t id,
-				uint8_t orientation, uint8_t covariance)
+				uint8_t orientation, uint8_t covariance,
+				float fov)
 	{
 		mavlink::common::msg::DISTANCE_SENSOR ds;
 
@@ -187,6 +188,9 @@ private:
 		ds.orientation = orientation;
 		ds.covariance = covariance;
 		// [[[end]]] (checksum: 6f6f9449d926ab618a3293e2091c7035)
+		ds.vertical_fov = fov;
+		ds.horizontal_fov = fov;
+		ds.quaternion = {0};
 
 		UAS_FCU(m_uas)->send_message_ignore_drop(ds);
 	}
@@ -300,7 +304,8 @@ void DistanceSensorItem::range_cb(const sensor_msgs::Range::ConstPtr &msg)
 				type,
 				sensor_id,
 				orientation,
-				covariance_);
+				covariance_,
+				msg->field_of_view);
 }
 
 DistanceSensorItem::Ptr DistanceSensorItem::create_item(DistanceSensorPlugin *owner, std::string topic_name)
--- a/mavros_extras/src/plugins/distance_sensor.cpp
+++ b/mavros_extras/src/plugins/distance_sensor.cpp
@@ -42,7 +42,10 @@ public:
 		orientation(-1),
 		covariance(0),
 		owner(nullptr),
-		data_index(0)
+		data_index(0),
+		horizontal_fov_ratio(1.0),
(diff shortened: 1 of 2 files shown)


mavros_extras/src/plugins/distance_sensor.cpp
```diff
@@ -187,6 +188,9 @@ class DistanceSensorPlugin : public plugin::PluginBase {
 		ds.orientation = orientation;
 		ds.covariance = covariance;
 		// [[[end]]] (checksum: 6f6f9449d926ab618a3293e2091c7035)
+		ds.vertical_fov = fov;
+		ds.horizontal_fov = fov;
+		ds.quaternion = {0};
```
TSC21: > It also sets the quaternion to zero (it is not used unless the orientation is set to MAV_SENSOR_ROTATION_CUSTOM, and the message description says: "Set it to 0 if invalid.").

OK so what if `MAV_SENSOR_ROTATION_CUSTOM` is set? It might be reasonable to add an parameter to the range sensor that allows to set it's quaternion instead. Like we have https://github.com/mavlink/mavros/blob/master/mavros/launch/px4_config.yaml#L145.

mavros_extras/src/plugins/distance_sensor.cpp
```diff
@@ -187,6 +188,9 @@ class DistanceSensorPlugin : public plugin::PluginBase {
 		ds.orientation = orientation;
 		ds.covariance = covariance;
 		// [[[end]]] (checksum: 6f6f9449d926ab618a3293e2091c7035)
+		ds.vertical_fov = fov;
```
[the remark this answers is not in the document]
sfalexrog: Added the fields to the `cog` block.

mavros_extras/src/plugins/distance_sensor.cpp
```diff
@@ -187,6 +188,9 @@ class DistanceSensorPlugin : public plugin::PluginBase {
...
+		ds.horizontal_fov = fov;
```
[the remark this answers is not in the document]
sfalexrog: Added `vertical_fov_ratio` and `horizontal_fov_ratio` as parameters. They are set to 1.0 by default.

mavros_extras/src/plugins/distance_sensor.cpp
sfalexrog: I've added `sensor_orientation` to the set of parameters, which allows the user to specify orientation using roll/pitch/yaw angles. But the `sensor_msgs/Range` message already contains the `frame_id` of the sensor, shouldn't we try and use that instead? The only thing that I'm confused by is the "vehicle body frame": is it FRD? FLU? Is it different from autopilot to autopilot?

mavros_extras/src/plugins/distance_sensor.cpp
```diff
@@ -54,6 +57,9 @@ class DistanceSensorItem {
 	int orientation;	//!< check orientation of sensor if != -1
 	int covariance;		//!< in centimeters, current specification
 	std::string frame_id;	//!< frame id for send
+	double horizontal_fov_ratio;		//!< horizontal fov ratio for ROS messages
```
amilcarlucas, line 60: You should align the comments (avoid tabs)

vooon:
Changes looks valuable, but PR completely stall. I am going to merge and address suggestions after that.

PR #1415, opened 2020-04-27
Mission item int

### LINKED ISSUE #1375: Waypoint push low resolution
This is only bug and feature tracker, please use it
to report bugs or request features.

----

### Issue details

Setting waypoints looses resolution with lat/long accurate only to approx 5 decimal places. This results in waypoints that are up to around +/- 1m out.

Mavros: 1.0.0
ROS: Melodic
Ubuntu: ?18.04?

[ X ] ArduPilot
[  ] PX4

bromeara:
This pull request adds a class that inherits from MISSION_ITEM_INT and allows for the usage of it with the parameter use_mission_item_int if it's not given it will use the flight controller capabilities. I used cog where it felt practical, I used boost variant to improve code reuse slightly. I am curious if I should have used cog for handle_mission_item and handle_mission_item_int since they are very similar.
I switched the send_waypoint and waypoint vectors to storing the ROS messages instead of the mavlink messages so that they can be converted to float or int at will without worrying about precision loss. I added 2 new states specifically for the _INT messages in order to help with the fallback to the non _INT message in the case that use_mission_item_int is set to true but it is not supported by the flight controller(not really an issue for Ardupilot but might help on other flight stacks). 
I was able to test the push and pull functionality for both INT and float messages using push and pull. I was unable to test it on PX4, when I tried pushing waypoints it gave me the messages:
WP: upload failed: Command is not supported.
FCU: IGN MISSION_ITEM: Busy 

I was running the jmavsim sitl and have not used px4 before so I was unsure if it was a problem with the plugin or something about the simulator I was not getting if. So help to verify on PX4 would be appreciated.

This should resolve #1375


mavros/src/plugins/waypoint.cpp
```diff
@@ -231,11 +386,16 @@ class WaypointPlugin : public plugin::PluginBase {
...
+	bool use_mission_item_int;
+	bool mission_item_int_support_confirmed;
+
 	static constexpr int BOOTUP_TIME_MS = 15000;	//! system startup delay before start pull
 	static constexpr int LIST_TIMEOUT_MS = 30000;	//! Timeout for pull/push operations
 	static constexpr int WP_TIMEOUT_MS = 1000;
 	static constexpr int RESCHEDULE_MS = 5000;
 	static constexpr int RETRIES_COUNT = 3;
+	static constexpr unsigned int MAV_PROTOCOL_CAPABILITY_MISSION_INT = 4;
+
```
vooon, line 398: Use `MAV_PROTOCOL_CAPABILITY::MISSION_INT`.

mavros/src/plugins/waypoint.cpp
```diff
@@ -98,33 +172,110 @@ class WaypointItem : public mavlink::common::msg::MISSION_ITEM {
...
 };
+template <class T>
+std::string waypoint_to_string(const T &waypoint){
+	return utils::format("#%u%1s F:%u C:%3u p: %f %f %f %f x: %i y: %i z: %i",
+		waypoint.seq,
+		(waypoint.current) ? "*" : "",
+		waypoint.frame, waypoint.command,
+		waypoint.param1, waypoint.param2, waypoint.param3, waypoint.param4,
+		waypoint.x, waypoint.y, waypoint.z);
+}
```
vooon, line 266: I think better to keep to_string in classes. You shouldn't use `%i` for float.
Also why not just to define common interface class for both variants?

mavros/src/plugins/waypoint.cpp
```diff
@@ -543,6 +818,13 @@ class WaypointPlugin : public plugin::PluginBase {
 			else {
 				enable_partial_push = m_uas->is_ardupilotmega();
 			}
+			if (wp_nh.hasParam("use_mission_item_int")) {
+				wp_nh.getParam("use_mission_item_int", use_mission_item_int);
+			}
+			else {
+				use_mission_item_int = m_uas->get_capabilities() & MAV_PROTOCOL_CAPABILITY_MISSION_INT;
+				mission_item_int_support_confirmed = use_mission_item_int;
```
vooon, line 826: As i remember, connection_cb() called before sys_status starts doing capabilities negotiation.

mavros/src/plugins/waypoint.cpp
bromeara: That might be better, so originally I had been planning on storing the variant type in the waypoint and send_waypoint vector and so I believe the lack of interface is left over from that because my first attempt was to give them a parent class and store that in the vector which didnt work very elegantly due to the difference in member variable types but I think I could give them a parent template class with this implementation which inherits from either MISSION_ITEM or MISSION_ITEM_INT and gives them a to_string in the class.

mavros/src/plugins/waypoint.cpp
bromeara: Ah thank you I was looking for that and forgot to look in the mavlink generated files.

mavros/launch/apm_config.yaml
```diff
@@ -122,7 +122,9 @@ setpoint_velocity:
 # waypoint
 mission:
   pull_after_gcs: true  # update mission if gcs updates
-
+  use_mission_item_int: true # use the MISSION_ITEM_INT message instead of MISSION_ITEM
+                             # for uploading waypoints to FCU
```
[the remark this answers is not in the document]
bromeara: if it is set to true but the flight controller still requests just a MISSION_ITEM like ardupilot then it will try using INT. This flag is more for explicitly turning it off, I'm not sure if it's necessary but it did make confirming that MISSION_ITEM was still functioning as intended easier.

mavros/src/plugins/waypoint.cpp
bromeara: This seems to be correct, I have not fixed this yet in the latest commits but is there a way that I can register a cb to the capabilities information being available or should I just do check in one of the actions that would execute earliest?

mavros/src/plugins/waypoint.cpp
bromeara: I realized that since I'm not using the classes to store data as I was originally intending I can now just give them to from ros message template functions and it should have cleaned everything up.

bromeara:
It seems that pymavlink.dialects.v20.common.enums['MAV_FRAME'] include some deprecated frame types which the CI build does not recognize. Is there a potential this would be fixed if cog was run on the build machine instead of mine?

.vscode/c_cpp_properties.json
```diff
@@ -0,0 +1,33 @@
...
+            ],
+            "name": "ROS",
+            "intelliSenseMode": "gcc-x64",
+            "compilerPath": "/usr/bin/gcc",
+            "cStandard": "c11",
+            "cppStandard": "c++17"
+        }
+    ],
+    "version": 4
+}
```
vooon, line 33: Please do not include editor configs that specific to your computer.

mavros/src/plugins/waypoint.cpp
```diff
@@ -28,67 +32,168 @@
...
+	case enum_value(MAV_FRAME::GLOBAL_TERRAIN_ALT):
+	case enum_value(MAV_FRAME::GLOBAL_TERRAIN_ALT_INT):
+		return 10000000;
+	case enum_value(MAV_FRAME::LOCAL_NED):
+	case enum_value(MAV_FRAME::LOCAL_ENU):
+	case enum_value(MAV_FRAME::LOCAL_OFFSET_NED):
+	case enum_value(MAV_FRAME::BODY_NED):
+	case enum_value(MAV_FRAME::BODY_OFFSET_NED):
+	case enum_value(MAV_FRAME::BODY_FRD):
+	case enum_value(MAV_FRAME::BODY_FLU):
```
vooon, line 74: Problem with that two frames arises because upstream mavlink changed recently.
I think that you just need to reinstall/update pymavlink to latest one, then run cog.

.vscode/c_cpp_properties.json
vooon: Note: you can add `.vscode` to `.gitignore`.

bromeara: Darn Ive been so careful about not staging that but was going to ask about the .gitignore.

bromeara:
@vooon  Would adding a handle for the autopilot_version message in the waypoint node interfere with the sys_status node? and if not would this be a good way to go about detecting the capabilities since there isn't really a setup function that runs after the capabilities negotiation happens.

mavros/src/plugins/waypoint.cpp
bromeara: Ok, so I updated both my upstream mavlink and pymavlink and it didn't change, then I installed pymavlink from the submodule in mavlink and it did work so I'm thinking they haven't fixed it in the pip repository yet? But the code should be correct now and hopefully they will update it by the time it needs to be generated again.

mavros/src/plugins/waypoint.cpp
vooon: As i remember there no such event in UAS yet. So i think you should add event processor to the UAS, then update base class to have `capabilities_cb()` and methods to support it.

That would be useful for other plugins too.

mavros/src/plugins/waypoint.cpp
vooon: Yes. As i remember pip releases does @tridge with different schedule.
I prefer to use names and not hardcode values, so yes, that code would require update later.
But because we keep generator code, that should be easy.

vooon:
Many plugins may subscribe to the same message, the is not a problem.
But in case of autopilot_version i think better to extend the plugin api to have method similar to `connection_cb()`.

bromeara:
Do you believe capabilities_cb() would be sufficient or are there other pieces of data resolved after connection that could perhaps have a blanket term like uas_info_cb(). I will be looking into this separately as well as I go to implement capabilities_cb(), but wanted to see if there was something you could think of that I might overlook.

vooon:
@bromeara sorry for the delay. I think it's ok to have capability function here too.
I'm not best at English, but i think better to use plural form, like in the message, so `capabilities_cb()`.

On sys_status it's a little trick. Find or create returns iterator of object that were emplaced to the map.
So accessing trough it.second ptr changes it's value.
What i'm a bit unsure - that we probably should use rwlock on that map, but on other hand, all callbacks runs on the same io thread, so it's unlikely to have race problem.

mavros/include/mavros/mavros_uas.h
```diff
@@ -65,17 +65,20 @@ namespace mavros {
...
 	using MAV_TYPE = mavlink::common::MAV_TYPE;
 	using MAV_AUTOPILOT = mavlink::common::MAV_AUTOPILOT;
 	using MAV_MODE_FLAG = mavlink::common::MAV_MODE_FLAG;
 	using MAV_STATE = mavlink::common::MAV_STATE;
+	using MAV_CAP = mavlink::common::MAV_PROTOCOL_CAPABILITY;
 	using timesync_mode = utils::timesync_mode;
 
+	// other UAS aliases
+	using ConnectionCb = std::function<void(bool)>;
+	using CapabilitiesCb = std::function<void(MAV_CAP)>;
```
bromeara, line 78: I have the capabilities callback take a MAV_PROTOCOL_CAPABILITY enum in order to increase visibility of where people can find the cpp enum to use in their code. After adding the has_capability / has_capabilities functions to UAS I don't think this is needed any longer and it could be a void func(); What do you think?

mavros/include/mavros/mavros_uas.h
```diff
@@ -310,8 +313,43 @@ class UAS {
 
 	/* -*- autopilot version -*- */
 	uint64_t get_capabilities();
+
+	/**
+	 * @brief Function to check if the flight controller has a capability
+	 *
+	 * @param capabilities can accept a multiple capability params either in enum or int from
+	 */
+	template<typename T>
```
bromeara, line 322: While these template functions are useful, I fear they may be increasing compile time, although Im not sure how to measure. The same effect could be accomplished with overloading functions but it would be more code and more runtime indirection. Let me know if these seem problematic.

mavros/include/mavros/mavros_uas.h
bromeara, line 78: I have the capabilities callback take a MAV_PROTOCOL_CAPABILITY enum in order to increase visibility of where people can find the cpp enum to use in their code. After adding the has_capability / has_capabilities functions to UAS I don't think this is needed any longer and it could be a void func(); What do you think?

mavros/include/mavros/mavros_uas.h
vooon: Well, here is following pro and cons:
- template may took more time, that is a problem for raspberry-pi users, if they had to compile
- overloaded functions requires more code from the human, so more room for error

From that two statements i prefer to have less code done by hand because of principle: smaller code - better code.

mavros/include/mavros/mavros_uas.h
vooon: Leave it with parameter, that would be more explicit.

mavros/include/mavros/mavros_uas.h
vooon: Keep them, looks good to have.

mavros/include/mavros/mavros_uas.h
```diff
@@ -310,8 +313,43 @@ class UAS {
...
+	bool has_capability(T capability){
+		return static_cast<uint64_t>(fcu_capabilities) & static_cast<uint64_t>(capability);
```
vooon, line 324: Because it's atomic uint64_t better first to copy it's state, then do OR.
Also maybe better to use `enum_value()` for capability cast.

bromeara:
I agree with assessment, I would think that if compile time is a bigger concern than readability, an alternate would be to employ cog to reduce human error but would still increase code affecting aspects of readability.
  

  
  I favor the compile-time approach as well though because I would hope raspberry pi users are pulling from the system repository and doing development work on more robust systems/cross compiling.
  

  
  Should I open and close an issue that documents the compile time slowdown for those users who might not have those options and they can edit it out themselves?
  

  
  
  

  
  
>

mavros/include/mavros/mavros_uas.h
vooon: Probably you should use `get_capabilities()` here.

mavros/src/plugins/waypoint.cpp
vooon: Please remove that line, as it is not used anymore.

vooon:
@bromeara there are several ways to speed up compile time, like usage of fast usb thumb stick / ssd and distcc.
Also there several distributions that builds images with relatively fresh mavros, like Maverick. So probably a couple of seconds of compile time not worth to make a bigger code.

If you can, it would be nice to have instructions on how to speedup builds.

bromeara:
I have tested this code with my use case on Ardupilot. When still relying on the parameter I was able to confirm pushing and pulling waypoints in both float and int operated as expected. Then after I added the capability callback I confirmed it could operate from an int state and also commented out some lines to get it operating in the float state since I didn’t   know an easy way to set sitl capabilities.   
  

  
If you or someone you know has more familiarity with PX4 sitl and could push some waypoints to make sure its not a breaking change I could make any changes to the code needed. I just had trouble figuring out the setup of the px4 sitl on my system.
  

  
  
  

  
  
>

vooon:
I think you can use docker container: https://dev.px4.io/v1.9.0/en/test_and_ci/docker.html
Also you can ask for help on PX4 Forums.

mavros/include/mavros/mavros_uas.h
bromeara: Ok, I'll use get_capabilities(). The reason I didn't use enum_value() is that underlying_type on a non-enum value is undefined behavior. I wrote the templated code so that either the int code or the Enum value could be used to check the capabilities, should I only accept enum values? This would remove the need for a  template I belive which might speed up the compile time and force people not to use magic numbers.

bromeara:
Are the issues the best place to document this if I end up keeping the templated code?

I am not familiar with the automated tests that are run, are these changes that should have tests created for them? I will try out the PX4 testing tonight, so probably tomorrow for you.

bromeara:
is there a way to resolve the travis build failure it looks like a mavconn error but Im not 100%

`<failure message="Unable to find test results for gtest-mavconn-test.xml, test did not run.
Expected results in /root/catkin_ws/build/libmavconn/test_results/libmavconn/gtest-mavconn-test.xml" type=""/>`

mavros/include/mavros/mavros_uas.h
bromeara: I am using static_cast instead of enum_value since the template can also accept ints and underlying_type on non-enum values is undefined behavior. Thinking about it though I did switch to enum_value and added a static assert to give a warning message if a non enum type is used so users cant use a magic number.

vooon:
About docs, i think better to try to go with github wiki.

PR #1416, opened 2020-04-28
mavros_msgs/CompanionProcessStatus: Add USER1 enum value

mortenfyhn:
Not much to say. Useful to me, maybe useful to someone else in the future.

For convenience: https://mavlink.io/en/messages/common.html#MAV_COMP_ID_USER1


mortenfyhn:
Oh right, I must have misunderstood (not uncommon :laughing: ). Is the proper use case that you have a person/group who needs a component id specific to their usage, and then they use component id 25-99 and give them meaningful names, and the user1..99 names are just placeholders?

mortenfyhn:
In any case I'll close this.

PR #1434, opened 2020-05-28
Feature obstacle distance fix

russellfromm:
Hello,

I was having some issues with ardupilot's Obstacle Avoidance Database; obstacles were showing up in unexpected places. The laserscanner on our drone is installed so that the forward vector is backwards, so I rotated the laserscan by 180 degrees on the z-axis. The rotation seemed to cause a strange discontinuity in the returned obstacles. So, I visualized the adsb vehicle messages I was receiving (green circles), the testing laserscan (red) and the obstacle_distance message reinterpreted as a laserscan (yellow). The results of this can be seen below. The obstacle_distance message has not been rotated to align with the laserscan; partially because of visibility, but mostly because of laziness.

original

As you can see, the initial part of the adsb messages are quite off the actual laserscan. You may also notice that the obstacle_distance's distance measurements exceed 360 degrees around; I believe the latter leads to the former. The increment appears to be the culprit here. The laserscan I'm providing as 2048 ranges; far exceeding the obstacle_distance's size limit of 72. In that case, two issues arise. Firstly, the obstacle's increment is set, but not the increment_f. These are both set when the number of ranges is less than 72, so it seems odd that it was not set in this case. Secondly, the increment is rounded upward, resulting in significant overlap which I believe causes that strange behavior in the adsb messages. Once again, this is not the case when a laserscan has fewer than 72 ranges; it is rounded normally.

Upon implementing the changes, I saw marked improvement of the adsb message accuracy. For reference, here's the result using the new increment_f:

float

And here's the normally rounded int increment:

int

As you can see, much improved. There is still a slight gap in the spiral, but it's far less of an issue.

### DIFF
--- a/mavros_extras/src/plugins/obstacle_distance.cpp
+++ b/mavros_extras/src/plugins/obstacle_distance.cpp
@@ -101,7 +101,9 @@ private:
 					}
 				}
 			}
-			obstacle.increment = ceil(req->angle_increment * RAD_TO_DEG * scale_factor);	//!< [degrees]
+			const float increment_deg = req->angle_increment * RAD_TO_DEG * double(fixed.ranges.size()) / obstacle.distances.size();
+			obstacle.increment = static_cast<uint8_t>(increment_deg + 0.5f);  //!< Round to nearest integer.
+			obstacle.increment_f = increment_deg;
 		}
 
 		obstacle.time_usec = req->header.stamp.toNSec() / 1000;					//!< [microsecs]
--- a/mavros_extras/src/plugins/obstacle_distance.cpp
+++ b/mavros_extras/src/plugins/obstacle_distance.cpp
@@ -84,7 +84,7 @@ private:
 			std::fill(obstacle.distances.begin() + req->ranges.size(), obstacle.distances.end(), UINT16_MAX);	//!< fill the rest of the array values as "Unknown"
 
 			const float increment_deg = req->angle_increment * RAD_TO_DEG;
-			obstacle.increment = static_cast<uint8_t>(increment_deg + 0.5f);  //!< Round to nearest integer.
+			obstacle.increment = static_cast<uint8_t>(increment_deg + 0.5f);	//!< Round to nearest integer.
 			obstacle.increment_f = increment_deg;
 		} else {
 			// all distances from sensor will not fit so we combine adjacent distances always taking the shortest distance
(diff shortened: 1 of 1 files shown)


TSC21:
@jkflying @mrivi

vooon:
CC @rmackay9

mavros_extras/src/plugins/obstacle_distance.cpp
```diff
@@ -101,7 +101,9 @@ class ObstacleDistancePlugin : public plugin::PluginBase {
 					}
 				}
 			}
-			obstacle.increment = ceil(req->angle_increment * RAD_TO_DEG * scale_factor);	//!< [degrees]
```
rmackay9, line 104: It certainly looks like master has a bug in that it's not setting the obstacle.increment_f.

I'm not a mavros expert so take this with a grain of salt but where is "fixed.ranges.size()" coming from?  Is it the same as req->ranges.size?

I think the new increment_deg calculation is maybe not correct.  It seems to be using obstacle.distances.size() which (i think) is always 72 but I think it instead needs to use the number of slots in obstacle.distsances that have actually been filled in with values.

> const float increment_deg = req->angle_increment * RAD_TO_DEG * scale_factor;
> obstacle.increment = static_cast<uint8_t>(increment_deg + 0.5f);  //!< Round to nearest integer.
> obstacle.increment_f = increment_deg;

As a small style thing we could move the set for obstacle.increment_f two lines higher and remove the intermediate variable increment_deg but it doesn't really matter.  The similar code several lines higher does it this way.

rmackay9:
I've put comments above with an alternative solution but in any case it's great that you've found a bug and nice testing diagrams.

mavros_extras/src/plugins/obstacle_distance.cpp
russellfromm: Ah, fixed was a result of my attempts to rotate the laserscan; I must have accidentally left them there. My apologies.

Looking at the code again, if you have more than 72 values in your scan, it will fill in all 72 slots in the obstacle distance. Unless you're talking about NaN values and such.

rmackay9: Russell, So you agree that my suggestion for how obstacle.increment_f should be calculated is correct?  If yes maybe give it a try if you have time and update the PR?

I think the calculation of the obstacle.increment_f as it is currently in this PR will end up with the objects appearing in the wrong place on the receiver side.

Thanks again for finding this issue and trying to get it fixed.

oroelipas:
Hey! 

I found this bug also by myself, and it is correctly fixed with the changes proposed. 
increment f has no value assigned in case the input LaserScan given has more than 72 values. 
Please accept the PR.

Thanks @russellfromm

oroelipas:
Hi @TSC21,
I am not a Git expert, do you want me to create a new PR in which the new commits are ahead the last commit in the master branch of this repo?

I would be glad to participate :)

TSC21:
> Hi @TSC21, I am not a Git expert, do you want me to create a new PR in which the new commits are ahead the last commit in the master branch of this repo?
> 
> I would be glad to participate :)

`git pull --rebase origin master`

oroelipas:
Hi @TSC21 
I created a new PR with this (an another) bugfix.
I already tested, I hope everything is correct.

vooon:
Replaced by #1700.

PR #1453, opened 2020-07-23
Add ESC Status plugin

RicardoM17:
Relates to: https://github.com/mavlink/mavlink/pull/1322

Any tips and suggestions are welcome.

### DIFF
--- a/mavros_extras/src/plugins/esc_status.cpp
+++ b/mavros_extras/src/plugins/esc_status.cpp
@@ -1,0 +1,154 @@
+/**
+ * @brief ESC status plugin
+ * @file esc_status.cpp
+ * @author Ricardo Marques <marques.ricardo17@gmail.com>
+ *
+ * @addtogroup plugin
+ * @{
+ */
+/*
+ * Copyright 2020 Ricardo Marques <marques.ricardo17@gmail.com>.
+ *
+ * This file is part of the mavros package and subject to the license terms
+ * in the top-level LICENSE file of the mavros repository.
+ * https://github.com/mavlink/mavros/tree/master/LICENSE.md
+ */
+
+#include <mavros/mavros_plugin.h>
+#include <mavros_msgs/ESCInfo.h>
+#include <mavros_msgs/ESCStatus.h>
+
+namespace mavros
+{
+namespace extra_plugins
+{
+/**
+ * @brief ESC status plugin
+ */
+class ESCStatusPlugin : public plugin::PluginBase
+{
+public:
+	ESCStatusPlugin() : PluginBase(),
+		_max_esc_count(0),
+		_max_esc_info_index(0),
+		_max_esc_status_index(0),
+		nh("~")
+	{}
+
+	void initialize(UAS &uas_)
+	{
+		PluginBase::initialize(uas_);
+
+		esc_info_pub = nh.advertise<mavros_msgs::ESCInfo>("esc_info", 10);
+		esc_status_pub = nh.advertise<mavros_msgs::ESCStatus>("esc_status", 10);
+
+		enable_connection_cb();
+	}
+
+	Subscriptions get_subscriptions()
+	{
+		return {
+			       make_handler(&ESCStatusPlugin::handle_esc_info),
+			       make_handler(&ESCStatusPlugin::handle_esc_status),
+		};
+	}
+
+private:
+	using lock_guard = std::lock_guard<std::mutex>;
+	std::mutex mutex;
+
+	ros::NodeHandle nh;
+
+	ros::Publisher esc_info_pub;
+	ros::Publisher esc_status_pub;
+	mavros_msgs::ESCInfo _esc_info;
+	mavros_msgs::ESCStatus _esc_status;
+	uint8_t _max_esc_count;
+	uint8_t _max_esc_info_index;
+	uint8_t _max_esc_status_index;
+	const uint8_t batch_size = 4;
+
+	void handle_esc_info(const mavlink::mavlink_message_t *msg, mavlink::common::msg::ESC_INFO &esc_info)
+	{
+		lock_guard lock(mutex);
+
+		_esc_info.header.stamp = m_uas->synchronise_stamp(esc_info.time_usec);
+
+		uint8_t esc_index = esc_info.index;
+
+		_esc_info.counter = esc_info.counter;
+		_esc_info.count = esc_info.count;
+		_esc_info.connection_type = esc_info.connection_type;
+		_esc_info.info = esc_info.info;
+
+		if (_esc_info.count > _max_esc_count)
+		{
+			_max_esc_count = _esc_info.count;
+		}
+
+		if (_esc_info.esc_info.size() < _max_esc_count)
+		{
+			_esc_info.esc_info.resize(_max_esc_count);
+		}
+
+		for (int i = 0; i < batch_size; i++)
+		{
+			_esc_info.esc_info[esc_index + i].header =  _esc_info.header;
+			_esc_info.esc_info[esc_index + i].failure_flags = esc_info.failure_flags[i];
+			_esc_info.esc_info[esc_index + i].error_count = esc_info.error_count[i];
+			_esc_info.esc_info[esc_index + i].temperature = esc_info.temperature[i];
+		}
+
+		_max_esc_info_index = std::max(_max_esc_info_index, esc_info.index);
+
+		if (_max_esc_info_index == esc_info.index)
+		{
+			esc_info_pub.publish(_esc_info);
+		}
+	}
+
+	void handle_esc_status(const mavlink::mavlink_message_t *msg, mavlink::common::msg::ESC_STATUS &esc_status)
+	{
+		lock_guard lock(mutex);
+
+		uint8_t esc_index = esc_status.index;
+
+		if (_esc_status.esc_status.size() < _max_esc_count)
+		{
+			_esc_status.esc_status.resize(_max_esc_count);
+		}
+
+		_esc_status.header.stamp = m_uas->synchronise_stamp(esc_status.time_usec);
+
+		for (int i = 0; i < batch_size; i++)
+		{
+			_esc_status.esc_status[esc_index + i].header = _esc_status.header;
+			_esc_status.esc_status[esc_index + i].rpm = esc_status.rpm[i];
+			_esc_status.esc_status[esc_index + i].voltage = esc_status.voltage[i];
+			_esc_status.esc_status[esc_index + i].current = esc_status.current[i];
+		}
+
+		_max_esc_status_index = std::max(_max_esc_status_index, esc_status.index);
+
+		if (_max_esc_status_index == esc_status.index)
+		{
+			esc_status_pub.publish(_esc_status);
+		}
+	}
+
+	void connection_cb(bool connected) override
+	{
+		lock_guard lock(mutex);
+
+		_max_esc_count = 0;
+		_max_esc_status_index = 0;
+		_max_esc_info_index = 0;
+		_esc_info.esc_info.resize(0);
+		_esc_status.esc_status.resize(0);
+	}
+};
+}	// namespace extra_plugins
+}	// namespace mavros
+
+#include <pluginlib/class_list_macros.h>
+PLUGINLIB_EXPORT_CLASS(mavros::extra_plugins::ESCStatusPlugin, mavros::plugin::PluginBase)
--- a/mavros_extras/src/plugins/esc_status.cpp
+++ b/mavros_extras/src/plugins/esc_status.cpp
@@ -91,9 +91,9 @@ private:
 			_esc_info.esc_info.resize(_max_esc_count);
 		}
 
-		for (int i = 0; i < batch_size; i++)
+		for (int i = 0; i < std::min<ssize_t>(batch_size, ssize_t(_max_esc_count) - esc_index); i++)
 		{
-			_esc_info.esc_info[esc_index + i].header =  _esc_info.header;
+			_esc_info.esc_info[esc_index + i].header = _esc_info.header;
 			_esc_info.esc_info[esc_index + i].failure_flags = esc_info.failure_flags[i];
 			_esc_info.esc_info[esc_index + i].error_count = esc_info.error_count[i];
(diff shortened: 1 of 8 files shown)


RicardoM17:
Hi @vooon thanks for the feedback. I made the PR just to get the conversation going, and as such I was already expecting quite some changes.

If I understand correctly you suggest not sending the messages in batches of 4 as they are coming out of MAVLink, but storing these in the plugin and always sending the information of all ESCs as soon as we get new information of a new batch. Is that correct? If so I have a few questions.

First instead of going back and forth to find the best implementation I ask if there's any existing mavros plugin that works similar to this that I could have a look into it's logic. If not any pointers here would help.

Secondly, again assuming I understood correctly your suggested approach, in the new message types there is no specified size for each field or item. They would all have the size that would come from the `count` field in the `ESC_INFO` mavlink message is that correct?

Lastly I performed uncrustify on the `esc_status.cpp` I believe that was it? let me know if that's ok. I did it also in the messages but the results were not good :sweat_smile: 

Thanks once again

mavros_msgs/msg/ESCStatus.msg
```diff
@@ -0,0 +1,14 @@
...
+# https://mavlink.io/en/messages/common.html#ESC_STATUS
+
+uint8 ESC_PER_MESSAGE = 4
+
+std_msgs/Header header
+
+uint8 index
+int32[4] rpm
+float32[4] voltage
+float32[4] current
```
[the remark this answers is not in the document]
RicardoM17: Just to confirm here, these would then be 2 more separate messages, in separate files (for a total of 4) and `ESCStatusItem[] statuses` should actually be something like `mavros_msgs/ESCStatusItem[] statuses`

vooon:
No, we don't have exactly same solution, but simpler version can be found here: https://github.com/mavlink/mavros/blob/master/mavros/src/plugins/rc_io.cpp#L71

My idea that you'll parse and keep results, then publish them all in one big message.
Each mavlink message can transfer states for 4 motors, so only last one would have count<=4 (e.g. for hexacopter you'll send 2 messages with 4 and 2 states, for 14-copter 3 (x4) + 1 (x2)).
So it's relatively easy to calculate indexes for storing vector.

Then you should check that msg.index == max_index_observed and publish.

Pseudocode:
```
void handle_esc_status(in) {
    with _mutex.lock

    for i=0, i<in.count, i++ {
        _esc_status[4*in.index + i] = parse(in, i)
    }

    _max_index = max(_max_index, in.index)
    if _max_index == in.index {
        publish(_esc)
    }
}

void on_connection(bool) {
    _max_index = 0;
}
```

RicardoM17:
Hi again @vooon that makes sense. I already have an "ugly" version working but it's unlikely I will finish this today or over the weekend as I have to better understand the mutex lock that you mention as I'm not familiar with it. But I assume I can reuse that part from the example you mentioned as it seems fairly simple.

If you want I can move this to WIP but definitely I will be making some progress on it next week.

Thanks and have a great weekend :)

mavros_extras/src/plugins/esc_status.cpp
```diff
@@ -0,0 +1,135 @@
...
+		}
+	}
+
+	void handle_esc_status(const mavlink::mavlink_message_t *msg, mavlink::common::msg::ESC_STATUS &esc_status)
+	{
+		lock_guard lock(mutex);
+
+		uint8_t esc_index = esc_status.index;
+
+		_esc_status.esc_status.resize(8);
```
vooon, line 104: Same here.

mavros_extras/src/plugins/esc_status.cpp
```diff
@@ -0,0 +1,135 @@
...
+namespace extra_plugins
+{
+/**
+ * @brief ESC status plugin
+ */
+class ESCStatusPlugin : public plugin::PluginBase
+{
+public:
+	ESCStatusPlugin() : PluginBase(),
+		nh("~")
```
vooon, line 32: Please add _max_esc_info_index & _max_esc_status_index to initializer list.
It's actually not necessary, but i think better to have more explicit code.

mavros_extras/src/plugins/esc_status.cpp
```diff
@@ -0,0 +1,135 @@
...
+			esc_status_pub.publish(_esc_status);
+		}
+	}
+
+	void connection_cb(bool connected) override
+	{
+		lock_guard lock(mutex);
+
+		_max_esc_status_index = 0;
+		_max_esc_info_index = 0;
```
vooon, line 128: Need also resize vectors to zero size.

mavros_extras/src/plugins/esc_status.cpp
```diff
@@ -0,0 +1,135 @@
...
+		_esc_info.header.stamp = m_uas->synchronise_stamp(esc_info.time_usec);
+
+		uint8_t esc_index = esc_info.index;
+
+		_esc_info.counter = esc_info.counter;
+		_esc_info.count = esc_info.count;
+		_esc_info.connection_type = esc_info.connection_type;
+		_esc_info.info = esc_info.info;
+
+		_esc_info.esc_info.resize(8);	//_esc_info.count
```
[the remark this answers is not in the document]
RicardoM17: Missed this part. I was testing at first with just size 8 to keep things simple and then I forgot to update (hence the _esc_info.count comment there)

But actually I think there's a slight error in your suggestion

`esc_info.count` should have the TOTAL number of esc in the messages (even if the messages can only house up to 4). so here I can just resize(esc_info.count)

The issue is that I can't really do the same for the esc_status part but there I can use some logic similar to what you suggest. probably checking if one of the parameters is empty or not.

mavros_extras/src/plugins/esc_status.cpp
RicardoM17: Sounds good.

mavros_extras/src/plugins/esc_status.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+		for (int i = 0; i < batch_size; i++)
+		{
+			_esc_info.esc_info[esc_index + i].header.stamp =  m_uas->synchronise_stamp(esc_info.time_usec);
```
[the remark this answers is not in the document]
RicardoM17: will do.

mavros_extras/src/plugins/esc_status.cpp
```diff
@@ -0,0 +1,148 @@
...
+		}
+	}
+
+	void handle_esc_status(const mavlink::mavlink_message_t *msg, mavlink::common::msg::ESC_STATUS &esc_status)
+	{
+		lock_guard lock(mutex);
+
+		uint8_t esc_index = esc_status.index;
+
+		size_t required_size = batch_size * (esc_index ) + batch_size;	// TODO FIX. We should only add as many fields as they exist.
```
RicardoM17, line 110: @vooon How should we proceed here. I can see several options but not really the "optimal" one.

1. Leave as is. It means that if we are receiving messages from a Hexacopter we will be publishing vector size 8 when size 6 would suffice. Pros: Simple and foolproof.

2. Incorporate the data from the `esc_info` message and so: `_esc_info.esc_info.resize(_esc_info.count);` . Cons: It requires a second message and if a user is only sending the `ESC_STATUS` mavlink message this won't really work.

3. Introduce a count field in the `ESC_STATUS` mavlink index. Cons: Mavlink message gets bigger but not that much tbf. It just requires a change from the start but it can be done.

4. Analyze the data fields in the `ESC_STATUS` message. This should be defined from the MAVLink side but if that field is empty it should have the max value. I guess in this case to be ultra safe we should check this for both `rpm` `voltage` and `current` . This will prevent messages being dropped just because a certain field is missing.

5. ??

Let me know what you think.

vooon: Umm, i thought that `esc_info.count` represent count of that particular message (wrong assumption based on older massages).

So yes, for esc_info vector we just `.resize(count)`. Then we should loop `i=0; i < min(count - 4*index, 4); i++`.

vooon: And we should use esc_info.count to process esc_status too. As both messages should have equal amount of items.
Status parser should check that message can fit into the vector, otherwise drop message with warning.

mavros_extras/src/plugins/esc_status.cpp
```diff
@@ -0,0 +1,148 @@
...
+		{
+			_esc_info.esc_info.resize(_esc_info.count);
+		}
+
+		for (int i = 0; i < batch_size; i++)
+		{
+			_esc_info.esc_info[esc_index + i].header =  _esc_info.header;
+			_esc_info.esc_info[esc_index + i].failure_flags = esc_info.failure_flags[i];
+			_esc_info.esc_info[esc_index + i].error_count = esc_info.error_count[i];
+			_esc_info.esc_info[esc_index + i].temperature = esc_info.temperature[i];
```
vooon, line 93: Error in index calculation: you should: `batch_size * esc_index + i`.

```
auto &e = _esc_info.esc_info[batch_size * esc_index + i];
e.header = _esc_info.header
...
```

mavros_extras/src/plugins/esc_status.cpp
```diff
@@ -0,0 +1,148 @@
...
+		}
+
+		_esc_status.header.stamp = m_uas->synchronise_stamp(esc_status.time_usec);
+
+		for (int i = 0; i < batch_size; i++)
+		{
+			_esc_status.esc_status[esc_index + i].header = _esc_status.header;
+			_esc_status.esc_status[esc_index + i].rpm = esc_status.rpm[i];
+			_esc_status.esc_status[esc_index + i].voltage = esc_status.voltage[i];
+			_esc_status.esc_status[esc_index + i].current = esc_status.current[i];
```
vooon, line 123: Same issue with index calculation.

mavros_extras/src/plugins/esc_status.cpp
RicardoM17: esc_index is either 0, 4, 8 etc

MAVLink message:

`<field type="uint8_t" name="index">Index of the first ESC in this message. minValue = 0, maxValue = 60, increment = 4.</field>`

So I think like this it's correct no? Let me know if further changes are needed here.

mavros_extras/src/plugins/esc_status.cpp
RicardoM17: Sorry @vooon this became a bit confusing as with the other message there seems to be a slight confusion here with count, index etc

Should I just store in a variable the `_max_esc_count` define it in the `handle_esc_info` and then also use it in the `handle_esc_status` ? I.e. suggestion 2 in my comment? This assumes that both messages are always being published however, hence my question to you.

vooon: Yes, just store total number of escs. I assume that usage of only ESC_STATUS is agains spec, so we shouldn't think host to support that.

mavros_extras/src/plugins/esc_status.cpp
vooon: Oh, again i am not reading carefully. But that also based on other messages, where index increment is 1.

mavros_extras/src/plugins/esc_status.cpp
RicardoM17: Sounds good. I'll do that.

However if you feel like suggestion number 3 is a bit more robust let me know and I can work towards that after this PR is merged. The mavlink messages are WIP so this is one of those cases where we can improve it after the fact.

mavros_extras/src/plugins/esc_status.cpp
RicardoM17: no worries @vooon I understand and it's why I'm here :)

mavros_extras/src/plugins/esc_status.cpp
```diff
@@ -0,0 +1,154 @@
...
+	void handle_esc_status(const mavlink::mavlink_message_t *msg, mavlink::common::msg::ESC_STATUS &esc_status)
+	{
+		lock_guard lock(mutex);
+
+		uint8_t esc_index = esc_status.index;
+
+		if (_esc_status.esc_status.size() < _max_esc_count)
+		{
+			_esc_status.esc_status.resize(_max_esc_count);
+		}
```
vooon, line 119: Because we get count from other message we have to check `index > count` and drop the message with warning.
That possible if we first receive ESC_STATUS and then ESC_INFO, that would result in segfault without boundaries check.

vooon: Another idea (signed math): `min<ssize_t>(batch_size, ssize_t(count) - index)`.

mavros_extras/src/plugins/esc_status.cpp
```diff
@@ -0,0 +1,154 @@
...
+		{
+			_max_esc_count = _esc_info.count;
+		}
+
+		if (_esc_info.esc_info.size() < _max_esc_count)
+		{
+			_esc_info.esc_info.resize(_max_esc_count);
+		}
+
+		for (int i = 0; i < batch_size; i++)
```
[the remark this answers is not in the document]
RicardoM17: Good call.

RicardoM17:
Hi again @vooon I made some changes but I'm not entirely sure about what you mean on the remaining comment that you have.

For the record in my tests, with the pushed version, when I don't publish the `ESC_INFO` MAVLink message from the FCU side (so only publishing `ESC_STATUS`) I get no published messages on the `/mavros/esc_info` topic as expected and only empty ones in the `/mavros/esc_status` topic and no errors. which is not perfect but reasonable.

I guess as a final step we could add some sort of check and only publish those topics  `if ( _max_esc_info_index == esc_info.index && _max_esc_count > 0)` . Or should we be publishing empty messages?

vooon:
Will merge after mavlink package release, because now CI fails.

RicardoM17:
No worries @vooon . Thanks a lot for all the reviews and suggestions. It's very easy to say that the final result is better than the initial approach and also that I learned quite a bit with these revisions.

Let me know if I do need to rebase this at some point beore you merge if not I'm very happy to have contributed to MAVROS and I hope to do so more in the future, and to have more PRs for you to review ;)

PR #1459, opened 2020-08-01
Add support for reporting rtk_baseline over ROS

ajshank:
Mavlink message [127 (GPS_RTK)](https://mavlink.io/en/messages/common.html#GPS_RTK) allows the FCU to report relative positions from an RTK base-station. In some multirobot applications I'm working on, I need this data to align maps to a common frame of reference (with origin at the base-station). This PR is a relatively simple addition to the `gps_rtk` plugin that already exists in `mavros_extras` (but only sends RTCM corrections). It now defines a publisher over `~rtk_baseline` topic, and adds a new ROS message in `mavros_msgs`.
I've field-tested that this new publisher works as expected on Copter 4.0. I suppose that message 128 (GPS2_RTK) would be somewhat similar, but I don't have extra hardware to test this right now.

Hopefully this is useful in general to folks who are using RTK.


ajshank:
Great, thanks @vooon. 
Apologies if this is silly, but I've not been able to deduce a good reason to do heap allocations with `boost::make_shared<>()` for ROS message variables (compared to declaring an object on the stack). This is not a problem for this plugin since there are no additional conversions that are done, but could you point to why it is done so [here](https://github.com/ajshank/mavros/blob/920887e98fc22c474e51645f1f6d5131de4d77b2/mavros/src/plugins/global_position.cpp#L161) for instance? I suppose at high-rates (50-100Hz) on small computers (Gumstix, Rasp.Pi etc) this could be a performance hit?

vooon:
That from some old best practices for ROS, passing smart pointer makes it possible to more effectively pass that message intra-process.
Not the case for ROS2...

PR #1477, opened 2020-08-31
Add error message in install_geographiclib_datasets.sh

bart02:
Now script hasn't got error handler. Some time ago it was the problem with SourceForge mirrors. But this script didn't print any errors. And it was so difficult to understand, what the problem with dataset.

After path if dataset isn't downloaded, the script will write an error meaasge.


vooon:
LGTM Thanks!

PR #1480, opened 2020-09-09
play_tune: Write new plugin

mortenfyhn:
Now that PX4 has proper handling of incoming PLAY_TUNE/PLAY_TUNE_V2 mavlink messages, I've written a tiny plugin that allows sending these tunes from ROS.

I've added a mavros_msgs/PlayTuneV2 for this, to make the interface clear, which includes the tune format and tune string.

Tested on a pixhawk with PX4 v1.11, and it plays as expected.

### DIFF
--- a/mavros_extras/src/plugins/play_tune.cpp
+++ b/mavros_extras/src/plugins/play_tune.cpp
@@ -6,52 +6,52 @@ namespace mavros
 {
 namespace extra_plugins
 {
-
 class PlayTunePlugin : public plugin::PluginBase
 {
-  public:
-    PlayTunePlugin() : PluginBase(), nh("~") {}
+public:
+	PlayTunePlugin() : PluginBase(), nh("~") {}
 
-    void initialize(UAS& uas_) override
-    {
-        PluginBase::initialize(uas_);
-        sub = nh.subscribe("play_tune", 1, &PlayTunePlugin::callback, this);
-    }
+	void initialize(UAS& uas_) override
+	{
+		PluginBase::initialize(uas_);
+		sub = nh.subscribe("play_tune", 1, &PlayTunePlugin::callback, this);
+	}
 
-    Subscriptions get_subscriptions() override { return { /* No subscriptions */ }; }
+	Subscriptions get_subscriptions() override {
+		return { /* No subscriptions */ };
+	}
 
-  private:
-    ros::NodeHandle nh;
-    ros::Subscriber sub;
+private:
+	ros::NodeHandle nh;
+	ros::Subscriber sub;
 
-    void callback(const mavros_msgs::PlayTuneV2::ConstPtr& tune)
-    {
-        auto msg = mavlink::common::msg::PLAY_TUNE_V2{};
-        m_uas->msg_set_target(msg);
+	void callback(const mavros_msgs::PlayTuneV2::ConstPtr& tune)
+	{
+		auto msg = mavlink::common::msg::PLAY_TUNE_V2{};
+		m_uas->msg_set_target(msg);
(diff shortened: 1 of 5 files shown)


mavros_extras/src/plugins/play_tune.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+  private:
+    ros::NodeHandle nh;
+    ros::Subscriber sub;
+
+    void callback(const mavros_msgs::PlayTuneV2::ConstPtr& tune)
+    {
+        auto msg = mavlink::common::msg::PLAY_TUNE_V2{};
+        msg.target_system = m_uas->get_tgt_system();
+        msg.target_component = m_uas->get_tgt_component();
```
vooon, line 31: Please use `UAS::msg_set_target()`

https://github.com/mavlink/mavros/blob/master/mavros/include/mavros/mavros_uas.h#L387

mavros_extras/src/plugins/play_tune.cpp
```diff
@@ -0,0 +1,59 @@
...
+                break;
+
+            case mavros_msgs::PlayTuneV2::TUNE_FORMAT_MML_MODERN:
+                msg.format = static_cast<uint32_t>(mavlink::common::TUNE_FORMAT::MML_MODERN);
+                break;
+
+            default:
+                ROS_ERROR_NAMED("play_tune", "Invalid tune format: '%i'", tune->format);
+                return;
+        }
```
vooon, line 46: Usually that is not necessary, we just follow mavlink enums in msgs.

mavros_extras/src/plugins/play_tune.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+        std::strncpy(msg.tune.data(), tune->tune.c_str(), sizeof(msg.tune));
+        msg.tune.back() = '\0';
```
vooon, line 49: Please use `mavlink::set_string_z()`.

https://github.com/ArduPilot/pymavlink/blob/b1efa7e5fd50561dfa9f8a491e4a55c8e0eedec9/generator/CPP11/include_v2.0/message.hpp#L143

vooon (changes requested):
Overall plugin looks good, but few changes needed:
1. see comments
2. please follow uncrustify part of https://github.com/mavlink/mavros/blob/master/CONTRIBUTING.md

mavros_extras/src/plugins/play_tune.cpp
mortenfyhn: :+1:

mavros_extras/src/plugins/play_tune.cpp
mortenfyhn: Thanks, that's much better. `strncpy` works, but it sure feels bug prone.

mavros_extras/src/plugins/play_tune.cpp
mortenfyhn: Yeah, it felt a bit redundant, but at the same time, repeating the enums in the message definitions means that the message sender doesn't have to depend on mavlink, just on mavros_msgs. Is that what you meant, have the sender do `msg.format = mavlink::common::TUNE_FORMAT::QBASIC1_1` and similar?

mortenfyhn: Looking at some other messages, I think I understand now. I'll update it to assign the format field directly from the mavros_msgs/PlayTuneV2 to the mavlink message. As long as the enum definition in the the mavros message matches the enum in mavlink, it's all good.

PR #1484, opened 2020-09-14
fake_gps: Fix assignment typo

mortenfyhn:
This colon should probably be an equals sign.
With the colon, this assignment becomes a label instead,
and `_gps_rate` after the colon becomes an unused
expression result.

Glancing at the code, I suppose the result of this bug is that the default gps rate of 5 Hz was used always, even if the "gps_rate" node param was set to something else.


vooon (approved):
LGTM, Thanks!

PR #1496, opened 2020-10-07
Fix warnings

mortenfyhn:
Hi, I tried building with `-Wall -Wextra -Wpedantic` and resolving the warnings that pop up. I have some clone of mavros in my workspace most of the time, so it's nice to get rid of the them. Most of these should be fairly uncontroversial, but the ECEF TF stuff should probably get some scrutiny.


vooon (approved):
LGTM, Thanks!

mortenfyhn:
Thanks for the merge! I might open another similar PR later, I think I might have missed some warnings.

TSC21:
Thanks @mortenfyhn!

PR #1499, opened 2020-10-15
mavros_msgs/State: Add PX4 flight mode constants

mortenfyhn:
Hi yall! What do you think about this? Can also add some ardupilot constants if you want. It makes it easier to work with the `mode` field in the State message.


vooon:
That looking good to me. But we also need APM modes as well.

mortenfyhn:
You got it!

vooon (approved):
LGTM, Thanks!

PR #1501, opened 2020-10-21
mavros_msgs/State: Fix PX4 flight mode constants

mortenfyhn:
I messed up in #1499:

Turns out ROS message string literals don't need quotes,
so adding quotes creates strings including the quotes.

This fix removes all the quotes, so you get `MANUAL`, not `"MANUAL"`.
(The generated strings in the C++ message header was `"\"MANUAL\""`.)


amilcarlucas (approved):

vooon (approved):
LGTM, Thanks!

PR #1507, opened 2020-11-09
Report custom firmware version

mortenfyhn:
Hi! I thought it might be useful to some to include the custom firmware version info in the vehicle info. This adds a string type field `flight_custom_version` to the VehicleInfo message, which corresponds to the field of the same name in MAVLink [AUTOPILOT_VERSION](https://mavlink.io/en/messages/common.html#AUTOPILOT_VERSION). Then it takes the custom version from the autopilot, which is an array of 8 bytes, converts it to hex string with the existing method, and sends it.


vooon (approved):
LGTM, Thanks!

PR #1512, opened 2020-11-25
A service for getting all autopilot parameters at once

mortenfyhn:
Hello again my friends,

I wanted a way to get the complete list of autopilot parameters, so I added a service server "get_all" to the param plugin. It returns unsuccessful if the param plugin doesn't have autopilot parameters yet. Otherwise it returns the full list of parameters.

I've also marked some actually constant methods as const, so that I can call `Parameter::to_msg` on a const instance of `Parameter`.

### DIFF
--- a/mavros/src/plugins/param.cpp
+++ b/mavros/src/plugins/param.cpp
@@ -223,7 +223,7 @@ public:
 	}
 
 	//! Make PARAM_SET message. Set target ids manually!
-	PARAM_SET to_param_set_apm_qurk()
+	PARAM_SET to_param_set_apm_qurk() const
 	{
 		PARAM_SET ret{};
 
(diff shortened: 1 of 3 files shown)


mortenfyhn:
Testing on hardware now, seems to not necessarily return _all_ params. Hang on.

mortenfyhn:
I believe it's solved now.

mortenfyhn:
I messed one thing up, but fixed it and rebased. (Made `to_param_set` const like the doofus I am.)

vooon:
Hello, i'm not sure that this is really needed because you could simply get rosparam get `/mavros/param/`

https://github.com/mavlink/mavros/blob/master/mavros/src/mavros/param.py#L193
https://github.com/mavlink/mavros/blob/master/mavros/scripts/mavparam#L48

mortenfyhn:
I didn't realize! I'll try that out tomorrow. The best code is no code so I'll be happy if this turns out to be unnecessary.

mortenfyhn:
Okay, so I first call the `mavros/param/pull` service, that populates all the parameters to the ROS parameter server, and then I can get them all with `rosparam get `/mavros/param`. Works!

Only one question then: Is there a way to verify that I have grabbed exactly all parameters, or get the "param_count" reported over MAVLink?

vooon:
Actually it's not necessary to call `pull`, rosparam should be updated by rx handler. (Update: no, not anymore, you have to call service).

That not possible to check that you have all params, but you can check success field, it should be flagged in case of timeout error.
And logic of the plugin checks for param_count and tries to re-request them.
So if you absolutely need to verify parameters - probably best possibility is to use force pull and check for status.

Also it's possible to catch message on param update topic.

mortenfyhn:
I understand. I'll experiment with that, and see if I'm able to do what I want without the new service. Closing this then. Thank you!

PR #1537, opened 2021-02-13
fixed yaw problems in setponit_raw and setponit_position when using BODY_NED and BODY_OFFSET_NED

zhouzhiwen2000:
When sending yaw in BODY_NED frame using setponit_raw,  not only is there a 90° offset but the direction is also reversed.
When sending yaw in BODY_NED frame using setponit_position,  the direction is reversed.
The yaw is supposed to follow the 'FLU' convention in ROS, thus, a positive angle would be from F to L(anticlockwise).
The changes made should fix both problems.
They were both tested and worked as expected.

### DIFF
--- a/mavros/include/mavros/frame_tf.h
+++ b/mavros/include/mavros/frame_tf.h
@@ -62,6 +62,8 @@ enum class StaticTF {
 	ENU_TO_NED,		//!< change from expressed WRT ENU frame to WRT NED frame
 	AIRCRAFT_TO_BASELINK,	//!< change from expressed WRT aircraft frame to WRT to baselink frame
 	BASELINK_TO_AIRCRAFT,	//!< change from expressed WRT baselnk to WRT aircraft
+	ABSOLUTE_FRAME_AIRCRAFT_TO_BASELINK,//!< change orientation from being expressed in aircraft frame to baselink frame in an absolute frame of reference.
+	ABSOLUTE_FRAME_BASELINK_TO_AIRCRAFT,//!< change orientation from being expressed in baselink frame to aircraft frame in an absolute frame of reference
 };
 
 /**
@@ -184,6 +186,26 @@ inline T transform_orientation_aircraft_baselink(const T &in) {
 template<class T>
 inline T transform_orientation_baselink_aircraft(const T &in) {
 	return detail::transform_orientation(in, StaticTF::BASELINK_TO_AIRCRAFT);
+}
+
+/**
+ * @brief Transform from attitude represented WRT aircraft frame to
+ *		  attitude represented WRT base_link frame, treating aircraft frame 
+ *		  as in an absolute frame of reference (local NED).
+ */
+template<class T>
+inline T transform_orientation_absolute_frame_aircraft_baselink(const T &in) {
+	return detail::transform_orientation(in, StaticTF::ABSOLUTE_FRAME_AIRCRAFT_TO_BASELINK);
+}
+
+/**
+ * @brief Transform from attitude represented WRT baselink frame to
+ *		  attitude represented WRT body frame, treating baselink frame 
+ *		  as in an absolute frame of reference (local NED).
+ */
+template<class T>
+inline T transform_orientation_absolute_frame_baselink_aircraft(const T &in) {
+	return detail::transform_orientation(in, StaticTF::ABSOLUTE_FRAME_BASELINK_TO_AIRCRAFT);
 }
 
 /**
--- a/mavros/src/plugins/setpoint_position.cpp
+++ b/mavros/src/plugins/setpoint_position.cpp
@@ -146,7 +146,7 @@ private:
 
 		auto q = [&] () {
 				if (mav_frame == MAV_FRAME::BODY_NED || mav_frame == MAV_FRAME::BODY_OFFSET_NED) {
-					return ftf::transform_orientation_baselink_aircraft(Eigen::Quaterniond(tr.rotation()));
+					return ftf::transform_orientation_absolute_frame_aircraft_baselink(Eigen::Quaterniond(tr.rotation()));
 				} else {
 					return ftf::transform_orientation_enu_ned(
 						ftf::transform_orientation_baselink_aircraft(Eigen::Quaterniond(tr.rotation())));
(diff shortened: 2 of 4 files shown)


mavros/src/plugins/setpoint_position.cpp
```diff
@@ -146,7 +146,7 @@ class SetpointPositionPlugin : public plugin::PluginBase,
 
 		auto q = [&] () {
 				if (mav_frame == MAV_FRAME::BODY_NED || mav_frame == MAV_FRAME::BODY_OFFSET_NED) {
-					return ftf::transform_orientation_baselink_aircraft(Eigen::Quaterniond(tr.rotation()));
+					return ftf::quaternion_from_rpy(M_PI, 0.0, 0.0)*Eigen::Quaterniond(tr.rotation());
```
TSC21, line 149: This is exactly the same rotation as the one in https://github.com/mavlink/mavros/blob/master/mavros/include/mavros/frame_tf.h#L185. What exactly are you solving here?

mavros/src/plugins/setpoint_raw.cpp
```diff
@@ -166,16 +166,21 @@ class SetpointRawPlugin : public plugin::PluginBase,
...
-		}
-
-		yaw = ftf::quaternion_get_yaw(
+			yaw = ftf::quaternion_get_yaw(
 					ftf::transform_orientation_aircraft_baselink(
 						ftf::transform_orientation_ned_enu(
 							ftf::quaternion_from_rpy(0.0, 0.0, req->yaw))));
+		}
+
+ 		ROS_INFO("FROM MAVROS,YAW=%f,YAW_ORG=%f",ftf::quaternion_get_yaw(
```
TSC21, line 182: Please remove the debug message.

zhouzhiwen2000: debug message removed.
the multiplication order is reversed.

mavros/src/plugins/setpoint_raw.cpp
```diff
@@ -166,16 +166,19 @@ class SetpointRawPlugin : public plugin::PluginBase,
 			position = ftf::transform_frame_baselink_aircraft(position);
 			velocity = ftf::transform_frame_baselink_aircraft(velocity);
 			af = ftf::transform_frame_baselink_aircraft(af);
+			yaw = ftf::quaternion_get_yaw(
+					ftf::quaternion_from_rpy(M_PI, 0.0, 0.0)*(
```
TSC21, line 170: Please use the `ftf::transform_orientation_aircraft_baselink` instead.

zhouzhiwen2000: With transform_orientation_aircraft_baselink, that is, the original multiplication order, the roll is flipped after the rotation, making no effect on yaw.

TSC21: Oh I see. So can you instead create a new method in frame_tf to handle this transform?

zhouzhiwen2000: on it.

mavros/include/mavros/frame_tf.h
```diff
@@ -186,6 +188,26 @@ inline T transform_orientation_baselink_aircraft(const T &in) {
 	return detail::transform_orientation(in, StaticTF::BASELINK_TO_AIRCRAFT);
 }
 
+/**
+ * @brief Transform from attitude represented WRT aircraft frame to
+ *		  attitude represented WRT base_link frame, treating aircraft frame 
+ *		  as if it were fixed frame as opposed to the function above.
```
TSC21, line 194: What do you mean with "fixed" frame?

mavros/src/plugins/setpoint_raw.cpp
TSC21: You are still not applying the new funtions

mavros/include/mavros/frame_tf.h
TSC21: Do you mean inertial frame instead?

zhouzhiwen2000: Sorry, I meant "absolute" from http://www.euclideanspace.com/maths/algebra/realNormedAlgebra/quaternions/transforms/

"absolute frame of reference"
changing

TSC21: Right makes sense. Can you please then change the fixed/FIXED refetences to absolute/ABSOLUTE?

mavros/include/mavros/frame_tf.h
```diff
@@ -191,21 +191,21 @@ inline T transform_orientation_baselink_aircraft(const T &in) {
 /**
  * @brief Transform from attitude represented WRT aircraft frame to
  *		  attitude represented WRT base_link frame, treating aircraft frame 
- *		  as if it were fixed frame as opposed to the function above.
+ *		  as if it were absolute frame as opposed to the function above.
```
TSC21, line 194:
```suggestion
 *		  as in an absolute frame of reference (local NED).
```

mavros/include/mavros/frame_tf.h
```diff
@@ -186,6 +188,26 @@ inline T transform_orientation_baselink_aircraft(const T &in) {
...
+ */
+template<class T>
+inline T transform_orientation_absolute_frame_aircraft_baselink(const T &in) {
+	return detail::transform_orientation(in, StaticTF::ABSOLUTE_FRAME_AIRCRAFT_TO_BASELINK);
+}
+
+/**
+ * @brief Transform from attitude represented WRT baselink frame to
+ *		  attitude represented WRT body frame, treating baselink frame 
+ *		  as if it were absolute frame.
```
TSC21, line 204:
```suggestion
 *		  as in an absolute frame of reference (local NED).
```

mavros/include/mavros/frame_tf.h
```diff
@@ -62,6 +62,8 @@ enum class StaticTF {
 	ENU_TO_NED,		//!< change from expressed WRT ENU frame to WRT NED frame
 	AIRCRAFT_TO_BASELINK,	//!< change from expressed WRT aircraft frame to WRT to baselink frame
 	BASELINK_TO_AIRCRAFT,	//!< change from expressed WRT baselnk to WRT aircraft
+	ABSOLUTE_FRAME_AIRCRAFT_TO_BASELINK,
```
TSC21, line 65: Can you please add thr description comment in these two, as in the others on top? Thanks!

TSC21 (changes requested):

TSC21:
@zhouzhiwen2000 please also squash your commits into one or two logically different.

TSC21 (approved):
LGTM!

vooon (approved):
LGTM, thanks!

PR #1571, opened 2021-04-28
New servo output raw plugin

jbmag:
Hi! I needed the servo_output_raw message from PX4 (https://mavlink.io/en/messages/common.html#SERVO_OUTPUT_RAW) in a project, so I added a plugin to publish it to ROS. I used to work using my fork, but it would be simpler if it was added to master, if it's possible.


vooon:
RC_IO already have that. Why it doesn't suit you?

https://github.com/mavlink/mavros/blob/5ba8ec36ddf884505addb244310cedf0109f89de/mavros/src/plugins/rc_io.cpp#L162

jbmag:
Ahh how did I miss this!! that's exactly what I need, sorry and thank you, I'll close the PR!

PR #1594, opened 2021-07-20
Add compass calibration feedback status. Add service to call the 'Next' button in calibrations.

BV-OpenSource:


mavros/src/plugins/mag_calibration_status.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+/**
+ * @brief MagCalStatus plugin.
+ *
+ * Example and "how to" for users.
+ */
+class MagCalStatusPlugin : public plugin::PluginBase {
+public:
+	MagCalStatusPlugin() : PluginBase(),
+		mcs_nh("~MagCalibration")
```
vooon, line 25: Mustn't use camel case for node handles.

mavros/src/plugins/mag_calibration_status.cpp
```diff
@@ -0,0 +1,100 @@
...
+			make_handler(&MagCalStatusPlugin::handle_report),
+		};
+	}
+
+private:
+	ros::NodeHandle mcs_nh;
+	ros::Publisher mcs_pub;
+	ros::Publisher mcr_pub;
+	uint8_t _rgCompassCalProgress[3] = {0};
+	bool calibration[3]={true, true, true};
```
vooon, line 59: Shouldn't use C-style arrays. Use `std::array` or `std::vector`!

mavros/src/plugins/mag_calibration_status.cpp
```diff
@@ -0,0 +1,100 @@
...
+	void handle_status(const mavlink::mavlink_message_t *, mavlink::ardupilotmega::msg::MAG_CAL_PROGRESS &mp) {
+		auto mcs = boost::make_shared<std_msgs::UInt8>();
+
+		// How many compasses are we calibrating?
+		int compassCalCount = 0;
+		for (int i=0; i<3; i++) {
+			if (mp.cal_mask & (1 << i)) {
+				compassCalCount++;
+			}
+		}
```
vooon, line 70: https://en.cppreference.com/w/cpp/utility/bitset/count

vooon: Also please do not use camel case for variables.

mavros/src/plugins/mag_calibration_status.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+		mcs->data = static_cast<uint8_t>((_rgCompassCalProgress[0] + _rgCompassCalProgress[1] + _rgCompassCalProgress[2])/compassCalCount);
+
+		mcs_pub.publish(mcs);
+	}
+
+	//Send report after calibration is done
+	void handle_report(const mavlink::mavlink_message_t *, mavlink::common::msg::MAG_CAL_REPORT &mr) {
+		if(calibration[mr.compass_id]) {
```
vooon, line 84: Where does it set?

vooon (changes requested):
First: that is not essential plugin, so it must go to extras.
Second: did you run uncrustify as described in contribution guide?

mavros/launch/apm.launch
```diff
@@ -3,7 +3,7 @@
 	<!-- example launch script for ArduPilot based FCU's -->
 
 	<arg name="fcu_url" default="/dev/ttyACM0:57600" />
-	<arg name="gcs_url" default="" />
+	<arg name="gcs_url" default="udp://@:14555" />
```
vooon, line 6: Please do not include changes to launch files.

mavros_extras/src/plugins/mag_calibration_status.cpp
```diff
@@ -0,0 +1,102 @@
...
+		std::bitset<8> compass_calibrating = mp.cal_mask;
+
+		if (_rg_compass_cal_progress.size() < 3) {
+			_rg_compass_cal_progress.resize(3);
+		}
+
+		if (calibration_show.size() < 3) {
+			calibration_show.resize(3);
+			std::fill(calibration_show.begin(), calibration_show.end(), true);
+		}
```
vooon, line 73: What would reset calibration_show?

mavros_extras/src/plugins/mag_calibration_status.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+		mcs->data = static_cast<uint8_t>((_rg_compass_cal_progress[0] + _rg_compass_cal_progress[1] + _rg_compass_cal_progress[2]) / compass_calibrating.count());
+
+		mcs_pub.publish(mcs);
+	}
+
+	//Send report after calibration is done
+	void handle_report(const mavlink::mavlink_message_t *, mavlink::common::msg::MAG_CAL_REPORT &mr) {
+		if (calibration_show[mr.compass_id]) {
```
vooon, line 87: Why do you need to filter reports?

And why not to make a message for progress?

vooon (changes requested):

mavros_extras/src/plugins/mag_calibration_status.cpp
BV-OpenSource: In order to publish only one time for each compass.

Since the movements requested for the calibration are independent of the compass, we think that the 0 to 100% value is all the information needed.

mavros_extras/src/plugins/mag_calibration_status.cpp
BV-OpenSource: A new calibration request. Would you do it different?

vooon: But where request comes from?
What would clear show flags? Because now i only see that only first time it comes it'll do initialization.

I also suggest to use array<> and define max size const.

vooon (approved):
Ok, LGTM now! Thanks!

PR #1600, opened 2021-07-29
Added support for ESC_TELEMETRY_N_TO_M message

bromeara:
Added a plugin to handle the ArduMega dialect messages of esc_telemetry_N_to_M.
Has been tested on a jetson tx2 connected to an orange cube.

### DIFF
--- a/mavros_extras/src/plugins/esc_telemetry.cpp
+++ b/mavros_extras/src/plugins/esc_telemetry.cpp
@@ -29,9 +29,9 @@ public:
     Subscriptions get_subscriptions() override
     {
         return {
-            make_handler(&ESCStatusPlugin::handle_esc_telemetry_1_to_4),
-            make_handler(&ESCStatusPlugin::handle_esc_telemetry_5_to_8),
-            make_handler(&ESCStatusPlugin::handle_esc_telemetry_9_to_12),
+            make_handler(&ESCTelemetryPlugin::handle_esc_telemetry_1_to_4),
+            make_handler(&ESCTelemetryPlugin::handle_esc_telemetry_5_to_8),
+            make_handler(&ESCTelemetryPlugin::handle_esc_telemetry_9_to_12),
         };
     }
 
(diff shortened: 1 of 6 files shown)


mavros_extras/src/plugins/esc_telemetry.cpp
```diff
@@ -0,0 +1,122 @@
...
+ */
+class ESCTelemetryPlugin : public plugin::PluginBase
+{
+public:
+    ESCTelemetryPlugin() : PluginBase(),
+        nh("~")
+    {}
+
+    void initialize(UAS &uas_) override
+    {
```
vooon, line 20: Please apply uncrustify as in contribution guide: https://github.com/mavlink/mavros/blob/master/CONTRIBUTING.md

mavros_extras/src/plugins/esc_telemetry.cpp
```diff
@@ -0,0 +1,122 @@
...
+        PluginBase::initialize(uas_);
+
+        esc_telemetry_1_to_4_pub = nh.advertise<mavros_msgs::ESCTelemetry>("esc_telemetry_1_to_4", 10);
+        esc_telemetry_5_to_8_pub = nh.advertise<mavros_msgs::ESCTelemetry>("esc_telemetry_5_to_8", 10);
+        esc_telemetry_9_to_12_pub = nh.advertise<mavros_msgs::ESCTelemetry>("esc_telemetry_9_to_12", 10);
```
vooon, line 25: Why not to combine all ESCs data into one topic? Do something similar to rc_io.

mavros_extras/src/plugins/esc_telemetry.cpp
```diff
@@ -0,0 +1,122 @@
...
+            _esc_telemetry_1_to_4.esc_telemetry[i].voltage = esc_telemetry.voltage[i];
+            _esc_telemetry_1_to_4.esc_telemetry[i].current = esc_telemetry.current[i];
+            _esc_telemetry_1_to_4.esc_telemetry[i].totalcurrent = esc_telemetry.totalcurrent[i];
+            _esc_telemetry_1_to_4.esc_telemetry[i].rpm = esc_telemetry.rpm[i];
+            _esc_telemetry_1_to_4.esc_telemetry[i].count = esc_telemetry.count[i];
+        }
+        
+        esc_telemetry_1_to_4_pub.publish(_esc_telemetry_1_to_4);
+
+    }
```
vooon, line 74: I'd made a template function which takes esc_telemetry and index offset.

mavros_msgs/msg/ESCTelemetryItem.msg
```diff
@@ -0,0 +1,6 @@
+uint8 temperature
```
vooon, line 1: Need to add Header and fillup receiving time if message do not have it.

Thus way you could check what items been updated.

vooon (changes requested):
Overall looks good, except that i'd prefer to see single topic with all data and with timestamps.

Also maybe it's better to extend esc_status, why not?

PR #1603, opened 2021-08-17
ros2: sanitized code

dayjaby:
Doing a colcon build resulted in lots of warnings. This gets rid of all of them, which should help with debugging the real errors :)

@vooon FYI Going to work on integrating the landing_target plugin into ros2 by following your way of doing it in e.g. https://github.com/mavlink/mavros/commit/e3d3a5963f400328d59d0fd2e5e7aa7c065b8d7a

Issue #1601, opened 2021-08-04
Topics can echo data with "ros2 topic echo" but not with "rclpy or rclcpp"


hexray-newbee:
### Issue details

`ros2 topic echo /mavros/mavros/raw/fix` will echo NavSatFix.
But, the following sample code cannot sub data.
```python
# refer: https://docs.ros.org/en/foxy/Tutorials/Writing-A-Simple-Py-Publisher-And-Subscriber.html#write-the-subscriber-node
import rclpy
from rclpy.node import Node

from sensor_msgs.msg import NavSatFix


class MinimalSubscriber(Node):

    def __init__(self):
        super().__init__('minimal_subscriber')
        self.subscription = self.create_subscription(
            NavSatFix,
            '/mavros/mavros/raw/fix',
            self.listener_callback,
            10)
        self.subscription  # prevent unused variable warning

    def listener_callback(self, msg):
        self.get_logger().info('I heard: "%f"' % msg.altitude)


def main(args=None):
    rclpy.init(args=args)

    minimal_subscriber = MinimalSubscriber()

    rclpy.spin(minimal_subscriber)

    # Destroy the node explicitly
    # (optional - otherwise it will be done automatically
    # when the garbage collector destroys the node object)
    minimal_subscriber.destroy_node()
    rclpy.shutdown()


if __name__ == '__main__':
    main()
```

### MAVROS version and platform

Mavros: 2.0.3
ROS: Foxy
Ubuntu: 20.04.2 (docker pull ros:foxy)

### Autopilot type and version

[x] ArduPilot
[  ] PX4

Version: 4.1.0-beta3

### Node logs

```
[INFO] [launch]: All log files can be found below /root/.ros/log/2021-08-04-07-35-09-268724-0ed951b0de46-19250
[INFO] [launch]: Default logging verbosity is set to INFO
[INFO] [mavros_node-1]: process started with pid [19252]
[mavros_node-1] [INFO] [1628062509.325559163] [mavros]: Starting mavros_node container
[mavros_node-1] [INFO] [1628062509.325624314] [mavros]: FCU URL: udp://:12345@
[mavros_node-1] [INFO] [1628062509.325634029] [mavros]: GCS URL: 
[mavros_node-1] [INFO] [1628062509.325640487] [mavros]: UAS Prefix: /uas1
[mavros_node-1] [INFO] [1628062509.325646294] [mavros]: Starting mavros router node
[...]

PR #1605, opened 2021-08-24
mavros_extras: Ported vision_pose_estimate plugin for ROS2

Peter010103:
Currently work in progress, but would appreciate comments on the changes.

### DIFF
--- a/mavros_extras/mavros_plugins.xml
+++ b/mavros_extras/mavros_plugins.xml
@@ -21,5 +21,10 @@ to the FCU and/or receive landing target tracking data coming from the FCU.</des
     <description>@brief 3DR Radio plugin.
 @plugin tdr_radio</description>
   </class>
+
+  <class name="vision_pose" type="mavros::plugin::PluginFactoryTemplate&lt;mavros::extra_plugins::VisionPoseEstimatePlugin&gt;" base_class_type="mavros::plugin::PluginFactory">
+    <description>@brief Vision pose estimate plugin
+@plugin vision_pose</description>
+  </class>
 </library>
 <!-- [[[end]]] (checksum: 6f77ca93b20723eab0d9b2a4a8a32ef1) -->
--- a/mavros_extras/src/plugins/vision_pose_estimate.cpp
+++ b/mavros_extras/src/plugins/vision_pose_estimate.cpp
@@ -15,12 +15,21 @@
  * https://github.com/mavlink/mavros/tree/master/LICENSE.md
  */
 
-#include <mavros/mavros_plugin.h>
-#include <mavros/setpoint_mixin.h>
-#include <eigen_conversions/eigen_msg.h>
+#include "rcpputils/asserts.hpp"
+#include "mavros/mavros_uas.hpp"
+#include "mavros/utils.hpp"
+#include "mavros/plugin.hpp"
+#include "mavros/plugin_filter.hpp"
+#include "mavros/setpoint_mixin.hpp"
 
-#include <geometry_msgs/PoseStamped.h>
-#include <geometry_msgs/PoseWithCovarianceStamped.h>
+#include <tf2_eigen/tf2_eigen.h>
+
+#include "geometry_msgs/msg/pose_stamped.hpp"
+#include "geometry_msgs/msg/pose_with_covariance_stamped.hpp" //
+#include "geometry_msgs/msg/vector3_stamped.hpp"
+#include "mavros_msgs/msg/landing_target.hpp"
+
+
 
 namespace mavros
 {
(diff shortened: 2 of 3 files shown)


Peter010103:
For reference, I get two warnings during compilation related to line 197 in `vision_pose_estimate.cpp`

Here is the exact error message :)

mavros_extras/mavros_plugins.xml
```diff
@@ -21,5 +21,10 @@ to the FCU and/or receive landing target tracking data coming from the FCU.</des
     <description>@brief 3DR Radio plugin.
 @plugin tdr_radio</description>
   </class>
+
+  <class name="vision_pose" type="mavros::plugin::PluginFactoryTemplate&lt;mavros::extra_plugins::VisionPoseEstimatePlugin&gt;" base_class_type="mavros::plugin::PluginFactory">
+    <description>@brief Vision pose estimate plugin
+@plugin vision_pose</description>
+  </class>
```
vooon, line 28: That's generated code by cog. I'll may do that after merge.

mavros_extras/src/plugins/vision_pose_estimate.cpp
```diff
@@ -15,57 +15,82 @@
...
-          " -> " << tf_child_frame_id);
-      tf2_start("VisionPoseTF", &VisionPoseEstimatePlugin::transform_cb);
-    } else {
-      vision_sub = sp_nh.subscribe("pose", 10, &VisionPoseEstimatePlugin::vision_cb, this);
-      vision_cov_sub = sp_nh.subscribe(
-        "pose_cov", 10, &VisionPoseEstimatePlugin::vision_cov_cb,
-        this);
-    }
+    node_declate_and_watch_parameter(
+      "tf/listen", false, [&](const rclcpp::Parameter & p) {
```
vooon, line 62: Use dot to separate subelements, e.g. `tf.listen`.

mavros_extras/src/plugins/vision_pose_estimate.cpp
```diff
@@ -15,57 +15,82 @@
...
+            tf2_start("VisionPoseTF", &VisionPoseEstimatePlugin::transform_cb);
+          } else{
+            vision_sub = node->create_subscription<geometry_msgs::msg::PoseStamped>(
+              "~/pose", 10, std::bind(
+                &VisionPoseEstimatePlugin::vision_cb, this, _1));
+            vision_cov_sub = node->create_subscription<geometry_msgs::msg::PoseWithCovarianceStamped>(
+              "~/pose_cov", 10, std::bind(
+                &VisionPoseEstimatePlugin::vision_cov_cb, this, _1));
+          }
+        });
```
vooon, line 78: NOTE: place "switching" parameter after it's config parameters. Now you would try to start TF with unset frame ids.
Also it's good to make those parameters to be possible to change in operation (you'd have to reset all things to initial state, then init subs//pubs again).

mavros_extras/src/plugins/vision_pose_estimate.cpp
```diff
@@ -111,18 +134,22 @@ class VisionPoseEstimatePlugin : public plugin::PluginBase,
...
 
-    vp.usec = stamp.toNSec() / 1000;
+    vp.usec = stamp.nanoseconds() / 1000;
     // [[[cog:
     // for f in "xyz":
     //     cog.outl("vp.%s = position.%s();" % (f, f))
     // for a, b in zip("xyz", ('roll', 'pitch', 'yaw')):
     //     cog.outl("vp.%s = rpy.%s();" % (b, a))
     // ]]]
+
```
vooon, line 152: Please don't change cog-generated code.

vooon (commented):
Overall looks good. Warning in FTF - i think you could replace index var types with Eigen one.

Also look carefully to tf mixin, it segfaults for landing_target.

Peter010103:
Hey @vooon . I'm a little confused as to whether this plugin is actually working or not.

When I do `ros2 run mavros mavros_node --ros-args --remap __ns:=/fpvquad_0/mavros -p fcu_url:=/dev/serial1:921600` and then run `ros2 topic list`, I can definitely see and 'echo' the ROS2 topic `/fpvquad_0/mavros/vision_pose/pose`. I am running the mavros node on a Raspberry Pi onboard the drone, which is connected to the same domain as my PC where I provide and publish fake pose data to the above topic.

I wanted to check whether I am actually receiving these topics on the flight computer, so I tried 3 methods:
1. Log the data by adding the `vehicle_visual_odometry` uORB topic in `etc/logging/logger_topics.txt` on the SD card.
2. Go to MAVLink Console and check that the topic is listed in `/obj`
3. Go to MAVLink Console and do `uorb top`

For 1, when I check the .ulg log file (and convert it to csv) I don't see the `vehicle_visual_odometry` topic 
For 2, after running the mavros node I can see the `vehicle_visual_odometry` topic listed
For 3, I get the below message. (I assume the instance of 0 is saying the FC did not receive the mavlink messages)
Screenshot from 2021-08-31 12-57-34

Could you provide some pointers?

vooon:
@Peter010103 that's sounds similar to #1601. But first you probably need to check that plugin is loaded.
Note that you can intercept all messages on UAS link (i.e. `/uas1/mavlink_sink`).

Peter010103:
@vooon I could see the plugin was loaded and was able to see the messages by doing a ros2 topic echo of `/uas1/mavlink_sink` :)

As an extra measure, I connected the RX and TX of the Raspberry to a logic analyzer and verified that I was receiving MAVLink 2.0 messages with msg_id: 102 (the VISION_POSITION_ESTIMATE message)

mavros_extras/src/plugins/vision_pose_estimate.cpp
dayjaby: I think this was caused by my sanitize PR: https://github.com/mavlink/mavros/pull/1603
I didn't find a way to get rid of the warnings of the `\` in the end without creating that extra blank line. So all the blame on me :(

vooon (approved):
I think i'll merge that PR as is and then fix it.

PR #1606, opened 2021-08-25
Add Mount angles message for communications with ardupilotmega.

BV-OpenSource:


Jaeyoung-Lim (commented):
Is there a specific reason that this cannot be included in the `mount_control` plugin?

BV-OpenSource:
> Is there a specific reason that this cannot be included in the `mount_control` plugin?

Not at all. Simply didn't want to mix functionalities since `mount_control` plugin already has a callback to return position status but mavlink message was different.

mavros_msgs/msg/MountStatus.msg
```diff
@@ -0,0 +1,9 @@
+# Mount Status message
+
+std_msgs/Header header
+
+uint8 target_system
+uint8 target_component
+int32 pointing_a
+int32 pointing_b
+int32 pointing_c
```
vooon, line 9: https://mavlink.io/en/messages/ardupilotmega.html#MOUNT_STATUS

From that definition i do not see any reason to define that message. I believe [Vector3Stamped](http://docs.ros.org/en/noetic/api/geometry_msgs/html/msg/Vector3Stamped.html) is enough, isn't it?

mavros_extras/src/plugins/mount_status.cpp
```diff
@@ -0,0 +1,81 @@
...
+	MountStatusPlugin() : PluginBase(),
+		nh("~"),
+		mount_nh("~mount_status")
+	{ }
+
+	void initialize(UAS &uas_) override
+	{
+		PluginBase::initialize(uas_);
+
+		mount_status_pub = mount_nh.advertise<mavros_msgs::MountStatus>("orientation", 10);
```
vooon, line 40: Orientation usually means that you provide quaternion. I think it's bad name for something custom.

vooon (changes requested):
I think it should be part of mount_control. Why not?

mavros_extras/src/plugins/mount_control.cpp
```diff
@@ -81,6 +83,27 @@ class MountControlPlugin : public plugin::PluginBase {
...
+
+		publish_msg.header.stamp = ros::Time::now();
+
+		publish_msg.header.frame_id = std::to_string(ms.target_component);
+
+		publish_msg.vector.x = ms.pointing_a;
+		publish_msg.vector.y = ms.pointing_b;
+		publish_msg.vector.z = ms.pointing_c;
+
+		mount_status_pub.publish(publish_msg);
```
vooon, line 104: Maybe it's better to publish quaternion as well as rpy? Like in the code above.

BV-OpenSource: Do you have any suggestion for topic's name?

vooon: Orientation. II suppose that APM do not send MOUNT_ORIENTATION, right?
So it's ok to publish both euler angles and quaternion.

mavros_extras/src/plugins/mount_control.cpp
```diff
@@ -102,6 +102,12 @@ class MountControlPlugin : public plugin::PluginBase {
 		publish_msg.vector.z = ms.pointing_c;
```
vooon, line 102: Ah, it's in centidegrees. Then you need to normalise them. Something like that:
```
auto vec = Eigen::Vector3d(ms.pointing_b, ms.pointing_a, ms.pointing_c) / 100.0;
tf::toMsg(vec, publish_msg.vector);
```

vooon: Sorry, wrong conversion, should be: `* M_PI / 18000.0`, because vectors should be in radians.

BV-OpenSource: I didn't make any conversion since [mavlink message documentation](https://mavlink.io/en/messages/ardupilotmega.html#MOUNT_STATUS) says that it is centidegree.

mavros_extras/src/plugins/mount_control.cpp
```diff
@@ -97,46 +126,46 @@ class MountControlPlugin : public plugin::PluginBase {
...
+			res.success = client.call(cmd);
+		}
+		catch (ros::InvalidNameException &ex) {
+			ROS_ERROR_NAMED("mount", "MountConfigure: %s", ex.what());
+		}
+
+		ROS_ERROR_COND_NAMED(!res.success, "mount", "MountCongifure: command plugin service call failed!");
+
+		return res.success;
+	}
```
Jaeyoung-Lim, line 168: Are these whitespace fixes necessary?

Jaeyoung-Lim (changes requested):

mavros_extras/src/plugins/mount_control.cpp
BV-OpenSource: They were added by uncrustify.

mavros_extras/src/plugins/mount_control.cpp
vooon: Yes, and ROS convention to use radians almost everywhere, that is why conversion needed.

BV-OpenSource: The switch between `pointing_b` and `pointing_a` were on purpose?

vooon: No, i just copy-pasted them from somewhere. Should be in following order: roll / pitch / yaw.

BV-OpenSource: Ok, so b, a, c is correct since a is pitch and b is roll.

vooon (approved):
LGTM, Thanks!

Issue #1338, opened 2019-10-30
Disconnected TF Trees Possibly As A Result of External Position Estimation?


Fisherman3:
### Issue details

I am somewhat new to PX4, ROS, and MAVROS, so apologies in advance if this issue is just due to me doing something dumb. I am trying to use an up board (essentially a suped-up RPi that runs Ubuntu) for onboard control of a quadrotor drone, and I'm using MAVROS to interface with the Pixhawk 4 (pretty standard so far). However, I'm flying indoors where there is no GPS signal, so I'm trying to use Vicon for external position estimation, and will eventually switch to optical flow using a RealSense camera. From what I read online, for external position estimation I can use the vrpn-client-ros package to get a PoseStamped msg from Vicon and, assuming I've set up my coordinates correctly, pipe the resulting pose topic directly to the /mavros/mocap/pose topic, and if the parameters are properly set, PX4 will just use the messages piped to that topic as the drone's current position (i.e. identical messages should be sent on the /mavros/local_position/pose topic).

Here's where things are going wrong. First, using rostopic echo I can vertify that the vrpn-client-ros package is working correctly, but initially, when I piped the output into /mavros/mocap/pose using a launch file with the line
`<remap from="/mavros/mocap/pose" to="/vrpn_client_node/[name of object in Vicon]/pose"/>`
there was no output at all on /mavros/local_position/pose. I ran this a few times (I realized that I had initially had mistakes in my launch file, so I fixed those), and eventually /mavros/local_position/pose started putting out messages that were completely unrelated to the vrpn-client-ros messages. Simultaneously, mavros_node started printing red messages:
`[ERROR] [1572456228.195270712]: ODOM: Ex: Could not find a connection between 'local_origin_ned' and 'fcu' because they are not part of the same tree.Tf has two or more unconnected trees.
`
Now, whenever I run mavros_node, even when I'm not trying to pipe things in or subscribe to topics or anything, it prints that message. I have tried overwriting my version of MAVROS with the most recent version, updating my PX4 firmware, and changing all my parameters back to how they were, and the problem persists. (I haven't tried reinstalling ROS or my operating system, and I really hope I don't need to.) My TF tree looks like this:
![frames](https://user-images.githubusercontent.com/23280007/67884604-7c06fe80-fb1c-11e9-9c7e-deb78afad7ac.png)
[...]

Issue #1383, opened 2020-02-13
RC_CHANNELS_OVERRIDE on mavlink 2.0 does not handle all channels


karthikdesai:
### Issue details

Mavlink 2.0 for RC_CHANNELS_OVERRIDE msg (https://mavlink.io/en/messages/common.html#RC_CHANNELS_OVERRIDE) handles upto 18 channels.
The rc_io.cpp plugin for RC override subscriber accepts only 8 channels. The plugin has to check if mavlink 2.0 is enabled, and then accept other channels.
I could have done a PR to fix this, but i am not sure how to check which mavlink version is currently active (within the plugin context)
https://github.com/mavlink/mavros/blob/master/mavros/src/plugins/rc_io.cpp#L229

### MAVROS version and platform

Mavros: HEAD
ROS: Melodic
Ubuntu: 18.04

### Autopilot type and version

[ * ] ArduPilot
[  ] PX4

Version: 4.0.0-rc3

vooon:
Yes, you just need to update cog.py script and message definition.
You do not have to check for mavlink version, if link is in v1 mode it just won't use extension fields.

karthikdesai:
Ah okay thanks for the suggestion.
cog.py is new to me, makes sense to use it for msg generation !

vooon:
Fixed by #1616.

PR #1616, opened 2021-09-06
Mavlink v2.0 specs for RC_CHANNELS_OVERRIDE accepts upto 18 channels.…

amilcarlucas:
… The plugin publishes channels 9 to 18 if the FCU protocol version is 2.0

### DIFF
--- a/mavros/src/plugins/rc_io.cpp
+++ b/mavros/src/plugins/rc_io.cpp
@@ -247,6 +247,18 @@ private:
 		ovr.chan6_raw = req->channels[5];
 		ovr.chan7_raw = req->channels[6];
 		ovr.chan8_raw = req->channels[7];
+		if (m_uas->fcu_link->get_protocol_version() == mavconn::Protocol::V20) {
+			ovr.chan9_raw = req->channels[8];
+			ovr.chan10_raw = req->channels[9];
+			ovr.chan11_raw = req->channels[10];
+			ovr.chan12_raw = req->channels[11];
+			ovr.chan13_raw = req->channels[12];
+			ovr.chan14_raw = req->channels[13];
+			ovr.chan15_raw = req->channels[14];
+			ovr.chan16_raw = req->channels[15];
+			ovr.chan17_raw = req->channels[16];
+			ovr.chan18_raw = req->channels[17];
+		}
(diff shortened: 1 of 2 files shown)


mavros/src/plugins/rc_io.cpp
```diff
@@ -247,6 +247,18 @@ class RCIOPlugin : public plugin::PluginBase {
...
+			ovr.chan10_raw = req->channels[9];
+			ovr.chan11_raw = req->channels[10];
+			ovr.chan12_raw = req->channels[11];
+			ovr.chan13_raw = req->channels[12];
+			ovr.chan14_raw = req->channels[13];
+			ovr.chan15_raw = req->channels[14];
+			ovr.chan16_raw = req->channels[15];
+			ovr.chan17_raw = req->channels[16];
+			ovr.chan18_raw = req->channels[17];
+		}
```
vooon, line 261: 1. That code should be generated by cog.
2. With change in message type you'll switching from std::array to std::vector (or boost:: for ros1), so you must verify size of `channels`.

vooon (changes requested):
You must check vector size to prevent sigfaults.

I also do not like that old `chan%d_raw` style messages, because you always need to write a lot of loc.
Can we do something like that (pseudocode)
```
auto mapper = std::array<*uint16_t, 18>(&ovr.chan1_raw ...);
std::copy(req->channels.cbegin(), req->channels.cend(), mapper.begin());
```

vooon:
NOTE: you don't need to check that lint is in v2.0 mode, just fill in fields, mavlink library would drop them automatically.

amilcarlucas:
@karthikdesai can you please follow up on this?

karthikdesai:
@vooon 
cool thanks for the suggestion. i had actually opened up an issue for this and you had suggested me to use the cog script at the time. https://github.com/mavlink/mavros/issues/1383
The new commit @amilcarlucas pushed actually are the changes what i did after i followed on your comment on the issue ^^

With regards to the message type, i had defined the array size as 18 at the time so it is not a vector but an array itself, so the verification wasn't needed

vooon (approved):
LGTM, Thanks!

vooon:
Fix #1338.

PR #1620, opened 2021-09-06
Added Rally point plugin

amilcarlucas:


vooon:
What is the different to [rallypoint](https://github.com/mavlink/mavros/blob/master/mavros/src/plugins/rallypoint.cpp) plugin?

amilcarlucas:
Ooops we did not noticed that one exist already. We will rebase on top of the existing one.

vooon:
@amilcarlucas but does it really needed? I tested ros2 port of rallypoint on APM simulator, and it works good.

amilcarlucas:
@karthikdesai does the upstream plugin replace our own? Looks like ours has more options.

PR #1622, opened 2021-09-13
sys_time: publish /clock for simulation times

dayjaby:
I hope ROS1 Pull Requests are still ok :-)

When replaying PX4 log files plus rosbag files, we need a way to sync the times between PX4 ulg and the rosbag. This can be accomplished by the following procedure:

- Set `/mavros/time/publish_sim_time` to true
- increasing the rate of `SYSTEM_TIME` being sent to mavros, e.g. by `mavlink stream -u 14556 -s SYSTEM_TIME -r 100`
- Set `/use_sim_time` to true
- Run `rosbag play` with `--clock` argument

Alternative: Don't invent a new parameter, but rely on `/use_sim_time` instead.

### DIFF
--- a/mavros/src/plugins/sys_time.cpp
+++ b/mavros/src/plugins/sys_time.cpp
@@ -19,6 +19,7 @@
 #include <sensor_msgs/TimeReference.h>
 #include <std_msgs/Duration.h>
 #include <mavros_msgs/TimesyncStatus.h>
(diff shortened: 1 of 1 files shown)


vooon (approved):
LGTM, Thanks!

I think in that case extra parameter is preferable as it gives you more flexibility for debugging.

PR #1625, opened 2021-09-22
Plugin for TUNNEL messages

mortenfyhn:
I wanted to use MAVLink TUNNEL messages to communicate between ROS and some hardware connected to the FCU. I made a MAVROS plugin that forwards tunnels between ROS and the FCU, in both directions, and a Tunnel message in mavros_msgs that mirrors the MAVLink message.

One thing I'm not sure about is how to best deal with the variable length payload field. In MAVLink, it's defined as a 128 size array, but I suppose the serializer makes sure to avoid sending a bunch of zeros if only some of the size is used. I've made it a `uint8[128]` in ros to enforce the maximum size, but I think that might mean the ROS messages always send the full 128 bytes. I could make it `uint8[]`, but then I would need to handle the max size somewhere. (Maybe just a comment in the message definition, and a check in the plugin when it copies between the ROS and MAVLink messages?)


mavros_msgs/msg/Tunnel.msg
```diff
@@ -0,0 +1,22 @@
...
+uint16 MAV_TUNNEL_PAYLOAD_TYPE_STORM32_RESERVED0 = 200
+uint16 MAV_TUNNEL_PAYLOAD_TYPE_STORM32_RESERVED1 = 201
+uint16 MAV_TUNNEL_PAYLOAD_TYPE_STORM32_RESERVED2 = 202
+uint16 MAV_TUNNEL_PAYLOAD_TYPE_STORM32_RESERVED3 = 203
+uint16 MAV_TUNNEL_PAYLOAD_TYPE_STORM32_RESERVED4 = 204
+uint16 MAV_TUNNEL_PAYLOAD_TYPE_STORM32_RESERVED5 = 205
+uint16 MAV_TUNNEL_PAYLOAD_TYPE_STORM32_RESERVED6 = 206
+uint16 MAV_TUNNEL_PAYLOAD_TYPE_STORM32_RESERVED7 = 207
+uint16 MAV_TUNNEL_PAYLOAD_TYPE_STORM32_RESERVED8 = 208
+uint16 MAV_TUNNEL_PAYLOAD_TYPE_STORM32_RESERVED9 = 209
```
vooon, line 22: Perhaps `PAYLOAD_TYPE_` prefix is enough. Also i would rather to have it generated by cog from pymavlink.

mavros_msgs/msg/Tunnel.msg
```diff
@@ -0,0 +1,22 @@
+# Tunnel
+#
+# https://mavlink.io/en/messages/common.html#TUNNEL
+
+uint8 target_system
+uint8 target_component
+uint16 payload_type
+uint8 payload_length
+uint8[128] payload
```
vooon, line 9: For ROS2 it's possible to set upper limit. For ROS1 i suppose better to keep sending that unneeded payload tail.

mavros_msgs/msg/Tunnel.msg
mortenfyhn: Good point, I'll try to cog it.

mavros_msgs/msg/Tunnel.msg
mortenfyhn: Sounds fair. (I'm looking forward to migrating my stuff to ROS2 at some point...)

mavros_msgs/msg/Tunnel.msg
mortenfyhn: Do you know how to access the correct enum in pymavlink? I'm trying, based on ADSBVehicle.msg, to do something like
```py
from pymavlink.dialects.v20 import common
# use common.enums['MAV_TUNNEL_PAYLOAD_TYPE']
```
but I can't find any tunnel enums in common.enums:
```py
>>> [x for x in common.enums]
['MAV_COMPONENT', 'MAV_RESULT', 'MAV_ROI', 'UAVCAN_NODE_MODE', 'MAV_PROTOCOL_CAPABILITY', 'MAV_SENSOR_ORIENTATION', 'MOTOR_TEST_ORDER', 'CONDITION_TYPE', 'MAV_STATE', 'LANDING_TARGET_TYPE', 'MAV_MODE_FLAG', 'ADSB_ALTITUDE_TYPE', 'MAV_AUTOPILOT', 'MAV_BATTERY_TYPE', 'MOTOR_TEST_THROTTLE_TYPE', 'MAV_SYS_STATUS_SENSOR', 'ADSB_FLAGS', 'MAV_MISSION_RESULT', 'MAV_CMD_ACK', 'MAV_DO_REPOSITION_FLAGS', 'RTK_BASELINE_COORDINATE_SYSTEM', 'MAV_LANDED_STATE', 'MAV_PARAM_TYPE', 'MAV_COLLISION_THREAT_LEVEL', 'MAV_TYPE', 'MAV_ESTIMATOR_TYPE', 'CAMERA_MODE', 'MAV_COLLISION_ACTION', 'FENCE_ACTION', 'GPS_FIX_TYPE', 'MAV_VTOL_STATE', 'FENCE_BREACH', 'MAV_GOTO', 'MAV_BATTERY_CHARGE_STATE', 'MAV_BATTERY_FUNCTION', 'MAV_CMD', 'MAV_MISSION_TYPE', 'MAV_MODE_FLAG_DECODE_POSITION', 'GPS_INPUT_IGNORE_FLAGS', 'MAVLINK_DATA_STREAM_TYPE', 'MAV_FRAME', 'MAV_MODE', 'UAVCAN_NODE_HEALTH', 'MAV_SEVERITY', 'MAV_ARM_AUTH_DENIED_REASON', 'MAV_POWER_STATUS', 'ESTIMATOR_STATUS_FLAGS', 'SERIAL_CONTROL_FLAG', 'FIRMWARE_VERSION_TYPE', 'MAV_MOUNT_MODE', 'MAV_COLLISION_SRC', 'MAV_DATA_STREAM', 'VTOL_TRANSITION_HEADING', 'MAV_DISTANCE_SENSOR', 'SERIAL_CONTROL_DEV', 'ADSB_EMITTER_TYPE', 'RC_TYPE']
```

vooon: Maybe your pymavlink is older than the date when that message and enum came in?

mortenfyhn: You're right! I got it now.

mortenfyhn:
I've fixed the cog generation now, and I've added a check on the payload length, so I don't try to `std::copy` out of bounds data if `payload_length` is accidentally set higher than what's allowed.

mavros_extras/src/plugins/tunnel.cpp
```diff
@@ -30,26 +31,49 @@ class TunnelPlugin : public plugin::PluginBase
...
+                copy_tunnel<mavros_msgs::Tunnel, mavlink::common::msg::TUNNEL>(
+                    *ros_tunnel);
 
-        UAS_FCU(m_uas)->send_message_ignore_drop(mav_tunnel);
+            UAS_FCU(m_uas)->send_message_ignore_drop(mav_tunnel);
+        }
+        catch (const std::overflow_error& e)
+        {
+            ROS_ERROR_STREAM_NAMED("tunnel", e.what());
+        }
```
vooon, line 45: You wouldn't get overflow exception because `_ignore_drop` variation of `send_message()` catch it.

mavros_extras/src/plugins/tunnel.cpp
```diff
@@ -30,26 +31,49 @@ class TunnelPlugin : public plugin::PluginBase
...
+                copy_tunnel<mavlink::common::msg::TUNNEL, mavros_msgs::Tunnel>(
+                    mav_tunnel);
 
-        pub_.publish(ros_tunnel);
+            pub_.publish(ros_tunnel);
+        }
+        catch (const std::overflow_error& e)
+        {
+            ROS_ERROR_STREAM_NAMED("tunnel", e.what());
+        }
```
vooon, line 62: I do not see any reason to catch exceptions here.

mavros_extras/src/plugins/tunnel.cpp
```diff
@@ -30,26 +31,49 @@ class TunnelPlugin : public plugin::PluginBase
...
+    static To copy_tunnel(const From& from) noexcept(false)
     {
+        static constexpr auto max_payload_length =
+            sizeof(mavlink::common::msg::TUNNEL::payload) /
+            sizeof(mavlink::common::msg::TUNNEL::payload[0]);
+
+        if (from.payload_length > max_payload_length)
+        {
+            throw std::overflow_error("too long payload length");
+        }
```
vooon, line 75: I would rather print error log here and limit length to max size than throw an exception.

mavros_extras/src/plugins/tunnel.cpp
vooon: Ah, few lines below...

mavros_extras/src/plugins/tunnel.cpp
mortenfyhn: Are you sure? My thinking is that if you try to send something that isn't fully valid, such as specifying a payload length greater than the maximum length, we should avoid sending the potentially invalid message.

But the most important part is to not copy beyond 128 bytes, so we don't mess up our memory.

vooon: You could `sz = (from.payload_length > from.payload.size()) ? from.payload.size() : from.payload_lenght`.
But yes, you're right about filtering out invalid messages.

vooon (approved):
LGTM, Thanks!

vooon:
Hmm, i'm thinking, what if make Tunnel.payload dynamically sized, then we could remove Tunnel.payload_length (as we have it in payload.size()).

mortenfyhn:
That's absolutely an option.

So the error handling would then be:

* Plugin receives a mavros tunnel message
* Check payload.size(), print error and don't send if > 128

and

* Plugin receives a mavlink tunnel message
* Check payload_length, print error and don't send if > 128

The only downside I can see is that the mavros and mavlink messages would be less similar, so the user of the plugin needs to figure out that payload_length is just encoded in the array size and not in a separate field.

vooon:
I think first option better.

In my opinion payload_length makes handling of that message harder, and only needed for mavlink to express dynamically sized payload.
So i suppose that we can just add a note to the message and that's it.

mavros_extras/src/plugins/tunnel.cpp
```diff
@@ -0,0 +1,95 @@
...
+            ROS_ERROR_STREAM_NAMED("tunnel", e.what());
+        }
+    }
+
+    template <typename From, typename To>
+    static To copy_tunnel(const From& from) noexcept(false)
+    {
+        static constexpr auto max_payload_length =
+            sizeof(mavlink::common::msg::TUNNEL::payload) /
+            sizeof(mavlink::common::msg::TUNNEL::payload[0]);
```
vooon, line 70: Since payload is `std::array` i think better to use `::size()`.

PR #1626, opened 2021-09-22
Show ENOTCONN error instead of crash on socket's shutdown

valbok:
When a client suddenly drops the connection,
socket.shutdown() will throw an exception:
boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::system::system_error> >
  what():  shutdown: Transport endpoint is not connected

Showing an error in this common case looks more reasonable than crashing.


vooon (approved):
LGTM, Thanks!

Issue #1597, opened 2021-07-26
libmavconn: thread race in callback API


asherikov:
### Issue details

Callback API of `MAVConnInterface` class is not thread safe, which is pretty much obvious from the code -> https://github.com/mavlink/mavros/blob/ros2/libmavconn/include/mavconn/interface.hpp#L210, and is detected by thread sanitizers.

### MAVROS version and platform

I guess all.

### Comment

It is also impossible to work around the issue in user code -- connection is established in the constructor.

I see two potential fixes that don't break API:
- Add 'delay_connection' flag to the constructor and `connect()` method, so that users can register their callbacks before connection. This leaves some room for mistake, but IMO constructor must not connect anyway.
- Mark callback pointers as deprecated and introduce methods for setting them in a safe way with a mutex, calls to callback methods should be adjusted accordingly.

vooon:
Well, yes, that lib needs overhaul, but it's unlikely to have any significant changes in ROS1 code.

asherikov:
We (and many others) are going to stick to ROS 1 for quite a while. I might find some time to fix this, but which approach should be taken?

vooon:
I suppose that you can introduce connect method, so you could setup pointers after `open_url()`.
I'm not quite sure that you really need to have lock, since that pointers were changed once.

Issue #1478, opened 2020-09-09
Error bad_weak_ptr when using roslaunch after success build from source


tinng81:
### Issue details

After success [source installation](https://github.com/mavlink/mavros/blob/master/mavros/README.md#source-installation) of *mavros* and dependencies, I encountered error `bad_weak_ptr` when using [launch instruction](https://dev.px4.io/v1.9.0/en/simulation/ros_interface.html#launching-rossimulation) from PX4.

`roslaunch mavros px4.launch fcu_url:="udp://:14540@127.0.0.1:14557"`

Some other steps I have tried:
- `catkin clean` and re-build all packages.
    - Build success, same error

- Remove everything, re-install dependencies and build from [source installation](https://github.com/mavlink/mavros/blob/master/mavros/README.md#source-installation) again.
    - Build success, same error

- Instead of using `wstool update`, cloned both *mavlink* and *mavros* as submodules and update with `git submodule update --init --recursive`.
    - Build failed, force update *mavlink* with `git submodule update --remote -f mavlink`; afterwards, build success but still same error

### MAVROS version and platform

Mavros: 1.30.0
ROS: Melodic
Ubuntu: Bionic 18.04

### Autopilot type and version

[  ] ArduPilot
[x] PX4

Version: 1.11.0

### Node logs

```
NODES
  /
    mavros (mavros/mavros_node)

auto-starting new master
process[master]: started with pid [24300]
ROS_MASTER_URI=http://localhost:11311

setting /run_id to af45f612-f26e-11ea-ad5e-001c42db7e93
process[rosout-1]: started with pid [24311]
started core service [/rosout]
process[mavros-2]: started with pid [24319]
[ INFO] [1599636780.251461931]: FCU URL: udp://:14540@127.0.0.1:14557
[ INFO] [1599636780.261582909]: udp0: Bind address: 0.0.0.0:14540
[ INFO] [1599636780.261684315]: udp0: Remote address: 127.0.0.1:14557
[ INFO] [1599636780.263160901]: GCS bridge disabled
terminate called after throwing an instance of 'std::bad_weak_ptr'
  what():  bad_weak_ptr
================================================================================REQUIRED process [mavros-2] has died!
process has died [pid 24319, exit code -6, cmd /home/winng/Documents/Belvedere/catkin_ws/devel/lib/mavros/mavros_node __name:=mavros __log:=/home/winng/.ros/log/af45f612-f26e-11ea-ad5e-001c42db7e93/mavros-2.log].
[...]

Issue #937, opened 2018-02-03
Bug: terminate called after throwing an instance of 'std::bad_weak_ptr'


TSC21:
This is an intermitent error that appears to crash the `mavros_node`. More than once happened in the PX4 Jenkins CI (example: http://ci.px4.io:8080/blue/organizations/jenkins/Firmware/detail/PR-8804/1/pipeline/). 

A more deep search for the usage of `std::weak_ptr` leads me to think that the `std::bad_weak_ptr` exception is triggered because the `weak_link` for the `mavconn` `diagnostic_update` object that was deleted (based on the `std::bad_weak_ptr` exception definition, `std::bad_weak_ptr is the type of the object thrown as exceptions by the constructors of std::shared_ptr that take std::weak_ptr as the argument, when the std::weak_ptr refers to an already deleted object`). I'm pointing to a possible gap in https://github.com/mavlink/mavros/blob/master/mavros/include/mavros/mavlink_diag.h.

A possible solution will probably involve the exception catch and handling, besides adding a verification to `std::weak_ptr::expired` (http://en.cppreference.com/w/cpp/memory/weak_ptr/expired) in order to verify whether the managed object has already been deleted.

@dagar FYI.

vooon:
[std::weak_ptr::lock()](http://en.cppreference.com/w/cpp/memory/weak_ptr/lock) as you see already calls `::expire()`.
Problem maybe exists because of bug in stdcpp lib?

Message 
```
terminate called after throwing an instance of 'std::bad_weak_ptr'
what():  bad_weak_ptr
```

Says almost nothing.

TSC21:
Maybe run it with gdb would help?

vooon:
Maybe if you can reproduce.

TSC21:
This is almost unreproducible cause it is rare and intermitent under (what appears to be) the same test conditions. @dagar do we have a way of running a stress test for this using `gdb` on Jenkins?

1184:
@TSC21 hello,I have the same problem with you.How did you solve it?
I ues ros kinetic and gazebo7.When run "roslaunch rotors_gazebo fixed_wing_px4_sim.launch",the terminal will show me this messege:

terminate called after throwing an instance of 'std::bad_weak_ptr'
  what():  bad_weak_ptr
Aborted (core dumped)
[...]

Issue #1655, opened 2021-11-24
ROS2: Refactor mavconn::Interface to postpone callback init


vooon:
Current interface have a race between callback initialization and data io.
Need to make separate start method.

PR #1658, opened 2021-11-26
Fix multiple bugs

asherikov:
- fix bad_weak_ptr on connect and disconnect
- introduce new API to avoid thread race when assigning callbacks
- fix uninitialized variable in TCP client constructor which would
randomly block TCP server

This is an API breaking change: if client code creates connections using
make_shared<>() instead of open_url(), it is now necessary to call new
connect() method explicitly.

Related:                                                                                                                                                                                       
https://github.com/mavlink/mavros/issues/1597                                                                                                                                                  
https://github.com/mavlink/mavros/issues/937                                                                                                                                                   
https://github.com/mavlink/mavros/issues/1478

### DIFF
--- a/libmavconn/include/mavconn/interface.h
+++ b/libmavconn/include/mavconn/interface.h
@@ -119,6 +119,13 @@ public:
 	MAVConnInterface(uint8_t system_id = 1, uint8_t component_id = MAV_COMP_ID_UDP_BRIDGE);
 
 	/**
+	 * @brief Establish connection, automatically called by open_url()
+	 */
+	virtual void connect(
+			const ReceivedCb &cb_handle_message,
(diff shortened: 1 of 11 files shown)


vooon (approved):
LGTM, Thanks!

vooon:
For #1655 i need to port that changes.

vooon:
Seems that PR broke rx path. Version 1.12 do not connect.

asherikov:
If there are any issues just revert it and let me know, I'll dig deeper.

vooon:
@asherikov it's fixed in https://github.com/mavlink/mavros/commit/002c5f138e6df3753f90299c6b75c2683fa8680d .
Although node uses open_url(), interface looks like connect() wasn't called and receive job not being started.

PR #1660, opened 2021-11-29
Fix warnings

mortenfyhn:
I have some build warnings in the mavros and mavros_extras packages when I build locally with Clang, which I've attempted to fix.

I think some of these changes are uncontroversial, such as
* reordering constructor initializer lists to avoid "-Wreorder-ctor",
* initializing yaw and yaw_rate explicitly to zero in the setpoint_trajectory plugin, and
* fixing the probably incorrect gps_rate initialization.

There was also a warning where `MissionBase::initialize(ros::NodeHandle *_wp_nh)` upsets the compiler because it "hides the overloaded virtual function `PluginBase::initialize(UAS &uas)` because MissionBase inherits from PluginBase. I've avoided this by renaming `MissionBase::initialize` but you may want to solve it differently, or give it a different name.

After this, all that remains when I build are unused variable warnings, mostly about `COV_SIZE` which is only used in debug builds. I suppose this could be fixed with the casting to void trick `(void)COV_SIZE` or with `[maybe_unused](https://en.cppreference.com/w/cpp/language/attributes/maybe_unused)` (but that would require C++17). Locally I've silenced this with a few `#pragma GCC diagnostic ignored "-Wunused-variable"`.

### DIFF
--- a/mavros/include/mavros/mission_protocol_base.h
+++ b/mavros/include/mavros/mission_protocol_base.h
@@ -237,10 +237,10 @@ public:
 		wp_state(WP::IDLE),
 		wp_type(WP_TYPE::MISSION),
 		wp_count(0),
-		wp_retries(RETRIES_COUNT),
 		wp_cur_id(0),
 		wp_cur_active(0),
 		wp_set_active(0),
+		wp_retries(RETRIES_COUNT),
 		is_timedout(false),
 		do_pull_after_gcs(false),
 		enable_partial_push(false),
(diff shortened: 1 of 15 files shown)


vooon (approved):
LGTM. Personally i prefer `[[maybe_unused]]`, but for v2...

mortenfyhn:
> LGTM. Personally i prefer `[[maybe_unused]]`, but for v2...

I agree, `[[maybe_unused]]` is much nicer than the (void) "hack". (A well-established hack but a hack nontheless.)

Anyway thanks for the merge!

PR #1664, opened 2021-12-01
update Contributing.md

AnneWendt:


vooon (approved):
Hello,

Good change, but it more applies to mavros v1 (for ROS1).
Because for ROS2 wee must use `ament_uncrustify` and `colcon test`.
I'll update ros1 branch and do some edits for ros2.

PR #1669, opened 2021-12-06
plugin: setpoint_raw: move getParam to initializer

Hs293Go:
Currently `attitude_cb()` attempts to get the `thrust_scaling` parameter. [This is problematic as documented in this question](https://answers.ros.org/question/192223/max-frequency-to-look-up-parameters), and is exacerbated when `attitude_cb()` is called in a fast control loop. In which case, `getParam()` may return false spuriously and trip a fatal error. This has been observed when a control loop is run at 150Hz on a Jetson Nano.

This PR moves the `getParam()` call to initialize to avoid this problem, at the cost of removing the ability to dynamically change `thrust_scaling` (which I believe is unlikely to be important)


mavros/src/plugins/setpoint_raw.cpp
```diff
@@ -71,6 +77,8 @@ class SetpointRawPlugin : public plugin::PluginBase,
 	ros::Subscriber local_sub, global_sub, attitude_sub;
 	ros::Publisher target_local_pub, target_global_pub, target_attitude_pub;
 
+	std::unique_ptr<double> thrust_scaling_parameter;
```
vooon, line 80: I think that you do not need to dynamically allocate that double. You just need a sign of unset scaling, so NAN is good for that.

vooon (changes requested):
I'm ok with the change, but don't think that you really need pointer.

mavros/src/plugins/setpoint_raw.cpp
Hs293Go: I'm actually inclined to use `boost::optional`. It would be elegant and does away with dynamic allocs. ROS already depends on boost so the added dependency is not a big issue. What do you think?

vooon: No. Please avoid Boost.

vooon: Well, i checked that there are std::optional in C++17, so it's ok to use boost::optional for ROS1. But i would rather limit boost dependency.

Hs293Go: @vooon Well, I'd hate to let my design choices get in the way of this PR. Just switched to setting a -1.0 sentinel value for `thrust_scaling` and pushed.

vooon (approved):
LGTM, Thanks!

PR #1670, opened 2021-12-07
Fix spurious bug because class field was uninitialized.

windelbouwman:
This bug caused mavros to crash randomly. Is there maybe a check that can be enabled in g++ to catch these kind of bugs?

Rationale:
For basic types, fields get initialized with there default value, but struct types get garbled memory, unless passed an initializer list.

See also: https://en.cppreference.com/w/cpp/language/default_initialization


vooon (approved):
LGTM, Thanks!

PR #1672, opened 2021-12-09
Set time/publish_sim_time to false by default

okalachev:
Problem: parameter `time/publish_sim_time = true` pretty much breaks the simulation.

As I understand using `getParam` without checking the return value keeps `publish_sim_time` variable random. So `param` with the default value must be used here.

### DIFF
--- a/mavros/src/plugins/sys_time.cpp
+++ b/mavros/src/plugins/sys_time.cpp
@@ -217,7 +217,7 @@ public:
 		m_uas->set_timesync_mode(ts_mode);
 		ROS_INFO_STREAM_NAMED("time", "TM: Timesync mode: " << utils::to_string(ts_mode));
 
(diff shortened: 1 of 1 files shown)


mavros/src/plugins/sys_time.cpp
```diff
@@ -217,7 +217,7 @@ class SystemTimePlugin : public plugin::PluginBase {
 		m_uas->set_timesync_mode(ts_mode);
 		ROS_INFO_STREAM_NAMED("time", "TM: Timesync mode: " << utils::to_string(ts_mode));
 
-		nh.getParam("time/publish_sim_time", publish_sim_time);
+		nh.param("time/publish_sim_time", publish_sim_time, false);
```
vooon, line 220: Hmm, it should be initialized in class too. Will do.

vooon (approved):
LGTM, Thanks!

PR #1674, opened 2021-12-10
added publication of waypoint reached

Vicidel:
This PR adds the publication of the current (=target) waypoint in ROS
The use case is to know what is the waypoint where the drone is going. I cannot easily use the reached waypoint because of loops messing up indexes.

One thing I noticed is that the message is published at high frequency, not sure if I should only publish when changed?


Vicidel:
Some things:
- I can also do the same PR on ros2 if needed 
- I can add a throttling to limit the publication, or publish only when modified, not really sure what's the best case here
- I can add a new mavros_msgs definition for WaypointCurrent instead of reusing WaypointReached (would be the same definition)

vooon:
Why not to subscribe to `~mission/waypoints` and then just check `is_current` flag?

Vicidel:
Well indeed that would work... Looking at the code it's logical but I was too focused on needing "current_wp" that I didn't thinkn of it. I'm closing this PR, thanks!

DinaAHMIM:
Hello,

I am trying to use waypoint_reached as I need to switch mode after reaching a certain waypoint 
I also tried subscribing to waypoints and using the waypoint list that I sent to PX4 but nothing worked (The flags never pass to true) I don't think I'm using the right syntax.

Can you maybe show me how to use it properly? either method (I am new to ROS/mavros and PX4 and I'm doing this project for a job assignement after the interview)
Or is there a method to import the wp_reached msg from PX4 to my script? 
Thank you very much

PR #1679, opened 2021-12-17
libmavconn: fix deadlock when call close()

SylvainPastor:
MavconnTCPClient: When calling the `close()` function (by a different thread), a mutex is taken at
  the start of this function which closes the socket and waits the end of io_service thread.

  Closing the socket causes the `Operation aborted` error in `do_recv()` function called by
  io_service thread which in turn calls the `close()` function.

  This causes a 'deadlock'.

  fix: Reduce the scope of the mutex in the `close()` function so that it is released before
  waiting the end of io_service thread.


vooon (approved):
LGTM, Thanks!

PR #1680, opened 2021-12-17
Add extra MAV_FRAMES to waypoint message

AndersonRayner:
Adds extra MAV_FRAME defines to the waypoint message as per the [MAV_FRAME enum](https://mavlink.io/en/messages/common.html#MAV_FRAME).  This change makes code that uploads terrain relative missions (rather than home altitude relative) missions cleaner to implement.


vooon (approved):
LGTM, Thanks!

PR #1682, opened 2021-12-20
fix std::system_error when tcp interface loses connection

SylvainPastor:
When the tcp connection is lost (remote TCP server stopped), an `End of file` error is caught
  by the `io_thread` in the `do_recv()` that calls the `close()` function.

  In the `close()` function, we stop the io_service and wait the end of the io_thread which
  causes an `std::system_error exception` _(cause: Resource deadlock avoided)_.

```
  Error:   mavconn: tcp0: receive: End of file at line 250 in libmavconn/src/tcp.cpp
           terminate called after throwing an instance of 'std::system_error'
           what():  Resource deadlock avoided
           Aborted (core dumped)
```

  fix:
    - close() function: stop io_service if current thread id != io_thread id
    - ~MAVConnTCPClient(): stop io_service and io_thread if thread is running

### DIFF
--- a/libmavconn/include/mavconn/tcp.hpp
+++ b/libmavconn/include/mavconn/tcp.hpp
@@ -83,6 +83,7 @@ private:
   asio::io_service io_service;
   std::unique_ptr<asio::io_service::work> io_work;
   std::thread io_thread;
+  std::atomic<bool> is_running; //!< io_thread running
 
   asio::ip::tcp::socket socket;
   asio::ip::tcp::endpoint server_ep;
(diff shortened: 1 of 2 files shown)


SylvainPastor:
Personally, in the `close()` function,  I will not stop the io_service and io_thread. 

Especially in the case where the connection is created by the `MAVConnTCPServer` (io_service used is the server service).

vooon (approved):
LGTM, Thanks!

SylvainPastor:
I have 2 style errors, I will correct them

PR #1696, opened 2022-01-21
Disable startup_px4_usb_quirk in px4_config.yaml

okalachev:
I guess this feature is not needed any more. 

Moreover it breaks connection with the most recent PX4 build, as it autostarts either MAVLink or shell on USB based on traffic it sees.


vooon (approved):
LGTM, Thanks!

PR #1700, opened 2022-02-07
Fix obstacle distance

oroelipas:
Hi!

This PR contains two bugfix in the /mavros_extras/src/plugins/obstacle_distance.cpp file.

**The first one** corrects the way the reduction of size of the LaserScan is done. 
The code should redux the resolution of the array to adjust to the mavlink OBSTACLE_DISTANCE message (72 values) taking the minimum value as aggregation function of the interpolation.

The scale_factor was taken as integer, so the interpolation worked wrong.

This image shows the error and the fix:

interpolarion bugfix explained

**The second bugfix** assigns a value to increment_f in case the size of the input LaserScan is bigger than the mavlink OBSTACLE_DISTANCE message.
This was missing, so ardupilot was usgin the value of increment (and NOT increment_f), showing wrong objects.

### DIFF
--- a/mavros_extras/src/plugins/obstacle_distance.cpp
+++ b/mavros_extras/src/plugins/obstacle_distance.cpp
@@ -88,16 +88,14 @@ private:
 			obstacle.increment_f = increment_deg;
 		} else {
 			// all distances from sensor will not fit so we combine adjacent distances always taking the shortest distance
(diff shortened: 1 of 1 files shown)


vooon (approved):
LGTM, Thanks!

PR #1704, opened 2022-02-15
sys_status.cpp: Do not use battery1 voltage for all batteries.

amilcarlucas:
Do not use battery1 voltage as voltage for all other batteries (bugfix).
Support both cell and total voltages above 65V
Support up-to 14S batteries
If available, add cell voltage information to the battery diagnostic

Tested on a drone with 3 batteries, and multiple cells

### DIFF
--- a/mavros/src/plugins/sys_status.cpp
+++ b/mavros/src/plugins/sys_status.cpp
@@ -286,10 +286,10 @@ class BatteryStatusDiag : public diagnostic_updater::DiagnosticTask
 public:
 	BatteryStatusDiag(const std::string &name) :
 		diagnostic_updater::DiagnosticTask(name),
-		voltage(-1.0),
-		current(0.0),
-		remaining(0.0),
-		min_voltage(6)
+		voltage(-1.0f),
+		current(0.0f),
+		remaining(0.0f),
+		min_voltage(6.0f)
 	{ }
 
 	void set_min_voltage(float volt) {
@@ -304,11 +304,16 @@ public:
 		remaining = rem;
 	}
 
+	void setcell_v(const std::vector<float> voltages) {
+		std::lock_guard<std::mutex> lock(mutex);
+		cell_voltage = voltages;
+	}
+
 	void run(diagnostic_updater::DiagnosticStatusWrapper &stat)
 	{
 		std::lock_guard<std::mutex> lock(mutex);
 
-		if (voltage < 0)
+		if (voltage < 0.0f)
 			stat.summary(2, "No data");
 		else if (voltage < min_voltage)
 			stat.summary(1, "Low voltage");
@@ -317,7 +322,14 @@ public:
 
 		stat.addf("Voltage", "%.2f", voltage);
 		stat.addf("Current", "%.1f", current);
-		stat.addf("Remaining", "%.1f", remaining * 100);
+		stat.addf("Remaining", "%.1f", remaining * 100.0f);
+		const int nr_cells = cell_voltage.size();
+		if (nr_cells > 1)
+		{
+			for (int i = 1; i <= nr_cells ; ++i) {
+				stat.addf(utils::format("Cell %u", i), "%.2f", cell_voltage[i-1]);
+			}
+		}
 	}
 
 private:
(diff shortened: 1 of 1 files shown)


amilcarlucas:
duplicated

mavros/src/plugins/sys_status.cpp
```diff
@@ -920,16 +925,64 @@ class SystemStatusPlugin : public plugin::PluginBase
...
+			{
+				voltage_acc = 0.0f;
+			}
+			cell_voltage = voltage_acc + (v * 0.001f);	// 1 mV
+			batt_msg->cell_voltage.push_back(cell_voltage);
+			total_voltage += cell_voltage;
+			nr_cells++;
+		}
+		batt_msg->voltage = total_voltage;
+		batt_msg->cell_voltage.erase(batt_msg->cell_voltage.begin()+nr_cells, batt_msg->cell_voltage.end()); // erase the unused cells
```
vooon, line 970: That's unclear to me. Why do you need `nr_cells`? Why not to use `cell_voltage.size()`?
And why do you need to `erase()`? [`reserve()`][1] do not change `size()`, only `capacity()`.

[1]: https://en.cppreference.com/w/cpp/container/vector/reserve

vooon (changes requested):
Please explain unclear operation with cell_voltage vector.
Also processing voltages and voltages_ext might be done in lambda, so you could remove duplicate.

mavros/src/plugins/sys_status.cpp
amilcarlucas: > Why do you need nr_cells? Why not to use cell_voltage.size()?
because if for some reason a cell has more than 65,534V it get coalesced with the next one. Hence the bookkeeping with the extra variable.

> And why do you need to erase()?
Using erase allows up to neatly only present the actual number of cells in the ROS message.

> reserve do not change size(), only capacity().
Yes I an aware of that. This method gets a neat result with very low complexity. Erase will not realloc, nor move data around.

> processing voltages and voltages_ext might be done in lambda, so you could remove duplicate.
The hadling in voltages and voltages_ext is different. The duplication is required. :(

amilcarlucas: Sorry, you are 100% correct about using .size() and removing nr_cells

I changed it now, and corrected other small innacuracy

vooon: `nr_cell` always will be equal to `size()` because it's increase done after push_back().
Lambda could look like that:
```
auto copy_cells = [&](uint16_t v, bool is_ext) {
  if (v == UINT16_MAX || (is_ext && v == 0))
    return;
    
  if (v == UINT18_MAX-1) {
    volt_acc = v;
    return;
  }
  
  auto cv = (static_cast<int32_t>(volt_acc) + v) * 0.001f;
  volt_acc = 0;
  batt_msg->cell_voltage.push_back(cv);
  total_volt += cv;
}

for (auto v : bs.voltages) copy_cell(v, false);
for (auto v : bs.voltages_ext) copy_cell(v, true);

vooon: And about ROS message: you don't have to worry about erasing capacity(), only size() matters.

amilcarlucas:
rebased

mavros/src/plugins/sys_status.cpp
vooon: // i'm curious, that chemistry type has so huge cell voltage?

amilcarlucas: yes, you are correct the erase is not needed at all. Removed it, sorry.

amilcarlucas: I'm not sure about chemistry but if you have a > 16S battery, you will need that mechanism to correctly get the battery voltage. According to the mavlink standard:
https://mavlink.io/en/messages/common.html#BATTERY_STATUS

vooon: Ah, yes, so that a hack to send above 65 total volts. I changed lambda to be able to handle more than one (2**16-2) "cells".
Please note that your code has error in `else {acc=0}` block because you're clearing accumulator before use.

amilcarlucas:
Thanks for spoting that nasty bug. fixed and squashed all commits.

vooon (approved):
LGTM, Thanks!

PR #1706, opened 2022-02-15
Diagnose up to 4 batteries

amilcarlucas:
This requires #1704

I am open to suggestions on how to make batt_diag an array.
The copy constructor is disabled and the diagnostics_updater class contains a unique mutex. That prevents me to use the "normal" C++ instance array initialization methods :(


vooon (commented):
Requires #1704 to review.

Suggestion: maybe you can use list initializer and array<>?

amilcarlucas:
I tried that, and I could not get it to run. That is one of the reason I did separated commits, so that the very last commit explicitly makes the ugly code prominent. I am very open to working suggestions :)

PR #1707, opened 2022-02-15
sys_status.cpp: ignore sys_status mavlink messages from gimbals

amilcarlucas:


mavros/src/plugins/sys_status.cpp
```diff
@@ -761,6 +761,9 @@ class SystemStatusPlugin : public plugin::PluginBase
 
 	void handle_sys_status(const mavlink::mavlink_message_t *msg, mavlink::common::msg::SYS_STATUS &stat)
 	{
+		if ((mavlink::minimal::MAV_COMPONENT)msg->compid == mavlink::minimal::MAV_COMPONENT::COMP_ID_GIMBAL)
```
vooon, line 764: Plase use static_cast & using.

```
using MC = mavlink::minimal::MAV_COMPONENT;

if (static_cast<MC>(msg->compid) == MC::COMP_ID_GIMBAL)
    return;
```

vooon (changes requested):
Do not use C-style casts please.

mavros/src/plugins/sys_status.cpp
amilcarlucas: done

vooon (approved):
LGTM, Thanks!

PR #1712, opened 2022-02-17
sys_status.cpp: fix enabling of mem_diag and hwst_diag

amilcarlucas:


amilcarlucas:
This makes the diagnostics run, tested in HW

vooon (approved):
That potentially can have threading access trouble because last_rcd isn't atomic (and can't be), so we have to use either mutex or std::atomic<uint64_t> to store stamp nanoseconds.

```
struct A {
  std::atomic<uint64_t> stamp;
};

a.stamp = ros::Time::now().toNSec();

auto stamp = ros::Time::fromNSec(a.stamp.load());
```

But overall looks ok.

PR #1714, opened 2022-02-18
initialize a plugin before getting subscriptions

standmit:
After initializing and receiving parameters from a ParameterServer, a custom plugin can decide which subscriptions it needs.

### DIFF
--- a/mavros/src/lib/mavros.cpp
+++ b/mavros/src/lib/mavros.cpp
@@ -325,6 +325,7 @@ void MavRos::add_plugin(std::string &pl_name, ros::V_string &blacklist, ros::V_s
 
 		ROS_INFO_STREAM("Plugin " << pl_name << " loaded");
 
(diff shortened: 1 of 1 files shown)


vooon (changes requested):
I don't like that approach. It's not portable to ROS2, as there you can't really expect parameter to be set on init stage.

I'd prefer just so subscribe to all possible messages and then ignore unneeded ones.

standmit:
> I'd prefer just so subscribe to all possible messages and then ignore unneeded ones.

There is no way to subscribe to all messages, isn't there? I mean, in current version you need to know MSG_ID at compile time. But earlier initializing solves this problem.

vooon:
If you don't know msg ID, you won't be able to decode the message. Could you please give overview on what you're trying to get. Maybe i can suggest some other ways?

Issue #1588, opened 2021-06-28
ROS2 `sys_status` plugin services callbacks not executing


blakermchale:
----

### Issue details

I've built `mavros` in ROS Foxy and am running it using `mavros_node`. Services like `~/arming` work for the `command` plugin. However, for the `sys_status` plugin I cannot call any of the services like `set_mode` or `vehicle_info_get`. I am using the command:  
```bash
ros2 service call /mavros/set_mode mavros_msgs/srv/SetMode "{'custom_mode': 'AUTO.TAKEOFF'}"
```  
This service call ends up hanging and never returning. It just says this message:  
```bash
requester: making request: mavros_msgs.srv.SetMode_Request(base_mode=0, custom_mode='AUTO.TAKEOFF')
```  
I've added in rclcpp log info messages to the service callbacks and determined they are never being entered since the messages are not logged.

### MAVROS version and platform

Mavros: 2.0.3
ROS: Foxy
Ubuntu: 20.04

### Autopilot type and version

[  ] ArduPilot
[X] PX4

Version: master or 9d3e18dbe10700a13f3048f2807dc4b1a62eee62

### Node logs

```
[mavros_node-5] [INFO][main():52]: Starting mavros_node container
[mavros_node-5] [INFO][main():53]: FCU URL: udp://:14540@127.0.0.1:14557
[mavros_node-5] [INFO][main():54]: GCS URL: 
[mavros_node-5] [INFO][main():55]: UAS Prefix: /uas1
[mavros_node-5] [INFO][main():57]: Starting mavros router node
[mavros_node-5] [INFO][Router():183]: Built-in SIMD instructions: SSE, SSE2
[mavros_node-5] [INFO][Router():184]: Built-in MAVLink package version: 2021.5.5
[mavros_node-5] [INFO][Router():185]: Known MAVLink dialects: common ardupilotmega ASLUAV all development icarous matrixpilot paparazzi standard uAvionix ualberta
[mavros_node-5] [INFO][Router():186]: MAVROS Router started
[mavros_node-5] [INFO][add_endpoint():108]: Requested to add endpoint: type: 0, url: udp://:14540@127.0.0.1:14557
[mavros_node-5] [INFO][add_endpoint():138]: Endpoint link[1000] created
[mavros_node-5] [INFO][add_endpoint():142]: link[1000] opened successfully
[mavros_node-5] [INFO][add_endpoint():108]: Requested to add endpoint: type: 2, url: /uas1
[...]

PR #1717, opened 2022-02-27
Maybe fix sys status callbacks

rob-clarke:
Tentative fix for #1588 

Adding the timer to the `srv_cg` callback group appears to resolve the issue. Not entirely sure why that is. Potentially rclcpp/rmw internals causing a deadlock.


vooon (approved):
Interesting. Thanks!

PR #1720, opened 2022-03-03
libmavconn: fix UDP deadlocks

SylvainPastor:
Same problems as for the TCP:
  - #1682: fix std::system_error when tcp interface loses connection
  - #1679: fix deadlock when call close()


SylvainPastor:
Hello, I can offer you to make changes to the libmavconn library to redo complete tests and refactor the common code between the different types of connections. What do you think ?

vooon (approved):
LGTM, Thanks!

Test fixes also would be nice!

Issue #1719, opened 2022-03-01
Fake GPS Velocity Frame Issue


jgoppert:
### Issue details

We are using fake GPS on PX4 and noticed that we get drift and high velocity residuals. We are still testing further, but suspect these lines. The velocity is calculated using a single step finite difference approximation. vel = ([pos[t_1] - pos[t_0]) / dt. The problem is that the frame is incorrect. These positions are in earth centered, earth fixed coordinates. This is not the north east down coordinate system. As soon as we move from hover, our residuals spike. We will submit a patch after test flying it.

https://github.com/mavlink/mavros/blob/c035b8ff1ffb042db235648371f5d7d2e531fae4/mavros_extras/src/plugins/fake_gps.cpp#L309
https://github.com/mavlink/mavros/blob/c035b8ff1ffb042db235648371f5d7d2e531fae4/mavros_extras/src/plugins/fake_gps.cpp#L343

![image](https://user-images.githubusercontent.com/473772/156104783-1a92908b-5c3c-4199-a5c5-2d988b360145.png)

### MAVROS version and platform

Mavros: Master
ROS: Noetic
Ubuntu: 20.04

Bearixal:
This is related: #1639

PR #1722, opened 2022-03-08
WIP: Correct velocity frame for fake_gps.

jgoppert:
Closes Issues #1719 

Do not merge yet, awaiting flight test.

### DIFF
--- a/mavros_extras/src/plugins/fake_gps.cpp
+++ b/mavros_extras/src/plugins/fake_gps.cpp
@@ -227,7 +227,21 @@ private:
 			ROS_INFO_STREAM("FGPS: Caught exception: " << e.what() << std::endl);
 		}
 
-		Eigen::Vector3d vel = (old_ecef - current_ecef) / (stamp.toSec() - old_stamp);	// [m/s]
+		Eigen::Matrix<double, 3, 3> dcm_ned_ecef;
+		double lat = geodetic.x();
+		double lon = geodetic.y();
+		dcm_ned_ecef(0, 0) = -sin(lat)*cos(lon);
+		dcm_ned_ecef(0, 1) = -sin(lat)*sin(lon);
+		dcm_ned_ecef(0, 2) = cos(lat);
+		dcm_ned_ecef(1, 0) = -sin(lon);
+		dcm_ned_ecef(1, 1) = cos(lon);
+		dcm_ned_ecef(1, 2) = 0;
+		dcm_ned_ecef(2, 0) = -cos(lat)*cos(lon);
+		dcm_ned_ecef(2, 1) = -cos(lat)*sin(lon);
+		dcm_ned_ecef(2, 2) = -sin(lat);
+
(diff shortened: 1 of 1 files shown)


mavros_extras/src/plugins/fake_gps.cpp
```diff
@@ -227,7 +227,21 @@ class FakeGPSPlugin : public plugin::PluginBase,
...
+		double lon = geodetic.y();
+		dcm_ned_ecef(0, 0) = -sin(lat)*cos(lon);
+		dcm_ned_ecef(0, 1) = -sin(lat)*sin(lon);
+		dcm_ned_ecef(0, 2) = cos(lat);
+		dcm_ned_ecef(1, 0) = -sin(lon);
+		dcm_ned_ecef(1, 1) = cos(lon);
+		dcm_ned_ecef(1, 2) = 0;
+		dcm_ned_ecef(2, 0) = -cos(lat)*cos(lon);
+		dcm_ned_ecef(2, 1) = -cos(lat)*sin(lon);
+		dcm_ned_ecef(2, 2) = -sin(lat);
```
vooon, line 241: Perhaps `<<` or initializer list would be more readable:

```
dcm_ned_ecef {
 {1, 2, 3},
 {4, 5, 6},
 {7, 8, 9},
};
```
https://eigen.tuxfamily.org/dox/group__TutorialMatrixClass.html

vooon (approved):

mavros_extras/src/plugins/fake_gps.cpp
jgoppert: Thanks for the review. I couldn't get brace initializer lists to work with eigen, so went with the stream op approach.

Bearixal:
Hey, i just noticed that the whole ECEF transformation is obsolete with this change. You could simply use ENU instead of ECEF as function argument in `send_fake_gps()` and remove the member variable `ecef_origin` completely.
To still have access to the local `geodetic` variable in `send_fake_gps()` use this function as described [here](https://geographiclib.sourceforge.io/html/classGeographicLib_1_1LocalCartesian.html#details): 
```
const GeographicLib::Geocentric& earth = GeographicLib::Geocentric::WGS84();
GeographicLib::LocalCartesian proj(map_origin.x(), map_origin.y(), map_origin.z(), earth);
proj.Reverse(pos_enu.x(), pos_enu.y(), pos_enu.z(), geodetic.x(), geodetic.y(), geodetic.z());
```

Because the ellipsoid height will be returned, it is necessary to convert the height to the geoid height like [here](https://geographiclib.sourceforge.io/html/classGeographicLib_1_1Geoid.html#details) or by using the mavros function:
```
geodetic.z() += m_uas->ellipsoid_to_geoid_height(...);
```

Don't forget to convert the enu coordinates into ned coordinates with the build in mavros helper function (`ftf::transform_frame_enu_ned(pos_enu);`).

PR #1744, opened 2022-05-12
mount_control.cpp: detect MOUNT_ORIENTATION stale messages

amilcarlucas:
fix pitch sign
correct MountConfigure response success

### DIFF
--- a/mavros_extras/src/plugins/mount_control.cpp
+++ b/mavros_extras/src/plugins/mount_control.cpp
@@ -39,16 +39,17 @@ class MountStatusDiag : public diagnostic_updater::DiagnosticTask
 public:
 	MountStatusDiag(const std::string &name) :
 		diagnostic_updater::DiagnosticTask(name),
+		_last_orientation_update(0, 0),
+		_debounce_s(NAN),
 		_roll_deg(NAN),
 		_pitch_deg(NAN),
 		_yaw_deg(NAN),
 		_setpoint_roll_deg(NAN),
 		_setpoint_pitch_deg(NAN),
 		_setpoint_yaw_deg(NAN),
-		_error_detected(false),
 		_err_threshold_deg(NAN),
-		_mode(255),
-		_debounce_s(NAN)
+		_error_detected(false),
+		_mode(255)
 	{ }
 
 	void set_err_threshold_deg(float threshold_deg) {
@@ -61,11 +62,12 @@ public:
 		_debounce_s = debounce_s;
 	}
 
-	void set_status(float roll_deg, float pitch_deg, float yaw_deg) {
+	void set_status(float roll_deg, float pitch_deg, float yaw_deg, ros::Time timestamp) {
 		std::lock_guard<std::mutex> lock(mutex);
 		_roll_deg = roll_deg;
 		_pitch_deg = pitch_deg;
 		_yaw_deg = yaw_deg;
+		_last_orientation_update = timestamp;
 	}
 
 	void set_setpoint(float roll_deg, float pitch_deg, float yaw_deg, uint8_t mode) {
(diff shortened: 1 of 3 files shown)


mavros_extras/src/plugins/mount_control.cpp
```diff
@@ -209,13 +218,14 @@ class MountControlPlugin : public plugin::PluginBase {
 	 */
 	void handle_mount_orientation(const mavlink::mavlink_message_t *msg, mavlink::common::msg::MOUNT_ORIENTATION &mo)
 	{
-		auto q = ftf::quaternion_from_rpy(Eigen::Vector3d(mo.roll, mo.pitch, mo.yaw) * M_PI / 180.0);
+		const auto timestamp = ros::Time::now();
+		mo.pitch = -mo.pitch; // pitch angle is inverted to fix a gremsy implementation bug
```
Jaeyoung-Lim, line 222: I am not sure if this is generally applicable and seems like a specific problem woth gremsy gimbals

amilcarlucas: I made it parameter configurable, so it will workaround any existing and/or future gimbal quirks

mavros_extras/src/plugins/mount_control.cpp
```diff
@@ -209,13 +242,24 @@ class MountControlPlugin : public plugin::PluginBase {
...
+		if (negate_measured_roll) {
+			mo.roll = -mo.roll;
+		}
+		if (negate_measured_pitch) {
+			mo.pitch = -mo.pitch;
+		}
+		if (negate_measured_yaw) {
+			mo.yaw = -mo.yaw;
+			mo.yaw_absolute = -mo.yaw_absolute;
+		}
```
Jaeyoung-Lim, line 256: This will make the frames inconsistent (oscillating between left hand coordinate or right hand coordinate systems). Shouldn't this be something that is fixed on the driver level?

amilcarlucas: Gremsy probably behaves differently from storm32. I can not test I only have greasy. But what I can say ist that this let's the users fix issues on their Gimbals without changing the code.

Jaeyoung-Lim: If this is a device specific fix that needs to happen, why not handle it on the user app level? I still think mavros is the wrong place to make a workaround for a device specific problem.

mavros_extras/src/plugins/mount_control.cpp
```diff
@@ -170,12 +179,33 @@ class MountControlPlugin : public plugin::PluginBase {
...
+		mount_nh.param<bool>("negate_measured_yaw", negate_measured_yaw, false);
+		if (!mount_nh.getParam("negate_measured_roll", negate_measured_roll)) {
+			ROS_WARN("Could not retrive negate_measured_roll parameter value, using default (%d)", negate_measured_roll);
+		}
+		if (!mount_nh.getParam("negate_measured_pitch", negate_measured_pitch)) {
+			ROS_WARN("Could not retrive negate_measured_pitch parameter value, using default (%d)", negate_measured_pitch);
+		}
+		if (!mount_nh.getParam("negate_measured_yaw", negate_measured_yaw)) {
+			ROS_WARN("Could not retrive negate_measured_yaw parameter value, using default (%d)", negate_measured_yaw);
+		}
```
vooon, line 195: That if's is useless after `param()`. Or it just left as a debug?

mavros_extras/src/plugins/mount_control.cpp
amilcarlucas: Because the diagnostics code needs to handle it at this level, take a look at the diagnostics class in the mount_control.cpp file

Jaeyoung-Lim: Not sure if I understand what you mean... why does the diagnostic status influenced by the negation in the mount control handling?

Jaeyoung-Lim: @amilcarlucas But if the gremsy gimbal driver is implemented the opposite way, shouldn't the user app command -70° in the first place?

amilcarlucas: **That** is the gremsy bug, you command 70 and it measures -70 !

Jaeyoung-Lim:
@vooon @amilcarlucas I still think this is the wrong place to fix the problem, However if it unblocks @amilcarlucas I think it doesn't hurt to have something like this merged in. I hope we don't forget about this and remove this as soon as the gremsy bug is fixed.

PR #1760, opened 2022-06-25
Update for GeographicLib 2.0

AchmadFathoni:
Library file now called libGeographicLib.so, etc., instead of libGeographic.so.

### DIFF
--- a/libmavconn/cmake/Modules/FindGeographicLib.cmake
+++ b/libmavconn/cmake/Modules/FindGeographicLib.cmake
@@ -8,7 +8,10 @@
 
 find_path (GeographicLib_INCLUDE_DIRS NAMES GeographicLib/Config.h)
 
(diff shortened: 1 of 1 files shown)


vooon:
Maybe it's better to check both names, to support older releases?

TSC21:
This needs to be made backwards compatible for sure.

vooon:
Where `GeographicLib_VERSION_MAJOR` is defined?

vooon:
https://cmake.org/cmake/help/latest/command/find_library.html

Also probably all what needs to be done - add new name to NAMES?

AchmadFathoni:
> Where `GeographicLib_VERSION_MAJOR` is defined?

I'm sorry, I thought Cmake automatically provides x_VERSION_MAJOR. I think I will learn more about Cmake before finishing this PR.

PR #1770, opened 2022-08-03
Fill out the rangefinder properties from the rangefinder message

AndersonRayner:
Fills out the rangefinder properties (`field of view`, `min_range`, and `max_range`) in the `rangefinder_msg` from the mavlink message.  

I haven't updated the `radiation_type` field as the enums in the [sensors_mgs::Range](http://docs.ros.org/en/noetic/api/sensor_msgs/html/msg/Range.html) don't match the [MAV_DISTANCE_SENSOR](https://mavlink.io/en/messages/common.html#MAV_DISTANCE_SENSOR) enums.


vooon (approved):
LGTM, Thanks!

vooon:
But need to wait for mavlink to have that new fields.

Issue #1774, opened 2022-08-21
Could not find GeographicLib


acxz:
This is only bug and feature tracker, please use it
to report bugs or request features.

----

### Issue details

Cannot build
GeographicLib not found

```
CMake Error at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find GeographicLib (missing: GeographicLib_LIBRARIES)
Call Stack (most recent call first):
  /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:594 (_FPHSA_FAILURE_MESSAGE)
  /opt/ros/noetic/share/libmavconn/cmake/Modules/FindGeographicLib.cmake:14 (find_package_handle_standard_args)
  CMakeLists.txt:42 (find_package)


-- Configuring incomplete, errors occurred!
```

GeographicLib version: 2.1.1

### MAVROS version and platform

Mavros: 1.13.0-1
ROS: Noetic
OS: Archlinux

### Autopilot type and version

[  ] ArduPilot
[  ] PX4

Version: ?3.7.1?

N/A

### Node logs

```
copy output of mavros_node. Usually console where you run roslaunch
```
N/A

### Diagnostics

```
place here result of:
rostopic echo -n1 /diagnostics
```
N/A
### Check ID

```
rosrun mavros checkid
```
N/A

vooon:
Fixed by #1775.

acxz:
I this issue also affects the ros2 branch and should stay open until those commits find their way there.

PR #1775, opened 2022-08-21
use already installed FindGeographicLib.cmake

acxz:
Solve #1774

The `FindGeographicLib.cmake` is already shipped with the distribution through the geographiclib package, hence our custom one (which seems to fail for later geographiclib versions) is not needed anymore.

Note: for debian based systems, `FindGeographicLib.cmake` is installed elsewhere, this path would needed to be added to the paths that `find_package` searches through.

I know that primary focus has been shifted to the ros2 branch, so let me know if I should make another PR to that branch. And if so, that commit would also close #1760


vooon:
I'd prefer to keep existing module, but try to use package one's first.
There are still lots of os versions with kinetic, where i'd prefer to keep possibility to build master.

acxz:
@vooon I believe the module offered by the package is included in kinetic as well.

This [link](https://packages.ubuntu.com/bionic/amd64/libgeographic-dev/filelist) shows that the module is located in bionic. Since kinetic is EOL, Ubuntu doesn't have the package database for xenial. However, I don't image that how the package has been built has been changed since it was first created.

Considering that xenial was released in [April 21st, 2016](https://wiki.ubuntu.com/XenialXerus/ReleaseSchedule), and that the package was first released in [2010 on Debian](https://metadata.ftp-master.debian.org/changelogs//main/g/geographiclib/geographiclib_1.46-2_changelog), xenial should have the same package structure and the `FindGeographicLib.cmake` file.

acxz:
> That's why i'd like to keep possibility to build there.

For sure, but even if we wanted to we couldn't bloom mavros into a version older than melodic anyway, so any user wouldn't have even have access to this commit. And even if they did want to compile the master branch onto unsupported ros versions, it would still work back to ROS Indigo (ubuntu 14) released in 2014. (I can check further if you want).

There is most likely going to be another compile issue that pops before this issue does when trying to compile the master branch for hardware/operating systems 8 years ago.

> But the same time i'm okay to drop deprecated things.

I don't think the FindGeographicLib.cmake that we have is deprecated, it just doesn't stay updated since we maintain it ourselves. Increasing the codebase burden we have to maintain. So instead of using upstream's FindGeographicLib.cmake which will work for whatever distro you are on, by maintaining it ourselves, we have to keep the burden of keeping a single FindGeographicLib.cmake that works for all the distros we want.

vooon (approved):

vooon:
@acxz okay. Thanks for the contribution!

PR #1780, opened 2022-09-07
guided_target: accept position-target-global-int messages

snktshrma:
This PR attempts to implement the suggestions in the PR #1184 by @rmackay9 to allow mavros to accept [POSITION_TARGET_GLOBAL_INT](https://mavlink.io/en/messages/common.html#POSITION_TARGET_GLOBAL_INT) message from FCU or GCS.

There were suggestions by @TSC21 and @vooon to implement a separate plugin for the purpose and use POSITION_TARGET_GLOBAL_INT message type.

There's an issue I found:

- After commanding the vehicle in off-board mode to a specific location, vehicle tends to go to a specific position(different from what it is commanded) everytime.

Hoping for a resolution to this issue.

### DIFF
--- a/mavros_extras/src/plugins/guided_target.cpp
+++ b/mavros_extras/src/plugins/guided_target.cpp
@@ -1,0 +1,199 @@
+/**
+ * @brief Guided target plugin
+ * @file guided_target.cpp
+ * @author Randy Mackay <rmackay9@yahoo.com> , Sanket Sharma <sharma.sanket272@gmail.com>
+ *
+ * @addtogroup plugin
+ * @{
+ */
+/*
+ * Copyright 2022 Sanket Sharma, Randy Mackay.
+ *
+ * This file is part of the mavros package and subject to the license terms
+ * in the top-level LICENSE file of the mavros repository.
+ * https://github.com/mavlink/mavros/tree/master/LICENSE.md
+ */
+
+#include <mavros/mavros_plugin.h>
+#include <mavros/setpoint_mixin.h>
+#include <eigen_conversions/eigen_msg.h>
+
+#include <geometry_msgs/PoseStamped.h>
+
+#include <mavros_msgs/SetMavFrame.h>
+#include <geographic_msgs/GeoPoseStamped.h>
+
+#include <GeographicLib/Geocentric.hpp>
+
+#include <mavros_msgs/PositionTarget.h>
+#include <geographic_msgs/GeoPointStamped.h>
+#include <mavros_msgs/GlobalPositionTarget.h>
+
+namespace mavros {
+namespace extra_plugins {
+	
+/**
+ * @brief guided target plugin
+ *
+ * Send and receive setpoint positions from FCU controller.
+ */
+class GuidedTargetPlugin : public plugin::PluginBase,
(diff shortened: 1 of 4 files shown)


vooon (commented):
I see only few minor style fixes, but overall looks good.

But i'd like to have some more meaningful name for the plugin. "Offboard" looks messy as it doesn't say what it does.

snktshrma:
> But i'd like to have some more meaningful name for the plugin. "Offboard" looks messy as it doesn't say what it does.

Thanks @vooon for the review! Can "guided_target" name be used instead?

vooon:
@snktshrma yes, that sounds better, but i'd like to hear others too. CC @rmackay9, @TSC21 .

rmackay9:
Hi, yes, I'm happy with the name "guided_target".  I'm very happy to see this plugin.

snktshrma:
Thanks @rmackay9 @vooon for review! I have made the required changes

mavros_extras/src/plugins/guided_target.cpp
```diff
@@ -0,0 +1,192 @@
+#include <mavros/mavros_plugin.h>
```
vooon, line 1: Please add copyright header, similar to other plugins.

mavros_extras/src/plugins/guided_target.cpp
```diff
@@ -0,0 +1,192 @@
...
+		setpointg_pub = spg_nh.advertise<geometry_msgs::PoseStamped>("/move_base_simple/goal", 10);
+
+
+		// Subscriber for global origin (aka map origin).
+		gp_origin_sub = spg_nh.subscribe("global_position/gp_origin", 10, &GuidedTargetPlugin::gp_origin_cb, this);
+
+	
+
+		
+	}
```
vooon, line 58: Please remove blank lines.

mavros_extras/src/plugins/guided_target.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+	Subscriptions get_subscriptions() override
+	{
+		//return { /* Rx disabled */ };
```
vooon, line 62: Please remove that comment.

vooon (approved):
Looks good, just a few style fix needed.

snktshrma:
@vooon thanks for the review! Made the required changes

shubham-shahh:
Hi @snktshrma can you please elaborate upon the issue you have mentioned in the description of this PR?

snktshrma:
Hi @shubham-shahh!
So the actual issue is when we set the target position for the vehicle, it doesn't go to that place. Instead it goes to only a specific position different from what's specified.
I'll try to add a video for the issue

shubham-shahh:
> Hi @shubham-shahh!
> So the actual issue is when we set the target position for the vehicle, it doesn't go to that place. Instead it goes to only a specific position different from what's specified.
> I'll create an issue for this and try to add a video of this

Sure, a video would be great, because I'm still not clear about the issue. Also is it only appearing in SILT? or on hardware?

snktshrma:
@shubham-shahh I'll upload a video soon!
I've not tested this part yet on hardware. It's appearing in SITL

PR #1805, opened 2022-11-22
mavros_extras: Fix buggy check for lat/lon ignored

mortenfyhn:
I stumbled upon what I think is a bug here, thanks to a `-Wparentheses` warning I got (clang 13, but I'm guessing this is on gcc too).

The warning text explains:

> & has lower precedence than >; > will be evaluated first

which I don't think is what we want here. The if statement will evaluate to false in some cases where it shouldn't, such as when:

* `type_mask` is 2 = 0b0010
* `IGNORE_LATITUDE` is 1 = 0b0001
* `IGNORE_LONGITUDE` is 2 = 0b0010

then the if statement evaluates to

* `0b0010 & (0b0001 | 0b0010)  > 0`
* `0b0010 & 0b0011 > 0`
* `0b0010 & true` because > evaluates first
* `0b0010 & 0b0001`
* `0b0000`
* `false`

which is wrong!

We can fix this by adding parentheses:

```cpp
if ((position_target.type_mask & (mavros_msgs::GlobalPositionTarget::IGNORE_LATITUDE | mavros_msgs::GlobalPositionTarget::IGNORE_LONGITUDE)) > 0)
```

or just by removing the `> 0` comparison and allowing the statement to implicitly convert to bool

```cpp
if (position_target.type_mask & (mavros_msgs::GlobalPositionTarget::IGNORE_LATITUDE | mavros_msgs::GlobalPositionTarget::IGNORE_LONGITUDE))
```

### DIFF
--- a/mavros_extras/src/plugins/guided_target.cpp
+++ b/mavros_extras/src/plugins/guided_target.cpp
@@ -149,7 +149,7 @@ private:
 	void handle_position_target_global_int(const mavlink::mavlink_message_t *msg, mavlink::common::msg::POSITION_TARGET_GLOBAL_INT &position_target)
 	{
 		/* check if type_mask field ignores position*/
-		if (position_target.type_mask & (mavros_msgs::GlobalPositionTarget::IGNORE_LATITUDE | mavros_msgs::GlobalPositionTarget::IGNORE_LONGITUDE) > 0) {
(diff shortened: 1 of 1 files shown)


vooon (approved):
Good catch!

PR #1807, opened 2022-11-28
mavros_extras: Fix a comparison that shouldn't be bitwise

mortenfyhn:
Hi all, I think this is a bug. This expression is written with a bitwise OR `|`, but it seems like it should have been written with a logical OR `||`.


vooon (approved):
LGTM, Thanks!

PR #1808, opened 2022-11-28
mavros_extras: Fix a sequence point warning

mortenfyhn:
This line triggers a `-Wsequence-point` warning. You can read about it here: https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html

It seems to be that doing `++` (which modifies a variable) and assigning the result back to the same variable can be undefined behaviour.

In this case, we don't need to pre-increment, we can just add one, and then save the result back to the variable. This makes the warning go away, and seems to make sense.


vooon (approved):
LGTM, Thanks!

PR #1811, opened 2022-12-19
Implement debug float array handler

sverrevr:
Implements the debug float array which was previously not implemented.
We chose to use the debug float array name to match https://github.com/mavlink/mavlink/pull/734 
Tested in flight and works.


vooon (approved):
LGTM, Thanks!

danieffe:
If I want to display the debug_array payload on QGC, for example in the flyview, what I have to do?

vooon:
@danieffe that plugin not sending any messages to the QGC. It's primary job is to receive values from the FCU, or in some cases send them back.

QGC only may receive the messages sent by FCU.

PR #1817, opened 2022-12-26
use hpp instead of deprecated .h pluginlib headers

lucasw:


vooon:
I'm not sure if that works for older releases, e.g. Kinetic.

lucasw:
This put them into kinetic-devel https://github.com/ros/pluginlib/commit/64829f90d73e68d2555c8ef4d425a7ab4dea4c51

vooon (approved):
Ok, LGTM, Thanks!

PR #1829, opened 2023-02-15
Made it such that the gp_origin topic publisher is latched.

snwu1996:
Sometimes when a subscriber subs to "gp_origin" topic it doesn't receive a message in time leading to race conditions. Made it such that the publisher is latched so subscribers always gets the gp_origin on demand.


vooon (approved):
Probably good to go.

vooon:
LGTM, Thanks!

PR #1833, opened 2023-02-22
plugins: ftp: fix null pointer dereference

### LINKED ISSUE #1695: ROS2: FTP segfault on list call
### Issue details

Node segfaults after any call to `/mavros/ftp/list`.

Mavros: 2.0.5 (master)
ROS: Galactic
Ubuntu: 20.04

[x] ArduPilot
[ ] PX4

Version: 4.0.7

lopsided98:
df4e529 mistakenly switched `PayloadHeader::data` from using a non-standards compliant (but accepted by most compilers) flexible array to a pointer. This resulted in an attempt to dereference the uninitialized contents of the array.

This patch eliminates `PayloadHeader::data` and instead makes `FTPRequest::data()` use pointer arithmetic to get the data buffer from within the payload buffer.

This fixes the crash, but the plugin still doesn't work quite right, at least with ArduPilot. The only command I have been able to execute successfully is checksum.

Fixes #1695

### DIFF
--- a/mavros/src/plugins/ftp.cpp
+++ b/mavros/src/plugins/ftp.cpp
@@ -71,7 +71,6 @@ public:
     uint8_t req_opcode;         ///< Request opcode returned in kRspAck, kRspNak message
     uint8_t padding[2];         ///< 32 bit aligment padding
     uint32_t offset;            ///< Offsets for List and Read commands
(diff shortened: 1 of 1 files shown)


vooon (approved):
LGTM, Thanks!

Issue #1564, opened 2021-04-12
ROS2: port mavros launch


vooon:
We could try to continue use of xml, but probably launch.py would be preferable.

Vicidel:
What would need to be done here?
Also, is there any solution to run the package without it for short-term testing?

vooon:
We need to port node lanuch and variants for PX4 and APM. Possibly we may use component loader instead of mavros_node.
But currently i had problems with it.

vooon:
@Vicidel probably i misinterpret your question. It is possible to run ether combined mavros_node or composition nodes without launch. They just helps to manage set of nodes, parameters and remappings.

You can run combined node thus way (params file provides settings, [see docs](https://docs.ros.org/en/foxy/Tutorials/Parameters/Understanding-ROS2-Parameters.html#load-parameter-file-on-node-startup)):
```
ros2 run mavros mavros_node --ros-args --params-file /ws/mavros_param.yaml
```

JosepBC:
Hello, I'm working on this, I've succesfully ported node.launch to ROS2, right now doing tests with Pixhawk 4 with ArduPilot but without plugin.yaml and config.yaml as I don't know which plugins/parameters are ported to ROS2. 
I'll do a PR with all I have in a few hours, someone has the config.yaml and plugins.yaml so I can test everything?

vooon:
Plugin lists should be the same, except that black/white checked to allow/deny lists.

As for parameters it's harder, because with new parameter protocol each plugin has it's own node.
And some of parameters have been renamed, so i suggest you to check sources.

eliabruni:
Hello, this is blocking our ROS2 development, unfortunately, do you have an ETA for tackling the issue? Thanks!

JosepBC:
I just ported node.launch as I had to switch back to ros1. If you want I can PR that.
But in general, you can use this article to port all the launch files.
https://docs.ros.org/en/foxy/How-To-Guides/Launch-files-migration-guide.html

Rainerino:
I have also manually ported `px4.launch` and `apm.launch`, but the parameters don't seem to be loaded correctly

eliabruni:
thanks for the pointer, @JosepBC.

@Rainerino , would you be so kind to share your code? perhaps we could try to tackle it together?

Rainerino:
@eliabruni Yeah, of course, I am new to ROS2 and had to face quite a steep learning curve coming from ROS, would love to have some pointers!

[...]

PR #1747, opened 2022-05-17
make Simple launch file

Ryujiyasu:
1. separate param file from launch folder
2. delete ros launch files
3. make simple launch file

PR #1834, opened 2023-02-23
fix px4 launch file for ROS2

vacabun:
I modified px4 launch file for ROS2, and tested it in the simulation environment.


vooon (commented):
Overally looks good. Can you please do the same for apm?

mavros/launch/px4_config.yaml
```diff
@@ -138,144 +152,160 @@ mission:
...
+      horizontal_fov_ratio: 1.0   # horizontal_fov = horizontal_fov_ratio * msg.field_of_view
+      vertical_fov_ratio: 1.0     # vertical_fov = vertical_fov_ratio * msg.field_of_view
+      # custom_orientation:       # Used for orientation == CUSTOM
+      #   roll: 0
+      #   pitch: 270
+      #   yaw: 0
+    laser_1_sub:
+      subscriber: true
+      id: 3
+      orientation: PITCH_270
```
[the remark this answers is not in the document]
vacabun: Ok, I see

vacabun:
OK, I will try to do the same for apm.

vooon:
Yes, bridge is not needed anymore, because you have router.
And I've removed event_launcher as it should be done differently, like a plugin to launch system, or something like that.

vooon:
Fix #1564
Replace #1747

vacabun:
OK，I have removed them.

mavros/launch/px4_config.yaml
```diff
@@ -44,87 +47,98 @@ cmd:
...
+    frame_id: "map"             # origin frame
+    child_frame_id: "base_link" # body-fixed frame
+    rot_covariance: 99999.0   # covariance for attitude?
+    gps_uere: 1.0             # User Equivalent Range Error (UERE) of GPS sensor (m)
+    use_relative_alt: true    # use relative altitude for local coordinates
+    tf:
+      send: false               # send TF?
+      frame_id: "map"  # TF frame_id
+      global_frame_id: "earth"  # TF earth frame_id
+      child_frame_id: "base_link" # TF child_frame_id
```
vooon, line 61: In ROS2 parameters are flat, so i had to change names. Please have a look:

https://github.com/mavlink/mavros/blob/6e86500318385b5770653c6e47e3c0cdda2e3181/mavros/src/plugins/global_position.cpp#L90-L107

mavros/launch/px4_config.yaml
```diff
@@ -44,87 +47,98 @@ cmd:
...
 
 # rc_io
 # None
 
-# safety_area
-safety_area:
-  p1: {x:  1.0, y:  1.0, z:  1.0}
-  p2: {x: -1.0, y: -1.0, z: -1.0}
+# # safety_area
+/mavros/safety_area:
```
vooon, line 90: Safety area is quite deprecated feature. I've removed it from ROS2.

mavros/launch/px4_config.yaml
```diff
@@ -138,144 +152,161 @@ mission:
...
-    #   roll: 0
-    #   pitch: 270
-    #   yaw: 0
-  laser_1_sub:
-    subscriber: true
-    id: 3
-    orientation: PITCH_270
+/mavros/distance_sensor:
+  ros__parameters:
+    config: |
```
vooon, line 157: Yes, that's what i mean. +1

mavros/launch/px4_config.yaml
```diff
@@ -138,144 +152,161 @@ mission:
...
-    rate_limit: 10.0      # TF rate
-  gps_rate: 5.0           # GPS data publishing rate
+/mavros/fake_gps:
+  ros__parameters:
+    # select data source
+    use_mocap: true         # ~mocap/pose
+    mocap_transform: true   # ~mocap/tf instead of pose
+    use_vision: false       # ~vision (pose)
+    # origin (default: Zürich)
+    geo_origin:
```
vooon, line 201: https://github.com/mavlink/mavros/blob/6e86500318385b5770653c6e47e3c0cdda2e3181/mavros_extras/src/plugins/fake_gps.cpp#L133-L144

So it should look like:
```
fake_gps:
  ros__parameters:
    geo_origin.lat: 47.3667
```

vooon (commented):
Have you tried to use this launch?
I just have an idea, that params file not fully loaded.

Do you know, is it possible to define namespace for params?
Just if you run several copies of the UAS to control few drones, it might be convenient to have same config.

mavros/launch/px4_config.yaml
vacabun: OK, I see.

mavros/launch/px4_config.yaml
vacabun: OK, I see.

mavros/launch/px4_config.yaml
vacabun: OK, I see.

vacabun:
I have used this launch, but just tested a little. 

I think is a good idea. That will be really convenient for using multiple drones.

vacabun:
The plugin `odometry` in apm launch has two parameters.

https://github.com/mavlink/mavros/blob/a55c747cf8386384bb2c7c2eb56bb62f0b6bb2c8/mavros/launch/apm_config.yaml#L228-L232

I can't find the parameters `frame_tf.desired_frame` and `estimator_type` in `mavros_extras/src/plugins/odom.cpp`. 
Should I modify it as px4 launch.

https://github.com/mavlink/mavros/blob/a55c747cf8386384bb2c7c2eb56bb62f0b6bb2c8/mavros/launch/px4_config.yaml#L232-L235

Finally, it looks like:
```
# odom
/mavros/odometry:
  ros__parameters:
    fcu.odom_parent_id_des: "map"    # desired parent frame rotation of the FCU's odometry
    fcu.odom_child_id_des: "base_link"    # desired child frame rotation of the FCU's odometry
```

vooon:
Odom have changed a lot, so it's not a surprise that conf outdated.

We should use parameters what ROS2 have.
https://github.com/mavlink/mavros/blob/6e86500318385b5770653c6e47e3c0cdda2e3181/mavros_extras/src/plugins/odom.cpp#L67-L74

vooon (approved):
LGTM, Thanks!

vooon:
Yes, in ROS2 i've implemented idea of splitting mavlink connections and UAS. So if previously you have to run several copies of whole mavros_node, and maybe gcs_nodes, now it utilizes ros2 loadable node classes.

```
FCU1 -> Router -> UAS1
FCU2 --^          &-> UAS2
                  &-> GCS
```

PR #1838, opened 2023-03-07
fix ROS2 launch parameters namespace and name.

vacabun:
1. The plugin denylist allowlist `plugin_allowlist` and `plugin_denylist`.

2. When change the node's namespace, the plugin parameters cannot be loaded normally.
    The parameter file needs to fit into the namespace.
    Like this：
    ```
    /mavros/**:
      ros__parameters:
        startup_px4_usb_quirk: false
    
    /mavros/**/conn:
      ros__parameters:
        heartbeat_rate: 1.0    # send heartbeat rate in Hertz
        heartbeat_mav_type: "ONBOARD_CONTROLLER"
    ```
    Use `/**` to fit the new namespace.
    It can alse fit the default namespace `mavros`.

### DIFF
--- a/mavros/launch/apm_config.yaml
+++ b/mavros/launch/apm_config.yaml
@@ -1,14 +1,14 @@
 # Common configuration for APM2 autopilot
 #
 # node:
-/mavros:
+/mavros/**:
   ros__parameters:
     startup_px4_usb_quirk: false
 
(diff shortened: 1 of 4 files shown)


vooon (approved):
LGTM, Thanks!

PR #1865, opened 2023-06-12
Fix / suppress some build warnings

mortenfyhn:
Hi all, this does three things:

* Use the [cmake SYSTEM option](https://cmake.org/cmake/help/latest/command/include_directories.html) to have it consider 3rd party library headers as system headers, and thus not emit build warnings from them.
* Fix some build warnings in mavros_extras.
* Remove unneeded trailing semicolon on some namespace declarations.

Feel free to use this if it's useful for you.

### DIFF
--- a/libmavconn/CMakeLists.txt
+++ b/libmavconn/CMakeLists.txt
@@ -35,6 +35,7 @@ catkin_package(
 
 include_directories(
   include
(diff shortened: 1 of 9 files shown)


vooon (approved):
LGTM, Thanks!

PR #1871, opened 2023-06-29
Addition of New OpticalFlow.msg

Vladislavert:
This PR proposes the addition of a new OpticalFlow.msg file to enhance the [optical flow](https://mavlink.io/en/messages/common.html#OPTICAL_FLOW) support in MAVROS

### DIFF
--- a/mavros_extras/src/plugins/optical_flow.cpp
+++ b/mavros_extras/src/plugins/optical_flow.cpp
@@ -1,0 +1,210 @@
+/*
+ * Copyright 2023 Vladislav Chuvarjov
+ *
+ * This file is part of the mavros package and subject to the license terms
+ * in the top-level LICENSE file of the mavros repository.
+ * https://github.com/mavlink/mavros/tree/master/LICENSE.md
+ */
+/**
+ * @brief OpticalFlow plugin
+ * @file optical_flow.cpp
+ * @author Vladislav Chuvarjov <chuvarevanv@gmail.com>
+ *
+ * @addtogroup plugin
+ * @{
+ */
+
+#include <string>
+
+#include "rcpputils/asserts.hpp"
+#include "mavros/mavros_uas.hpp"
+#include "mavros/plugin.hpp"
+#include "mavros/plugin_filter.hpp"
+
+#include "mavros_msgs/msg/optical_flow.hpp"
+#include "sensor_msgs/msg/range.hpp"
+
+namespace mavros
+{
+namespace extra_plugins
+{
+using namespace std::placeholders;      // NOLINT
+
+/**
+ * @brief Optical Flow custom plugin
+ * @plugin optical_flow
+ *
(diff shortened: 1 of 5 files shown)


mavros_extras/src/plugins/optical_flow.cpp
```diff
@@ -0,0 +1,209 @@
...
+    auto int_flow_comp_m_xy = ftf::transform_frame_aircraft_baselink(
+      Eigen::Vector3d(
+        flow.flow_comp_m_x,
+        flow.flow_comp_m_y,
+        0.0));
+    auto int_flow_rate_xy = ftf::transform_frame_aircraft_baselink(
+      Eigen::Vector3d(
+        flow.flow_rate_x,
+        flow.flow_rate_y,
+        0.0));
```
vooon, line 131: After that conversion i'm not sure that we can just keep XY. Maybe better Vector3?

mavros_msgs/msg/OpticalFlow.msg
```diff
@@ -0,0 +1,9 @@
+# OPTICAL_FLOW message data
+
+std_msgs/Header header
+
+int16[2] flow
+float32[2] flow_comp_m
+uint8 quality
+float32 ground_distance
+float32[2] flow_rate
```
vooon, line 9: Usually ROS recommend to use full Vector3 and just don't set Z.

vooon: Well, arrays looks worse than _x/_y.
But, why not to try this:
```
geometry_msgs/Vector3 flow
Vector3 flow_comp_m
Vector3 flow_rate
```

vooon (changes requested):

mavros_msgs/msg/OpticalFlow.msg
Vladislavert: flow in mavlink is int16_t, convert to float?

mavros_extras/src/plugins/optical_flow.cpp
Vladislavert: The z component of the vector is not used in the optical flow algorithm. The function `ftf::transform_frame_aircraft_baselink`, which multiplies the vector by `quaternion_from_rpy(M_PI, 0.0, 0.0)`, it becomes clear that the z component does not contribute to the algorithm. Consequently, it is recommended to preserve only the XY components of the vector and explicitly set the z component to 0.0 during the conversion process. This approach aligns with the requirements of the algorithm and ensures proper functioning.

mavros_msgs/msg/OpticalFlow.msg
vooon: I think yes, easier to use existing and well established type. We can define variety of Vector2, I16Vector2 and so, but for other ros users that anyway would need more and more conversions.

vooon (approved):
LGTM, Thanks!

Issue #1842, opened 2023-03-17
Can't compare times with different time sources - Precision landing


eMrazSVK:
### Issue details

Hello,

I am trying to use precision landing through Mavros with ArduPilot. What I am trying to achieve is to send `PoseStamped` messages to `/mavros/landing_target/pose` topic. Basically I was not able to go very far with this. So the workflow is like this:

- I detect AprilTag from the simulated camera in Gazebo
- Then I forward (publish) the pose of AprilTag detection to `/mavros/landing_target/pose` topic
- Mavros crashes

I looked at these parameters inside `apm_config.yaml`, that might be related, but I am not sure of what to put there:
```
# sys_time
/mavros/**/time:
  ros__parameters:
    time_ref_source: "fcu"  # time_reference source
    timesync_mode: MAVLINK
    timesync_avg_alpha: 0.6 # timesync averaging factor
```
Any help is appreciated!

### MAVROS version and platform

Mavros: ros2
ROS: Foxy
Ubuntu: 20.04

### Autopilot type and version

[ x ] ArduPilot
[  ] PX4

Version: Copter-4.2

### Node logs

> [mavros_node-1] terminate called after throwing an instance of 'std::runtime_error'
[mavros_node-1]   what():  can't compare times with different time sources
[ERROR] [mavros_node-1]: process has died [pid 18491, exit code -6, cmd '/home/edo/ros2_ws/install/mavros/lib/mavros/mavros_node --ros-args -r __ns:=/mavros --params-file /tmp/launch_params_dghd2yaz --params-file /tmp/launch_params_b6a4exat --params-file /tmp/launch_params_6xg2fmis --params-file /tmp/launch_params_95ugyi9g --params-file /tmp/launch_params_9y0f43q2 --params-file /home/edo/ros2_ws/install/mavros/share/mavros/launch/apm_pluginlists.yaml --params-file /home/edo/ros2_ws/install/mavros/share/mavros/launch/apm_config.yaml'].

vooon:
I'm afraid you can't coupe that problem via parameters.
At some point of time cpp std lib started checking time source. We need to review sys_time.cpp and related UAS code.

PR #1889, opened 2023-08-28
Fixed header.stamp conversion in landing target

MKargus0:
This PR proposes the solution for error declared in this [issue](https://github.com/mavlink/mavros/issues/1842).
The main problem was caused by builtin_interfaces::msg::Time to rclcpp::Time conversion. By default it set RCL_ROS_TIME time type, which in turn caused an error when comparing time stamps.


vooon (approved):
LGTM, Thanks! I'm going to merge once CI passes.

Issue #1888, opened 2023-08-28
Local takeoff and landing on ros2


IdoGuzi:
This is only bug and feature tracker, please use it
to report bugs or request features.

----

### Issue details
There are topic for takeoff and landing for global settings (gps: lat, long, alt),
but there are no topic for local takeoff and land (x,y,z).
Those topic are "needed" for offboard/guided modes without GPS.
I'm working on adding the topics to my fork at the moment, 
and I wanted to create a discussion on it to see if there is anything specific that should be done.

Is this feature in progress by someone else?
Is there a reason why those topic does not exist in the first place?
Is there a specific guidelines I should follow while implementing those topics?

I'm expecting to finish testing and to create a PR by the end of this week.

### MAVROS version and platform

Mavros: 2.5.0
ROS: Humble
Ubuntu: 22.04

### Autopilot type and version

[ x ] ArduPilot
[  ] PX4

Version: 4.3.7

IdoGuzi:
created a draft PR for the moment
https://github.com/mavlink/mavros/pull/1890

IdoGuzi:
PR was merged, closing this issue.

PR #1890, opened 2023-08-29
local takeoff and land topics

IdoGuzi:
PR for issue: https://github.com/mavlink/mavros/issues/1888
Adding topics for local takeoff and landing:
`/mavros/cmd/takeoff_local`
`/mavros/cmd/land_local`

Both will take the `CommandLocalTOL` service type (file: mavros_msgs/srv/CommandLocalTOL.srv)

Creating this as draft as it is not tested properly yet, but it will allow others to test if needed,
I'm expecting to finish testing by the end of the week (ardupilot, pixhawk 6c fcu).

Please comment if something else need to be done for the PR to be approved.

Note: on local land (the mavlink message) https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_LAND_LOCAL
I wasn't sure what `Target` means (param 1 of the message), let me know how should it be handled.


mavros_msgs/srv/CommandLocalTOL.srv
```diff
@@ -0,0 +1,12 @@
+#Common type for LOCAL Take Off and Landing
+
+float32 min_pitch	# used by takeoff
+float32 offset    	# used by land (landing position accuracy)
+float32 rate		# speed of takeoff/land in m/s
+float32 yaw		# in radians
+float32 y		# in meters
+float32 x		# in meters
+float32 z		# in meters
```
vooon, line 9: Please use Vector3 instead.

mavros/src/plugins/command.cpp
```diff
@@ -115,11 +116,22 @@ class CommandPlugin : public plugin::Plugin
       std::bind(
         &CommandPlugin::takeoff_cb, this, _1, _2,
         _3), rmw_qos_profile_services_default, srv_cg);
+    takeoff_local_srv =
+      node->create_service<mavros_msgs::srv::CommandLocalTOL>(
```
vooon, line 120: I'm not entirely sure, but probably CommandTOLLocal will be better name, since everywhere 'local' used as a suffix.

vooon (changes requested):

mavros/src/plugins/command.cpp
IdoGuzi: changed new service type name to `CommandTOLLocal` as requested.
marking as resolved.

vooon (approved):
Looks good to merge. Thanks!

vooon (approved):

PR #1904, opened 2023-10-06
Fix errata in GPSRAW.msg

SwiftGust:
I fixed errata in GPSRAW.msg for main branch ROS2
I think this may required to applied across all the branches


vooon (approved):
LGTM. Need to backport to ros1.

PR #1909, opened 2023-10-12
Define parameters for the frame names of the static transforms

mzahana:
This PR solves the issue of duplicate (non unique) static transforms when mavros is used in multi-vehicle SITL setup.
Before this PR, it was not possible to have a proper TF tree where each vehicle has its unique base_link(base_link_frd and odom (odome_ned) frames as it was hard-coded in `mavros_uas` node. Also the `odom` plugin would keep throwing an error because of disconnected trees.

With this PR, the user can define the statice transform frames (base_link, odom, map) as node parameters of the `mavros_node` which is the passed to the `mavros_uas` node. In addition, the `odom` plugin is modified to infer the `_frd` frames directly from the odom/map base_link (or simply the parent and the child frames)m which are defined in the config.yaml file.

With this PR, I am able to have a proper and connected trees in multi-vehicle simulation.

### DIFF
--- a/mavros/include/mavros/mavros_uas.hpp
+++ b/mavros/include/mavros/mavros_uas.hpp
@@ -570,6 +570,10 @@ private:
 
   std::string uas_url;
 
+  std::string base_link_frame_id;
+  std::string odom_frame_id;
+  std::string map_frame_id;
+
   StrV plugin_allowlist;
   StrV plugin_denylist;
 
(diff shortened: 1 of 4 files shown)


vooon (approved):
LGTM, Thanks!

FPSychotic:
Please, any chance to port this to ROS Noetic?

NarmadaBalasooriya:
@mzahana Is there any chance to port this to ROS Noetic? I am using noetic and I have the same issue with using unique frames in multi-uav simulation.

PR #1910, opened 2023-10-18
Removed unused variable

haavarpb:
I stumbled upon a variable that does not seem to be used anywhere in the file. And since it's a private variable it won't be used anywhere else outside the class.


vooon (approved):
Maybe it was used in older versions...

PR #1914, opened 2023-10-24
fix typo wheel_odometry.cpp and mavros_plugins.xml

Moma98:
fix typo


vooon (approved):
XML should be re-generated with cog. Script inside it parses plugin cpp comment and extract definition from it.

But okay, i'll do that. LGTM, Thanks!

PR #1915, opened 2023-10-24
Remove hardcoded namespace from px4 config yaml files

mgiurato:


vooon (approved):
Please include some explanation next time.

UAS node can be run several times with different names, so default mavros::uas::UAS node name "mavros" might be a problem.
At the same time i'm not so sure that it's a good idea to use same config in multi-drone setup...

Issue #1919, opened 2023-11-02
Multiple nodes with the same name created if mavros::uas::UAS is loaded into a component container


ugol-1:
This is only bug and feature tracker, please use it
to report bugs or request features.

----

### Issue details

When the `mavros::uas::UAS` component is loaded into a component container, it creates multiple nodes with the same name as the component manager name. It seems that each plugin node gets the name of the component manager.

Launch file:
```python
from launch import LaunchDescription
from launch.actions import DeclareLaunchArgument
from launch.substitutions import LaunchConfiguration
from launch_ros.actions import ComposableNodeContainer
from launch_ros.descriptions import ComposableNode

def generate_launch_description():
    fcu_url_launch_arg = DeclareLaunchArgument(
        "fcu_url",
        default_value = "/dev/ttyACM0:115200"
    )

    mavros_uas_node = ComposableNodeContainer(
        name = 'MavrosComponentManager',
        namespace = 'mavros',
        package = 'rclcpp_components',
        executable = 'component_container_mt',
        composable_node_descriptions = [
            ComposableNode(
                name = "uas",
                namespace = 'mavros',
                package = 'mavros',
                plugin = 'mavros::uas::UAS',
                parameters = [{
                    "fcu_url": LaunchConfiguration("fcu_url"),
                    "gcs_url": "",
                    "tgt_system": 1,
                    "tgt_component": 1,
                    "fcu_protocol": "v2.0",
                    "plugin_allowlist": [
                        "sys_status",
                    ]
                }],
                extra_arguments = [{
                    "use_intra_process_comms": True
                }]
            )
        ],
    )

    return LaunchDescription([
        fcu_url_launch_arg,
        mavros_uas_node,
    ])
```

### MAVROS version and platform

Mavros: 2.6.0-1jammy.20230912.120519
ROS: Iron
Ubuntu: 22.04

### Autopilot type and version

[  ] ArduPilot
[  ] PX4

Version: 4.3.7

### Node logs

```
[INFO] [launch]: All log files can be found below /home/pi/.ros/log/2023-11-02-19-48-02-226047-malina-2-14196
[INFO] [launch]: Default logging verbosity is set to INFO
[INFO] [component_container_mt-1]: process started with pid [14211]
[component_container_mt-1] [INFO] [1698947283.164913486] [mavros.MavrosComponentManager]: Load Library: /opt/ros/iron/lib/libmavros.so
[...]

PR #1920, opened 2023-11-02
On the way to fixing #1919

ugol-1:


mavros/include/mavros/plugin.hpp
```diff
@@ -110,7 +101,7 @@ class Plugin : public std::enable_shared_from_this<Plugin>
   }
 
 protected:
-  UASPtr uas;                       // uas back link
+  std::weak_ptr<UAS> uas_;                       // uas back link
```
vooon, line 104: What for? Shared ptr good enough.

ugol-1: `shared_ptr` is not good here, because it creates cyclic references. The UAS node holds `shared_ptr`s to plugin nodes. In turn, each plugin node holds a `shared_ptr` to the UAS node. The result is that the UAS node and the plugin nodes never get destroyed.
You can check this by loading the `mavros::uas::UAS` component into a component container and then unloading it. The nodes will still be there. And then load it again, every node will get duplicated.

PR #1921, opened 2023-11-05
fix mav service  call and wp load

sathak93:
Fixes missing call method in cli_push and cli_close in mav  service client.
Fixes typo in mission file


vooon (approved):
LGTM, Thanks!

PR #1923, opened 2023-11-14
[camera plugin] Fix image_index and capture_result not properly filled

bpozzan-archangel:
## Problem Solved
Both `image_index` and `capture_result` fields of the `mavros_msgs::CameraImageCaptured` were left unassigned.

## Proposed Solution
Assign the proper values to them according to the `CAMERA_IMAGE_CAPTURED` message


vooon (approved):
LGTM, Thanks!

PR #1925, opened 2023-11-15
Fix PR 1922 regarding EVENT message

Vicidel:
Just realized I did not commit this fix which fixes CI for PR 1922... Sorry about that, should be good, but could you run the CI again to be sure?


Vicidel:
Well now the `build_target_workspace` is not failing anymore, the one failing is `run_target_test`. Does not seems linked to my changes though. I'll check what I can do here

vooon (approved):
Thanks!

PR #1932, opened 2024-02-08
Removed warning from geometry2 header

ahcorde:
Recently I removed some obsolete headers from geometry2 https://github.com/ros2/geometry2/pull/645. This PR update the code to fix the warning and the buildfarm https://build.ros2.org/view/Rbin_uJ64/job/Rbin_uJ64__mavros__ubuntu_jammy_amd64__binary/257/console


vooon (approved):
I kept `.h` for  foxy. But since all of them gone EoL possibly it's okay to move.

Issue #1381, opened 2020-01-29
[global_position] 'gp_origin' and 'set_gp_origin' mismatching


standmit:
### Issue details
I found some mismatching in **global_position** plugin.
When I send coordinates (LLA) to topic `mavros/global_position/set_gp_origin`, the plugin accepts this coordinates as LLA and sends it to FCU as is.
```cpp
gpo.latitude = req->position.latitude * 1E7;
gpo.longitude = req->position.longitude * 1E7;
gpo.altitude = req->position.altitude * 1E3 + m_uas->ellipsoid_to_geoid_height(&req->position);

UAS_FCU(m_uas)->send_message_ignore_drop(gpo);
```

Then FCU confirms this with GPS_GLOBAL_ORIGIN message containing the same coordinates as LLA. But plugin transforms it to ECEF and sends to topic `mavros/global_position/gp_origin`.
```cpp
g_origin->position.latitude = glob_orig.latitude / 1E7;
g_origin->position.longitude = glob_orig.longitude / 1E7;
g_origin->position.altitude = glob_orig.altitude / 1E3 + m_uas->geoid_to_ellipsoid_height(&g_origin->position);	// convert height amsl to height above the ellipsoid

try {
    /**
     * @brief Conversion from geodetic coordinates (LLA) to ECEF (Earth-Centered, Earth-Fixed)
     * Note: "earth" frame, in ECEF, of the global origin
     */
    GeographicLib::Geocentric earth(GeographicLib::Constants::WGS84_a(), GeographicLib::Constants::WGS84_f());
    earth.Forward(g_origin->position.latitude, g_origin->position.longitude, g_origin->position.altitude,
            g_origin->position.latitude, g_origin->position.longitude, g_origin->position.altitude);

    gp_global_origin_pub.publish(g_origin);
}
```
Topic `mavros/global_position/gp_origin` has type `geographic_msgs/GeoPointStamped` and must contain coordinates in LLA. When I send a request containing LLA, I expect to receive a response containing LLA too in order to compare it and ensure that global_origin has been successfully set. Why **global_position** plugin transforms **global_origin** to ECEF?

### MAVROS version and platform

Mavros: 1.0.0
ROS: Melodic
Ubuntu: 18.04

### Autopilot type and version

[ X ] ArduPilot
[  ] PX4

Version: ArduCopter V4.0.0-dev (8c413d3c)

TSC21:
[...]

PR #1942, opened 2024-04-05
gps_global_origin: remove LLA to ECEF conversion

beniaminopozzan:
GPS_GLOBAL_ORIGIN MAVLink message is published in ROS on the topic `mavros/global_position/gp_origin` with type [geographic_msgs::GeoPointStamped](https://docs.ros.org/en/api/geographic_msgs/html/msg/GeoPointStamped.html).
This message contains https://docs.ros.org/en/api/geographic_msgs/html/msg/GeoPoint.html and expects
> Geographic point, using the WGS 84 reference ellipsoid.

However the messages are published in ECEF coordinates.

This PR removes such conversion matching message requirements and message contents.

FIX https://github.com/mavlink/mavros/issues/1381


vooon (approved):
Unsure, But if that's misleading i'm ok.

PR #1825, opened 2023-02-02
Gimbal protocol v2 plugin

Mark-Beaty:
This plugin has been tested with ROS2 Foxy and Humble and everything appears to be working correctly.  Certain services and topics haven't been observed as working due to limitations of my testing hardware (Freefly Astro) and software (Auterion Simulator), but everything follows the documentation currently available for MAVLink's Gimbal Protocol v2.  I also plan on adding another branch on my fork with the plugin separate from Mavros to enable use as a standalone plugin prior to the PR being approved/incorporated in a release.

PR #1953, opened 2024-05-21
Gimbal protocol v2 plugin - addressing PR comments

Crowdedlight:
Addressing the PR comments from #1825 to help get the work done by Mark-Beaty merged in. 

Copied in his PR description here: 

> This plugin has been tested with ROS2 Foxy and Humble and everything appears to be working correctly. Certain services and topics haven't been observed as working due to limitations of my testing hardware (Freefly Astro) and software (Auterion Simulator), but everything follows the documentation currently available for MAVLink's Gimbal Protocol v2. I also plan on adding another branch on my fork with the plugin separate from Mavros to enable use as a standalone plugin prior to the PR being approved/incorporated in a release.


mavros_extras/mavros_plugins.xml
```diff
@@ -69,6 +69,12 @@ Sends fake GPS from local position estimation source data (motion capture,
 vision) to FCU - processed in HIL mode or out of it if parameter MAV_USEHILGPS
 is set on PX4 Pro Autopilot Firmware; Ardupilot Firmware already supports it
 without a flag set.</description>
+  </class>
+  <class name="gimbal_control" type="mavros::plugin::PluginFactoryTemplate&lt;mavros::extra_plugins::GimbalControlPlugin&gt;" base_class_type="mavros::plugin::PluginFactory">
+    <description>@brief Gimbal Control Plugin
+@plugin gimbal_control
+
+Adds support for Mavlink Gimbal Protocol v2.</description>
```
Crowdedlight, line 77: Original comment by "vooon": 
"I think you haven't used cog.py to re-generate that file?" 

I am not sure how to run it so it effects this file? I can the `./mavros/tools/cogall.sh` but it did not seem to regenerate this file. Could you point me in the right direction to figure out how to regenerate it?

mavros_extras/src/plugins/gimbal_control.cpp
```diff
@@ -0,0 +1,690 @@
...
+    gimbal_manager_camera_track = node->create_service<mavros_msgs::srv::GimbalManagerCameraTrack>(
+      "~/manager/camera_track", std::bind(
+        &GimbalControlPlugin::manager_camera_track,
+        this, _1, _2));
+    
+    // Client used by all services for sending mavros/cmd/command service calls, on a separate callback group to support nested service calls
+    cmdClient = node->create_client<mavros_msgs::srv::CommandLong>("cmd/command", rmw_qos_profile_services_default, cb_group);
+    while (!cmdClient->wait_for_service(std::chrono::seconds(5))) {
+        RCLCPP_ERROR(node->get_logger(), "GimbalControl: mavros/cmd/command service not available after waiting");
+    }
```
Crowdedlight, line 185: Original comment by vooon:
> Locking in the constructor might be a bad idea.
> Maybe better to check for that service later?

As the rest is callbacks based on ros/mavlink message traffic, then you would have to add a check to every callback needing this client, instead of having it in the constructor once. 
The locking here is also set with a timeout of 5s, so worst-case is locked for 5s before moving on. 

If you have a suggestion to how to handle it better, let me know :)

Crowdedlight: edit: I miswrote. I can see it will block here in periods of 5s forever. 
Potential change could be to make it to only try once with that timeout and then abandon the plugin and throw the error, if the service can't be made?

mavros_extras/mavros_plugins.xml
vooon: https://github.com/mavlink/mavros/blob/ros2/cog-requirements.txt

Cogall depends on installed cog.py, probbly in today's world that be easier with pipx, but i haven't exact directions for that.

So in general, you must do like that (beware, not tested!):
```
python3 -m venv ~/ros/.venv
source ~/ros/.venv
pip3 install -r cog-requirements.txt
pip3 install pymavlink
ln mavros_cog.py ~/ros/.venv/lib/python3.11/site-packages/
```

Then
```
cog -cr mavros_extras/mavros_plugins.xml
```

mavros_extras/src/plugins/gimbal_control.cpp
Crowdedlight: I can change to do that, I just thought the constant recreation of the client would be worse? 
As if a client is using the service to set ROI region, or pitchyaw with even just 1hz it is still initializing a client and destroying it per call. 

I could create it in the constructor, but move the check if the service is available to each message using it? And change so it only tries for 5s then just returns error to the caller to let them know? 

Would keep the constructor free from blocking, but still ensure the service exist before calling it, and otherwise return message to caller that the request could not be completed due to the mavros/cmd service not being found?

vooon: You can cache the client, but must create it on request, not in the constructor.

Something like that pseudo:
```cpp
*Client get_cmd_long() {
  mutex Locker();

  if _cache {
    return _cache;
  }

  client = node_->create_client(...);
  // error handling!
  _cache = client;
  return client;  
}
```

Crowdedlight: ofc, that makes sense. I'll do that.

Crowdedlight:
Thanks for fixing the last parts and getting it merged! Much Appricated :)

I had some priority work dumped on me at work and busy weekends so had to postpone it. Sorry about that!

vooon:
@Crowdedlight thank you too. When you have time, please test it, as i didn't have all required things to setup an environment.

Crowdedlight:
> @Crowdedlight thank you too. When you have time, please test it, as i didn't have all required things to setup an environment.

I just tested part of the release in simulation. I found no errors. Gimbal moves as I expect sending the quaternion to it with this topic. 
image

And Using the service to take control over the gimbal with ``/mavros/gimbal_control/manager/configure`` works as well. Due to other commitments I have not had time to go through all the services and topics, but at least these work which should give you full gimbal attitude control. :)

PR #1964, opened 2024-07-07
Fix compile error when compiling with gcc 13

wentasah:
The error is:

    src/plugins/mag_calibration_status.cpp:64:22: error: ‘bitset’ is not a member of ‘std’


vooon (approved):
Please add a blank line after include.

wentasah:
> Please add a blank line after include.

Done.

vooon (approved):
LGTM, Thanks!

PR #1967, opened 2024-07-15
Add parameters for frame_id of static transforms for multi-vehicle systems

EnderMandS:
Similar to [another PR at branch-ros2](https://github.com/mavlink/mavros/pull/1909), this PR solves the problem of static transforms namespace conflicts in multi-vehicle systems by adding parameters. Also dealt with a similar issue with the `lookup_static_transform` function in the odometry plugin.

Users can modify the `frame_id` by adding the following parameter to the launch file :

```xml
<param name="base_link_frame_id" value="$(arg base_link_frame_id)" />
<param name="odom_frame_id" value="$(arg odom_frame_id)" />
<param name="map_frame_id" value="$(arg map_frame_id)" />
```

While changing the launch file also change the corresponding `frame_id` section in the `px4_config.yaml` file :

```yaml
# odom example
odometry:
  fcu:  # change frame_id to vehicle_0/...
    map_id_des: "vehicle_0/map"    # desired parent frame rotation of the FCU's odometry
    odom_child_id_des: "vehicle_0/base_link"    # desired child frame rotation of the FCU's odometry
    odom_parent_id_des: "vehicle_0/odom"
```

### DIFF
--- a/mavros/include/mavros/mavros_uas.h
+++ b/mavros/include/mavros/mavros_uas.h
@@ -444,6 +444,26 @@ public:
 	 */
 	bool cmode_from_str(std::string cmode_str, uint32_t &custom_mode);
 
+	inline void set_base_link_frame_id(const std::string frame_id) {
+		base_link_frame_id = frame_id;
+	}
+	inline void set_odom_frame_id(const std::string frame_id) {
+		odom_frame_id = frame_id;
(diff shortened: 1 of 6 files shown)


vooon (approved):
LGTM, Thanks!

PR #1973, opened 2024-08-06
fix depcrecation errors

dakejahl:
Building mavros (ubuntu 22.04 ros2 humble) fails due to deprecation warnings treated as errors.

### DIFF
--- a/mavros/include/mavros/mavros_uas.hpp
+++ b/mavros/include/mavros/mavros_uas.hpp
@@ -161,7 +161,7 @@ public:
   inline double geoid_to_ellipsoid_height(const T lla)
   {
     if (egm96_5) {
-      return GeographicLib::Geoid::GEOIDTOELLIPSOID * (*egm96_5)(lla->latitude, lla->longitude);
+      return int(GeographicLib::Geoid::GEOIDTOELLIPSOID) * (*egm96_5)(lla->latitude, lla->longitude);
     } else {
       return 0.0;
     }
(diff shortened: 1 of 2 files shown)


mavros/include/mavros/mavros_uas.hpp
```diff
@@ -161,7 +161,7 @@ class Data
   inline double geoid_to_ellipsoid_height(const T lla)
   {
     if (egm96_5) {
-      return GeographicLib::Geoid::GEOIDTOELLIPSOID * (*egm96_5)(lla->latitude, lla->longitude);
+      return int(GeographicLib::Geoid::GEOIDTOELLIPSOID) * (*egm96_5)(lla->latitude, lla->longitude);
```
vooon, line 164: That casting doesn't look correct to me. Geo lib has overloaded operators, so I'm not sure.

dakejahl: the error is about using an enum for the multiplication. The enum is an int
https://github.com/bakercp/ofxGeographicLib/blob/master/libs/GeographicLib/include/GeographicLib/Geoid.hpp#L182-L197

but maybe we need to cast it to T? Not sure... I fixed this real quick without thinking too hard about it

vooon: https://github.com/bakercp/ofxGeographicLib/blob/master/libs/GeographicLib/include/GeographicLib/Geoid.hpp#L308

Since now we have helper func, i think better to replace that with just a call ConvertHeight() (with h=0.0).

vooon (approved):
LGTM, Thanks!

Issue #1960, opened 2024-06-15
mavros_extras plugins not loaded with ROS2 Jazzy on Ubuntu 24.04


igorsgcampos:
### Issue details

I build mavros from source according to the Jazzy instructions (commit b0da849a06eb1a215b9205b92b8cf39c6d7cf88f (HEAD, tag: 2.8.0, origin/ros2, origin/HEAD, ros2)), but upon running the mavros node, it fails to load all plugins in the mavros_extras, although the std_plugins are loaded successfully.

### MAVROS version and platform

Mavros: 2.8.0
ROS: Jazzy
Ubuntu: 24.04

### Autopilot type and version

[  ] ArduPilot
[  ] PX4

Version: N/A

### Node logs

```
[INFO] [1718454301.043908190] [mavros]: Plugin actuator_control created
[INFO] [1718454301.045180490] [mavros]: Plugin actuator_control initialized
[ERROR] [1718454301.076783145] [mavros]: Plugin adsb load exception: Failed to load library /home/igor.campos/ros2_ws/install/mavros_extras/lib/libmavros_extras_plugins.so. Make sure that you are calling the PLUGINLIB_EXPORT_CLASS macro in the library code, and that names are consistent between this macro and your XML. Error string: Could not load library dlopen error: /home/igor.campos/ros2_ws/install/mavros_extras/lib/libmavros_extras_plugins.so: undefined symbol: _ZTIN4YAML13BadConversionE, at ./src/shared_library.c:99
[INFO] [1718454301.102932325] [mavros]: Plugin altitude created
[INFO] [1718454301.103591558] [mavros]: Plugin altitude initialized
[ERROR] [1718454301.121035135] [mavros]: Plugin cam_imu_sync load exception: Failed to load library /home/igor.campos/ros2_ws/install/mavros_extras/lib/libmavros_extras_plugins.so. Make sure that you are calling the PLUGINLIB_EXPORT_CLASS macro in the library code, and that names are consistent between this macro and your XML. Error string: Could not load library dlopen error: /home/igor.campos/ros2_ws/install/mavros_extras/lib/libmavros_extras_plugins.so: undefined symbol: _ZTIN4YAML13BadConversionE, at ./src/shared_library.c:99
[...]

Issue #1958, opened 2024-06-04
ros2-jazzy no apt package available


julled:
### Issue details

For Ros2 jazzy i cant install mavros via apt. When will it be released?

### MAVROS version and platform

Mavros: 2.7.0
ROS: ros2-jazzy
Ubuntu: 24

### Autopilot type and version

[ x] ArduPilot
[  ] PX4

Version: ?3.7.1?

### Node logs
x
### Diagnostics

```
> apt-cache search mavros 
ros-jazzy-mavros-msgs - mavros_msgs defines messages for MAVROS.
ros-jazzy-mavros-msgs-dbgsym - debug symbols for ros-jazzy-mavros-msgs
ros-rolling-mavros-msgs - mavros_msgs defines messages for MAVROS.
ros-rolling-mavros-msgs-dbgsym - debug symbols for ros-rolling-mavros-msgs
```

### Check ID

x

PR #1988, opened 2024-09-14
Add dependency on yaml-cpp in mavros_extras

evan-palmer:
This PR resolves the compilation errors described in #1960 by adding the `yaml-cpp` dependency to `mavros_extras` for ROS 2 Jazzy and Rolling. These changes also aid in closing #1958 (once the build is updated to support those releases). 

Tridge has already made the [relevant changes](https://github.com/ArduPilot/pymavlink/blob/533161c47995ba47f5548636eed37a47e91072c0/generator/mavgen.py) to pymavlink to support Python 3.12, which is included in the latest version pointed to by [mavlink](https://github.com/mavlink/mavlink). If this looks good to be released for Jazzy and Rolling, I can submit an additional PR with a section in the README pointing to the Jazzy build status.


vooon (approved):
LGTM, Thanks!

PR #2005, opened 2024-10-20
Update Manual Control to support `6dof`

InvincibleRMC:
Tested on my pixhawk running ArduSub by reading /mavros/rc/out.

### DIFF
--- a/mavros/src/plugins/manual_control.cpp
+++ b/mavros/src/plugins/manual_control.cpp
@@ -97,7 +97,7 @@ private:
     msg.z = req->z;
     msg.r = req->r;
     msg.buttons = req->buttons;
-    
+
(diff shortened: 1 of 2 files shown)


vooon (approved):
Overall looks good. Except I'm not sure about aux, maybe better `aux[6]`. But at the same time command params are after mavlink...

InvincibleRMC:
> Overall looks good. Except I'm not sure about aux, maybe better `aux[6]`. But at the same time command params are after mavlink...

Let me know if you for certain want me to make it `aux[6]` or if it is fine the way it is. I do not have a strong preference.

vooon:
I don't have strong preference on array, so it's okay to stick to mavlink definition.

Issue #1997, opened 2024-09-30
No input from RC chanells (mavros/rc/in)


r8d8:
This is only bug and feature tracker, please use it
to report bugs or request features.

----

### Issue details

Mavros doesn't post any rc channel values in /mavros/rc/in topic
Ardupilot's Mavlink Inspectors shows that data is present.


### MAVROS version and platform

Mavros: 2.8
ROS: Humble
Ubuntu: 22.04


### Autopilot type and version

[*] ArduPilot

Version: 4.6.0-dev
Plane


### Node logs

```
[...]

PR #2006, opened 2024-10-25
Fix RC channels count being  zero, when channels values exists

r8d8:
Mission Planner don't ever set channel count. 
Mavros skips all RC values in this case
Refers #1997


vooon:
Closed as no activity

PR #2011, opened 2024-12-18
Support specified frames of reference in landing target module

pryre:
This change implements a check to support the set landing target frame of reference when using the "~/landing_target/raw" interface. 

* The transformations for the `MAV_FRAME::LOCAL_NED` case are the same as the previous implementation
* The transformations for the `MAV_FRAME::BODY_FRD` case have been derived by myself, and seem to work as expected under my testing. A cross-check is probably a good idea. 

One issue with this commit is in regards to handling the case where a use does not want to use the position data at all, and in doing so, is likely to not set the frame parameter. It is expected that this would result in a value of "0" being set, which aligns to MAV_FRAME::GLOBAL. This is not a valid setting for any autopilots I'm aware of, so I've used this is a selector for whether to use the position data or not.

Ideally this would be addressed by exposing "position_valid" in the `mavros_msgs/LandingTarget`, and going from there. An alternative could be to check if a valid quaternion is set.

Another contentious point is the use of MAV_FRAME to interpret the frame from the message. This does not align with the enum in `mavros_msgs/LandingTarget`, however, I don't think this enum is well-aligned with either ROS options (which a user's perspective would be) or the MavLink options, which have the same names, but with different values for some.

This could all probably be addressed in this pull request, I was just hesitant to break any message functionality. without some input. Ideas welcome!

### DIFF
--- a/mavros_extras/src/plugins/landing_target.cpp
+++ b/mavros_extras/src/plugins/landing_target.cpp
@@ -461,22 +461,55 @@ private:
     Eigen::Affine3d tr;
     tf2::fromMsg(req->pose, tr);
 
-    /** @todo these transforms should be applied according to the MAV_FRAME */
-    auto position = ftf::transform_frame_enu_ned(Eigen::Vector3d(tr.translation()));
-    auto orientation = ftf::transform_orientation_enu_ned(
-      ftf::transform_orientation_baselink_aircraft(Eigen::Quaterniond(tr.rotation())));
-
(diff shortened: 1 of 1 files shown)


pryre:
I've just realised that this patch should also include similar functionality when receiving from the drone on `L385`:
```
    /** @todo these transforms should be applied according to the MAV_FRAME */
```

I will update the pull request with these additions shortly.

vooon (approved):
LGTM, Thanks!

PR #2023, opened 2025-03-11
Allow disabling the router

Ryanf55:
# Purpose

Allow disabling the mavlink router in the mavros_node. It's not always necessary.

### DIFF
--- a/mavros/launch/apm.launch
+++ b/mavros/launch/apm.launch
@@ -10,6 +10,7 @@
 	<arg name="fcu_protocol" default="v2.0" />
 	<arg name="respawn_mavros" default="false" />
 	<arg name="namespace" default="mavros"/>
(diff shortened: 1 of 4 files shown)


vooon (changes requested):
What for? Just load `mavros::uas::UAS` into generic runner then.

Ryanf55:
> What for? Just load `mavros::uas::UAS` into generic runner then.

Are you saying I should write my own node? I would like to avoid duplicating the code in mavros_node.cpp.

vooon:
I'm saying that the UAS is the Composable node, one of the core concepts of ROS2. It's a bit similar to nodelets from ROS1.
And you don't have to write any bit of code to load it.

https://docs.ros.org/en/foxy/Tutorials/Intermediate/Composition.html

vooon:
Close as no activity

PR #2028, opened 2025-03-24
Complete Local Setpoint implementation

Dat-Bois:
DRAFT

An initial rough implementation for local setpoints allowing for pos, vel, accel, yaw, and yaw rate for only the localNED frame. 

Issues:
 - Topic is not populated for some reason (HELP NEEDED)
 - Need to shift over to Extra Plugins rather than in std plugins
 - Use COG for generation rather than by hand (should happen after shifting to mavros_extras).
 
Right now my main goal is to just get this simple plugin to work and then I can iron it out to follow the proper guidelines. I used the Setpoint Accel plugin as my reference. However, when running, it doesn't show up with ros2 topic list. It also doesn't have the created/initialized message for the plugin when starting mavros.


Dat-Bois:
@vooon Could you take a brief glance at what might be wrong with the topic not populating?

Dat-Bois:
Nevermind, got cog to work and that fixed whatever hidden issue there was. Closing this PR for now and will re-open when its more cleaned up.

PR #2033, opened 2025-04-23
Address Warnings in `Rolling`

InvincibleRMC:


mavros/src/plugins/setpoint_attitude.cpp
```diff
@@ -16,10 +16,15 @@
  * @{
  */
 
-#include <message_filters/subscriber.h>
-#include <message_filters/synchronizer.h>
-#include <message_filters/sync_policies/approximate_time.h>
-
+#if __has_include(<message_filters/subscriber.hpp>)
```
vooon, line 19: Hm, looks interesting, haven't seen that macro...

vooon (approved):
LGTM, thanks!

PR #2049, opened 2025-07-11
Update ArduPlane mode 14 from LAND to AVOID_ADSB.

### LINKED ISSUE #2048: Arduplane LAND is now AVOID_ADSB
Since ArduPlane-3.7.0, mode 14, previously assigned to LAND, has changed to AVOID_ADSB.
[V3.7.0](https://github.com/ArduPilot/ardupilot/blob/ArduPlane-3.7.0/ArduPlane/defines.h#L61) 
[Current version](https://github.com/ArduPilot/ardupilot/blob/master/ArduPlane/mode.h#L52)

This does not match the stringify for ArduPlane in mavros lib, for both [ROS2](https://github.com/mavlink/mavros/blob/ros2/mavros/src/lib/uas_stringify.cpp#L49) and [ROS1](https://github.com/mavlink/mavros/blob/1.20.1/mavros/src/lib/uas_stringify.cpp#L45), and could result in unwanted behaviors if the user asks for LAND mode using mavros, which would translate to a request for the flight controller to change to AVOID_ADSB mode.

FAC94:
Fix #2048

### DIFF
--- a/mavros/src/lib/uas_stringify.cpp
+++ b/mavros/src/lib/uas_stringify.cpp
@@ -42,7 +42,7 @@ static const cmode_map arduplane_cmode_map{{
 	{ 10, "AUTO" },
 	{ 11, "RTL" },
 	{ 12, "LOITER" },
-	{ 14, "LAND" },		// not in list
+	{ 14, "AVOID_ADSB" },	// renamed, prev name: LAND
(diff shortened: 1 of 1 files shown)


vooon (approved):
LGTM, Thanks!

Could you please make a PR for ROS2?

PR #2053, opened 2025-08-21
[Bugfix] bad_weak_ptr from Router when starting as composable node in kilted

amarburg:
Testing with current `ros2` head under Kilted.   Starting just Router as a composable node with (as an example):

```python
from launch import LaunchDescription
from launch_ros.actions import ComposableNodeContainer
from launch_ros.descriptions import ComposableNode

def generate_launch_description():

    """Generate launch description for MAVROS composable node."""
    container = ComposableNodeContainer(
        name="mavros_container",
        namespace="",
        package="rclcpp_components",
        executable="component_container_mt",
        composable_node_descriptions=[
            ComposableNode(
                package="mavros",
                plugin="mavros::router::Router",  # Or the specific MAVROS component you need
                name="mavros_router",
                parameters=[
                    {"fcu_urls": ["tcp://localhost:6777"]},
                    {"fcu_protocol": "v2.0"},
                    {"uas_urls": ["/uas1"]},
                    
                ],
                extra_arguments=[{"use_intra_process_comms": True}],
            )
        ],
        output="screen",
    )

    return LaunchDescription([container])
```

throws an exception:

```
[component_container_mt-1] [INFO] [1755795754.777550090] [mavros_router]: Requested to add endpoint: type: 0, url: tcp://192.168.2.2:6777
[component_container_mt-1] [ERROR] [1755795754.781504411] [mavros_container]: Component constructor threw an exception: bad_weak_ptr
[ERROR] [launch_ros.actions.load_composable_nodes]: Failed to load node 'mavros_router' of type 'mavros::router::Router' in container '/mavros_container': Component constructor threw an exception: bad_weak_ptr
```

This can be traced back to `Router::on_set_parameters_cb` being called before the constructor is finished, leading to a `shared_from_this` deep in setting up endpoints.

This pull request places the call to `add_on_set_parameters_callback` and parameter setup in a timed functional called 10ms after completion of the constructor, following the design pattern used in [mavros_uas.cpp](https://github.com/amarburg/mavros/blob/b0da849a06eb1a215b9205b92b8cf39c6d7cf88f/mavros/src/lib/mavros_uas.cpp#L88)

Tested under Kilted, I am able to load a router node and multuiple UAS nodes in a single component container.

### DIFF
--- a/mavros/include/mavros/mavros_router.hpp
+++ b/mavros/include/mavros/mavros_router.hpp
@@ -183,16 +183,7 @@ public:
     //
     // Repeat the pattern from mavros_uas of using a delay timer
     startup_delay_timer = this->create_wall_timer(
-      10ms, [this]() {
-        RCLCPP_WARN(get_logger(), "In timer callback");
-        startup_delay_timer->cancel();
-        set_parameters_handle_ptr =
-        this->add_on_set_parameters_callback(std::bind(&Router::on_set_parameters_cb, this, _1));
-
-        this->declare_parameter<StrV>("fcu_urls", StrV());
-        this->declare_parameter<StrV>("gcs_urls", StrV());
-        this->declare_parameter<StrV>("uas_urls", StrV());
-      });
+      10ms, std::bind(&Router::initialize_parameters, this));
   }
 
   void route_message(Endpoint::SharedPtr src, const mavlink_message_t * msg, const Framing framing);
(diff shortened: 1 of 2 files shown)


vooon:
Well i actually see those timed calls as a nasty workaround. Unfortunately i'm not good at modern C++ to find better solution.

vooon:
Huh, looks like rolling finally dropped `.h` headers. Need to fix that later...

vooon:
Also router test needs to be adjusted - now it fails because of that delayed init.
That probably not the best way, but easiest probably to add condition variable and a method like `initialized()`.

vooon:
@amarburg , well, now that's not often, but probably i'll do release at September 9'th.

PR #2064, opened 2025-09-13
Fix: yaml_cpp_vendor linking in mavros_extras

ArielSulton:
## Summary
Fixes yaml_cpp_vendor linking issue in mavros_extras CMakeLists.txt

## Changes
- Updated CMake configuration in `mavros_extras/CMakeLists.txt` to properly handle
yaml_cpp_vendor target for older ROS2 versions
- Ensures compatibility across different ROS2 distributions

## Testing
- [✓] Built successfully with `colcon build`
- [✓] Code style checked with `ament_uncrustify --reformat`
- [✓] Tested on ROS2 Humble

## Related Issues
Fixes build/linking issues with yaml-cpp vendor package on pre-Jazzy ROS2 distributions

## Type of Change
- [✓] Bug fix (non-breaking change which fixes an issue)
- [✗] New feature
- [✗] Breaking change
- [✗] Documentation update

### DIFF
--- a/mavros/include/mavros/mavros_router.hpp
+++ b/mavros/include/mavros/mavros_router.hpp
@@ -186,10 +186,11 @@ public:
     // Delay parameter callback initialization because
     // add/del endpoints calls have to use shared_from_this(),
     // which cannot be used before we leave the constructor.
-    startup_delay_timer = this->create_wall_timer(10ms, [this]() {
-          this->startup_delay_timer->cancel();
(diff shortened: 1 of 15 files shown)


vooon (approved):
LGTM, but in that case no need to run ament_uncrustirfy.
Actually it's even bad - because they have changed uncrustify config in Jazzy,
so code style linting became incorrect for current releases.

PR #2072, opened 2025-10-06
Fixing TF timestamps to use system time in landing_target plugin

gm14:
Fixes bug where default timestamp from TF is initialized by default as RCL_ROS_TIME but is getting compared to RCL_SYSTEM_TIME in send_landing_target, causing a crash. 

Tested ROS2 version: jazzy. Tested in simulation environment.


vooon (approved):
LGTM, Thanks!

Issue #2068, opened 2025-09-17
ROS2: Documentation book


vooon:
Ros wiki is outdated and not really helpful for a new people to find info.
Need to make something to generate plugin documentation.

peakyquest:
I agree — the ROS Wiki is definitely outdated and hard to navigate for new users. How can we help with generating the plugin documentation?

peakyquest:
the MAVROS documentation could definitely benefit from a more structured, Ardupilot-style approach. It shouldn’t just list topics and services, but also include sections like installation, getting started, configuration, and example demos. That way, users can better understand how to set things up correctly instead of running into common issues like “this topic isn’t publishing.” A well-organized doc would make onboarding and troubleshooting much smoother.

vooon:
While i agree that examples should be there, i do not have one.
So at minimum i want to have ROS API documented, that already be very handy.

peakyquest:
I can work on the examples. I have few examples which I made for my self. Can be useful for the other users aswell.

peakyquest:
I'm considering whether it would be appropriate to open a pull request to contribute MAVROS-related examples, specifically focusing on services and topics. There are two potential approaches: one is to start with an initial set of essential examples by creating a mavros_examples package and then encourage community contributions by opening “good first issues” for additional examples. The other is to wait until a complete and comprehensive set of examples is ready before merging. The purpose of the examples would be to address common pain points—particularly misunderstandings around how specific MAVROS topics and services behave or are expected to function. These examples would serve as clear reference implementations to help users avoid frequent mistakes. I’d appreciate feedback on which direction would be more suitable for the project, and I’m happy to proceed with either.

vooon:
I think `mavros_examples` would be good, similar to turtlebot or other study bots, where you can run some common examples.

peakyquest:
I wanted to confirm the right structure for documentation.

Should we create a **`mavros_examples`** package that includes example nodes along with a **README** describing each example?  
And in the **`mavros/docs`** folder, should we add separate **`.md` files** where all the topics and services are described in detail, with a short introduction for each?

vooon:
I think better to to describe each example on `docs/examples/` so they will be available here https://mavros.readthedocs.io/en/latest/

Package itself must have minimal readme, just describing what's that package is about.

vooon:
Well, llm's nowaday is a quite good at coding. Finally got api extractor done (but actually feels like simply writing docs bee much more cheaper in terms of token use...).

PR #2073, opened 2025-10-12
added support for SIM_STATE message

zekesarosi:
We add support for the SIM_STATE message: https://mavlink.io/en/messages/common.html#SIM_STATE.
This message is helpful for developing autonomous flight modes and companion computer software.

### DIFF
--- a/mavros_extras/src/plugins/sim_state.cpp
+++ b/mavros_extras/src/plugins/sim_state.cpp
@@ -1,0 +1,177 @@
+/**
+ * @brief SIM_STATE plugin
+ * @file sim_state.cpp
+ * @author Zeke Sarosi
+ *
+ * @addtogroup plugin
+ * @{
+ */
+/*
+ * Copyright 2025 Zeke Sarosi
+ * 
+ * This file is part of the MAVROS package and subject to the license terms
+ * in the top-level LICENSE file of the MAVROS repository.
+ */
+
+#include "rcpputils/asserts.hpp"
+#include "mavros/mavros_uas.hpp"
+#include "mavros/plugin.hpp"
+#include "mavros/plugin_filter.hpp"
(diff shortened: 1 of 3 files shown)


vooon (approved):
LGTM, Tnaks!

But could you please add some more documentation into plugin brief and at each publisher creation?
Format is not firm yet, i just need something which i can reword later for #2068.

mavros_extras/src/plugins/sim_state.cpp
```diff
@@ -0,0 +1,183 @@
...
+namespace mavros
+{
+namespace extra_plugins
+{
+using namespace std::placeholders;
+
+/**
+ * @brief SIM_STATE plugin.
+ * @plugin sim_state
+ */
```
vooon, line 49: Please add explanation what's it for here.

mavros_extras/src/plugins/sim_state.cpp
```diff
@@ -0,0 +1,183 @@
...
+class SimStatePlugin : public plugin::Plugin
+{
+public:
+  explicit SimStatePlugin(plugin::UASPtr uas_)
+  : Plugin(uas_, "sim_state")
+  {
+    attitude_pub = node->create_publisher<sensor_msgs::msg::Imu>(
```
vooon, line 58: Please add description comment before each pub/sub/srv creation. I think to parse cpp to extract ROS API definition from `node->create_*()`.

zekesarosi: thanks I added some brief comments, is this what you had in mind?

vooon: Thanks!

vooon (approved):
LGTM, Thanks!

PR #2085, opened 2025-12-11
Fix overflow in ros_time_from_ns()

paulerikf:
## Summary
This pull request resolves an issue where MAVROS may throw a runtime exception when constructing `rclcpp::Time` under certain timing conditions. The problem occurs when the FCU transitions from boot-time timestamps to GPS epoch time, leading to extremely large timestamp values that overflow the `(sec, nsec)` constructor and appear negative.

```
terminate called after throwing an instance of 'std::runtime_error' what(): 
  cannot store a negative time point in rclcpp::Time Aborted (core dumped) 
  make: *** [Makefile:58: run] Error 134
```

## Root Cause
The previous implementation decomposed a nanosecond timestamp into `(sec, nsec)`:

```cpp
rclcpp::Time(stamp_ns / 1000000000UL, stamp_ns % 1000000000UL);
```

When combined timestamps exceeded the `int32_t` range for seconds, the constructor interpreted the result as negative, causing `rclcpp::Time` to throw an exception. This scenario commonly arises when:
- The FCU operates initially on monotonic boot-time.
- MAVROS applies a large offset to approximate epoch time.
- The FCU later switches to GPS time.

## Fix
The implementation now uses the `rclcpp::Time(int64_t nanoseconds)` constructor, which safely handles large 64-bit timestamps:

```cpp
return rclcpp::Time(static_cast<int64_t>(stamp_ns));
```

## Added Test
Added a test to `test_uas.cpp` which simulates:
- A large boot-time offset (~1.7e9 seconds).
- A subsequent shift to GPS time.

### DIFF
--- a/mavros/src/lib/uas_timesync.cpp
+++ b/mavros/src/lib/uas_timesync.cpp
@@ -21,9 +21,7 @@ using namespace mavros::uas;    // NOLINT
 
 static inline rclcpp::Time ros_time_from_ns(const uint64_t stamp_ns)
 {
-  return rclcpp::Time(
-    stamp_ns / 1000000000UL,                            // t_sec
-    stamp_ns % 1000000000UL);                           // t_nsec
+  return rclcpp::Time(static_cast<int64_t>(stamp_ns));
 }
 
 rclcpp::Time UAS::synchronise_stamp(uint32_t time_boot_ms)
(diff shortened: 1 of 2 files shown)


vooon:
Please apply uncrustify - https://github.com/mavlink/mavros/blob/ros2/CONTRIBUTING.md nr.7.

vooon (approved):
I don't remember why sec/nsec used. Maybe old release do not support plain nanoseconds or dimension wasn't specified...

Anyway, LGTM, Thanks!

paulerikf:
Thanks! Apologies for forgetting uncrustify!

PR #2089, opened 2025-12-18
Fix:: Reduce CPU usage by enabling intra-process communication and b…

faizan-185:
Reduced CPU usage by enabling intra-process communication and by switching to sensor Qos.

Without intra-process comms, messages traverse the RMW and DDS layers even when publisher and subscriber are in the same process. Also there is a chance of aggressive default QoS on internal bus.

1. uncomment this -> options.use_intra_process_comms(true);
2. Use SensorDataQoS instead of Default Qos.
3. I have also reduced the number of threads, but 'std::thread::hardware_concurrency()' can be configurable by using an env or something.

### DIFF
--- a/mavros/src/lib/mavros_router.cpp
+++ b/mavros/src/lib/mavros_router.cpp
@@ -469,8 +469,7 @@ std::pair<bool, std::string> ROSEndpoint::open()
   }
 
   try {
-    auto qos = QoS(
-      1000).best_effort().durability_volatile();
+    auto qos = rclcpp::SensorDataQoS();
     this->source =
       nh->create_publisher<mavros_msgs::msg::Mavlink>(
       utils::format(
--- a/mavros/src/lib/uas_executor.cpp
+++ b/mavros/src/lib/uas_executor.cpp
@@ -29,7 +29,7 @@ UASExecutor::UASExecutor(const rclcpp::ExecutorOptions & options)
 size_t UASExecutor::select_number_of_threads()
 {
   // return std::max<size_t>(16, std::min<size_t>(std::thread::hardware_concurrency(), 4));
-  return std::clamp<size_t>(std::thread::hardware_concurrency(), 4, 16);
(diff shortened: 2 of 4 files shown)


mavros/src/lib/mavros_router.cpp
```diff
@@ -469,8 +469,7 @@ std::pair<bool, std::string> ROSEndpoint::open()
   }
 
   try {
-    auto qos = QoS(
-      1000).best_effort().durability_volatile();
+    auto qos = rclcpp::SensorDataQoS();
```
vooon, line 472: What for? It's queue size is 5, which may be small.

mavros/src/mavros_node.cpp
```diff
@@ -29,7 +29,7 @@ int main(int argc, char * argv[])
   rclcpp::executors::MultiThreadedExecutor exec(rclcpp::ExecutorOptions(), 2);
 
   rclcpp::NodeOptions options;
-  // options.use_intra_process_comms(true);
+  options.use_intra_process_comms(true);
```
vooon, line 32: Have you tested it? Does it work if the router and uas in the different processes (i mean if that scenario continue to work)?

mavros/src/lib/uas_executor.cpp
```diff
@@ -29,7 +29,7 @@ UASExecutor::UASExecutor(const rclcpp::ExecutorOptions & options)
 size_t UASExecutor::select_number_of_threads()
 {
   // return std::max<size_t>(16, std::min<size_t>(std::thread::hardware_concurrency(), 4));
-  return std::clamp<size_t>(std::thread::hardware_concurrency(), 4, 16);
+  return std::clamp<size_t>(std::thread::hardware_concurrency(), 2, 8);
```
vooon, line 32: I'd like to introduce UAS_MAXPROC env var, but it must be >= 2.

PR #2117, opened 2026-02-26
imu: fix missing hPa to Pa conversion in HIGHRES_IMU handler

otooleat:
abs_pressure and diff_pressure in HIGHRES_IMU messages are in hPa, but sensor_msgs/FluidPressure.fluid_pressure is in Pascals. MILLIBAR_TO_PASCAL  was already defined but not applied, resulting in pressure values 100 times too small.

Also adjusted the handle_scaled_pressure to use MILLIBAR_TO_PASCAL for consistency instead of 100 magic number.


vooon (approved):
LGTM, Thanks!

PR #2139, opened 2026-03-10
UAS class, stop thread executor before plugin and subscriptions begin tearing down

JonasPerolini:
`exec_spin_thd` deleter calls `exec.cancel() ` but `exec_spin_thd` is declared before `exec` in `mavros_uas.hpp`. Since the destructor is called in the reverse order in which they appear in the class declaration (https://learn.microsoft.com/en-us/cpp/cpp/destructors-cpp?view=msvc-170), `exec` is destroyed first, then the deleter runs, and calls `exec.cancel()`. 

Solution: call `exec_spin_thd.reset();` in the destructor of `UAS` (and change the order in the hpp)


vooon (approved):
Good catch!

PR #2148, opened 2026-03-13
Odom plugin, reject observations with invalid static transforms

JonasPerolini:
Inside the Odom plugin, if `lookup_static_transform` failed, we call `.linear()` and `.inverse()` using an uninitialized `Eigen::Affine3d` object

Solution: reject the observation if the frame transformation is invalid.

Example:  
- `ros2 topic pub -r 10 /CONTAINER_NAME/mavros/odometry/out nav_msgs/msg/Odometry "{header: {frame_id: 'missing_map'}, child_frame_id: 'missing_base_link'}"
`

Will now output: 
- `[ERROR] ODOM: static transform missing_map -> missing_map_ned unavailable: "missing_map_ned" passed to lookupTransform argument target_frame does not exist.`

and then early return


vooon (approved):
LGTM, Thanks!

PR #2149, opened 2026-03-13
Sys status: add mutex to M_VehicleInfo vehicles unordered map

JonasPerolini:
`M_VehicleInfo vehicles` is an `unordered_map` which can be accessed concurrently by:
- `handle_heartbeat()` and `handle_autopilot_version()`, which create/update entries.
- Disconnect handling via `connection_cb(false)`, triggered from the heartbeat timeout path, which clears the map.
- The `vehicle_info_get` service callback, which iterates or looks up entries and copies `VehicleInfo` objects into the response.

`vehicle_info_get_cb()` runs in the plugin service callback group, while MAVLink handlers are invoked from the UAS multithreaded executor.

Solution: shared mutex.

- Unique lock for write calls: `handle_heartbeat()`, `handle_autopilot_version()`, `connection_cb(false)` / `vehicles.clear()`, and the helper that may insert into the map (`find_or_create_vehicle_info_unlocked()` is only called while holding that unique lock).

- Shared lock for read calls: `vehicle_info_get_cb()`.

To reproduce the issue: 
- Set `"conn_timeout", 0.05` to force _CON: Lost connection, HEARTBEAT timed out_
- Add `std::this_thread::sleep_for(10ms);` right after: 
   -  `auto it = find_or_create_vehicle_info(msg->sysid, msg->compid);` inside of `handle_heartbeat`
   -  `for (const auto & got : vehicles) {` inside `vehicle_info_get_cb`
- Run: 

Result: 

```
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
[ros2run]: Aborted
```

or: 

```
[1773437058.811066636] [WARN] CON: Lost connection, HEARTBEAT timed out.
[1773437058.816088957] [INFO] IMU: Attitude quaternion IMU detected!
[ros2run]: Segmentation fault
```

### DIFF
--- a/mavros/src/plugins/sys_status.cpp
+++ b/mavros/src/plugins/sys_status.cpp
@@ -20,6 +20,8 @@
 #include <vector>
 #include <unordered_map>
 #include <memory>
+#include <mutex>
+#include <shared_mutex>
 #include <utility>
 
 #include "rcpputils/asserts.hpp"
@@ -678,6 +680,7 @@ private:
 
   using M_VehicleInfo = std::unordered_map<uint16_t, mavros_msgs::msg::VehicleInfo>;
   M_VehicleInfo vehicles;
+  mutable std::shared_mutex vehicles_mutex;
 
   /* -*- mid-level helpers -*- */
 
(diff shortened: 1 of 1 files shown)


vooon (approved):
LGTM, Thanks!

PR #2150, opened 2026-03-13
Global position: handle ECEF conversion failures

JonasPerolini:
`global_position` continues after ECEF conversion failures and either uses an uninitialized `map_point` or marks the map origin initialized with invalid/stale origin data, leading to bad local odometry/TF output.

Solution: stop publishing derived local odometry/TF when fix-to-ECEF conversion fails, and make home-origin updates only after a successful conversion.


vooon (approved):
LGTM, Thanks!

PR #2160, opened 2026-03-26
fix: std::future_error in command.cpp when the set_value() promise called twice

Lexciese:
A small fix and workaround when the promise.set_value() in handle_command_ack is called twice. The proposed method is already tested on my system and with colcon test.  MAVROS can continue the execution of the plugin.


vooon (approved):
LGTM

PR #2161, opened 2026-03-26
Fix missing semicolon in tools/plugin_doc_extract.cpp

MathewHobson03:
Fixes a compilation error caused by a missing semicolon in tools/plugin_doc_extract.cpp:

if (api.plugin.empty()) {return PluginApi{}}

Updated to:
if (api.plugin.empty()) {return PluginApi{};}

This caused build failure with GCC.


vooon:
Actually i want to refactor it. I wanted to use clang to parse the code, but eventually steered to bunch of regex.

MathewHobson03:
Sounds good. My change should only touch tools/plugin_doc_extract.cpp. The mavros-router-test failure appears to originate from upstream.

vooon (approved):
LGTM, Thanks!

PR #2179, opened 2026-04-03
bugfix: ftp reads return empty

### LINKED ISSUE #1928: MAVROS FTP read from PX4 1.14.1 always return 0 length data.
This is only bug and feature tracker, please use it
to report bugs or request features.

----

### Issue details

mavros ftp READ always read 0 length

Mavros: latest
ROS: noetic
Ubuntu: 20.4

[  ] ArduPilot
[ X ] PX4

Version: 1.14.1

### Node logs

```

```

### Diagnostics

```
In mavros/src/plugins/ftp.cpp send_read_command() it has,
req.header()->size = 0 /* FTPRequest::DATA_MAXSZ */;

But the px4 mavlink_ftp.cpp changed from

int bytes_read = ::read(_session_info.fd, &payload->data[0], kMaxDataLength);

to

int bytes_read = ::read(_session_info.fd, &payload->data[0], payload->size);

So FTP read always return 0 length on px4 1.14.1
```

### Check ID

```
```

tonygoldman:
Fixes #1928

Recent versions of PX4 and ArduCopter read `payload->size` bytes instead of `kMaxDataLength`.
Mavros was sending `size = 0`, causing all reads to return empty.

* Set `req.header()->size` to the actual bytes requested
* Guard the read loop against a redundant trailing request when `read_size` is an exact multiple of `DATA_MAXSZ`

### DIFF
--- a/mavros/src/plugins/ftp.cpp
+++ b/mavros/src/plugins/ftp.cpp
@@ -595,7 +595,7 @@ private:
 
     read_buffer.insert(read_buffer.end(), req.data(), req.data() + bytes_to_copy);
 
-    if (bytes_to_copy == FTPRequest::DATA_MAXSZ) {
+    if (read_buffer.size() < read_size && hdr->size == FTPRequest::DATA_MAXSZ) {
       // Possibly more data
       read_offset += bytes_to_copy;
       send_read_command();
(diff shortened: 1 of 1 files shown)


vooon (approved):
LGTM, Thanks!

PR #2185, opened 2026-04-17
fix: use dedicated callback group for blocking services

Luka-Filipovic:
Fixes a bug where blocking services would prevent timeout callbacks from executing as they were all part of the same `MutuallyExclusiveCallbackGroup`. This change moves all services to a dedicated `srv_cg` callback group as was already done in `command` plugin.


vooon (approved):
LGTM, Thanks!

vooon:
CI green, good to go. Hvala!

PR #2189, opened 2026-05-11
mavros: make UAS executor threads configurable

### LINKED ISSUE #2031: CPU usage 4x higher in ROS 2 than ROS 1
Hello,

When testing the same MAVROS set of plugins and stream rates, we observe MAVROS using 4x the CPU usage of ROS1.
Is this expected, or are there any mitigations?

Best,
Ryan

jacog2:
## Summary

Adds a `MAVROS_UAS_EXECUTOR_THREADS` environment override for the UAS plugin executor worker count.

This gives ROS 2 users a low-risk mitigation path for high CPU usage by allowing deployments to run the UAS executor with 2 worker threads, while preserving the current hardware-concurrency-based default when the variable is unset.

Closes #2031.

## Validation

- `git diff --check`

I could not run `colcon test` locally in this Windows environment because ROS 2/colcon and a C++ compiler are not installed here.

### DIFF
--- a/mavros/src/lib/uas_executor.cpp
+++ b/mavros/src/lib/uas_executor.cpp
@@ -12,12 +12,40 @@
  */
 
 #include <algorithm>
+#include <charconv>
+#include <cstdlib>
+#include <string_view>
+#include <system_error>
 
 #include "mavros/uas_executor.hpp"
 
 using namespace mavros;                 // NOLINT
 using namespace mavros::uas;            // NOLINT
 using namespace std::chrono_literals;   // NOLINT
+
+namespace
+{
+constexpr size_t kMinExecutorThreads = 2;
+constexpr size_t kMaxExecutorThreads = 16;
+constexpr std::string_view kExecutorThreadsEnv = "MAVROS_UAS_EXECUTOR_THREADS";
+
+size_t configured_number_of_threads()
+{
+  auto value = std::getenv(kExecutorThreadsEnv.data());
+  if (value == nullptr || value[0] == '\0') {
+    return 0;
+  }
+
+  size_t threads = 0;
+  const auto input = std::string_view(value);
+  const auto result = std::from_chars(input.data(), input.data() + input.size(), threads);
+  if (result.ec != std::errc{} || result.ptr != input.data() + input.size() || threads == 0) {
+    return 0;
+  }
+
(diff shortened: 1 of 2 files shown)


mavros/src/lib/uas_executor.cpp
```diff
@@ -12,13 +12,41 @@
...
 
 using namespace mavros;                 // NOLINT
 using namespace mavros::uas;            // NOLINT
 using namespace std::chrono_literals;   // NOLINT
 
+namespace
+{
+constexpr size_t kMinExecutorThreads = 2;
+constexpr size_t kMaxExecutorThreads = 16;
+constexpr std::string_view kExecutorThreadsEnv = "MAVROS_UAS_EXECUTOR_THREADS";
```
vooon, line 30: I prefer old C-convention for the constants.
Also i don't think it really needs string view.

mavros/src/lib/uas_executor.cpp
```diff
@@ -12,13 +12,41 @@
...
+  if (value == nullptr || value[0] == '\0') {
+    return 0;
+  }
+
+  size_t threads = 0;
+  const auto input = std::string_view(value);
+  const auto result = std::from_chars(input.data(), input.data() + input.size(), threads);
+  if (result.ec != std::errc{} || result.ptr != input.data() + input.size() || threads == 0) {
+    return 0;
+  }
```
vooon, line 44: `std::stoi()` seems much more compact...

mavros/src/lib/uas_executor.cpp
```diff
@@ -12,13 +12,41 @@
...
+
+  return std::clamp(threads, kMinExecutorThreads, kMaxExecutorThreads);
```
vooon, line 46: If user sets amount of threads we should only limit minimum to be 2+.

vooon (changes requested):

apples-kksk:
Hi @jacog2, I prepared a small local patch for the three review points here: switch the env parsing to `std::stoi()`, use C-style constant names without `string_view`, and only clamp user-provided thread counts to the minimum of 2. I also adjusted the README wording to remove the old 2-16 limit.

Would you like me to send that patch somehow, or would you prefer to update the PR yourself?

vooon:
Replaced by #2197

Issue #2024, opened 2025-03-12
BUG: plugin_allowlist accepts garbage input without error


Ryanf55:
# Expected Behavior

If you request plugins to load that don't exist, mavros should either print an error

# Actual behavior

No error is emitted, a garbage plugin name is silently discarded. This results in a typo in a plugin name resulting in not loading what you expect.

```yaml
/**:
  ros__parameters:
    plugin_allowlist:
       - 'sys_status'
       - 'sys_time'
       - 'garbage'
```

```
$ ros2 param dump /mavros/mavros
/mavros/mavros:
  ros__parameters:
    base_link_frame_id: base_link
    component_id: 191
    diagnostic_updater:
      period: 1.0
      use_fqn: false
    fcu_protocol: v2.0
    map_frame_id: map
    odom_frame_id: odom
    plugin_allowlist:
    - sys_status
    - sys_time
    - garbage
```

PR #2192, opened 2026-05-12
mavros: validate plugin list patterns

apples-kksk:
## Requested
Issue #2024 reports that `plugin_allowlist` accepts garbage input without reporting an error.

## Implemented
- Validate `plugin_allowlist` and `plugin_denylist` patterns against declared plugin classes before loading plugins.
- Preserve existing glob matching behavior for valid exact and wildcard patterns.
- Add unit coverage for valid patterns and invalid allowlist/denylist values.

## Not covered
No changes to plugin loading order or plugin implementations.

## Validated
- `git diff --check HEAD~1..HEAD`
- Attempted `cmake -S mavros -B /tmp/mavros-2024-cmake-check`, but this environment is missing `ament_cmake`.

`colcon` and `ament_uncrustify` are not available in the current environment.

### DIFF
--- a/mavros/src/lib/mavros_uas.cpp
+++ b/mavros/src/lib/mavros_uas.cpp
@@ -158,28 +158,28 @@ UAS::UAS(
 
       const auto declared_plugins = plugin_factory_loader.getDeclaredClasses();
       const auto warn_unmatched_plugin_patterns =
-        [this, &declared_plugins](
+      [this, &declared_plugins](
         const StrV & patterns,
         const std::string & parameter_name)
-        {
-          for (auto & pattern : patterns) {
-            bool matched = false;
+      {
+        for (auto & pattern : patterns) {
+          bool matched = false;
 
-            for (auto & plugin : declared_plugins) {
-              if (pattern_match(pattern, plugin)) {
-                matched = true;
-                break;
-              }
-            }
-
-            if (!matched) {
-              RCLCPP_WARN(
-                get_logger(),
-                "%s pattern '%s' does not match any declared plugin",
-                parameter_name.c_str(), pattern.c_str());
+          for (auto & plugin : declared_plugins) {
+            if (pattern_match(pattern, plugin)) {
+              matched = true;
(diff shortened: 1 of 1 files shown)


mavros/src/lib/uas_executor.cpp
```diff
@@ -12,13 +12,41 @@
...
+  size_t threads = 0;
+  const auto input = std::string_view(value);
+  const auto result = std::from_chars(input.data(), input.data() + input.size(), threads);
+  if (result.ec != std::errc{} || result.ptr != input.data() + input.size() || threads == 0) {
+    return 0;
+  }
+
+  return std::clamp(threads, kMinExecutorThreads, kMaxExecutorThreads);
+}
+}  // namespace
```
vooon: That was addressed in different PR - #2189 please don't mix the things.

vooon (changes requested):
I think it's a bad idea crashing if something in a list doesn't match.
I think better just to emit warning.

apples-kksk:
Updated, thanks for the review.\n\n- Rebased the branch so the #2189 executor-thread changes are no longer included in this PR.\n- Changed unmatched plugin allow/deny patterns to emit a warning instead of throwing/crashing.\n\nI also checked the final PR diff: it now only touches `mavros/src/lib/mavros_uas.cpp`.

vooon (approved):
Now looks good, but why not to use lambda?

apples-kksk:
No strong reason. I used a small named helper to keep the startup block readable and reuse the same logic for both `plugin_allowlist` and `plugin_denylist`.

A local lambda would work too; I can switch to that if you prefer.

vooon:
Yes, it likely be cleaner.

vooon:
Please do uncrustify: https://github.com/mavlink/mavros/blob/ros2/CONTRIBUTING.md

apples-kksk:
Applied the uncrustify formatting in 9966b4ba. Thanks.

PR #2221, opened 2026-06-21
mavros_extras: Fix dangling pointers and uninitialized memory with Eigen

klaxalk:
While testing builds with GCC 15.2 (while adopting Nix which has new versions of underlying libraries and compilers), I noticed several `-Wdangling-pointer` and `-Wmaybe-uninitialized` warnings that were causing undefined behavior and silent memory corruption.

The issue stems from using `auto` to store the result of Eigen math formulas (e.g., `auto vec = Vector3d(...) * M_PI`). Under the hood, this evaluates to an expression template rather than an actual vector. Since the original temporary vector gets destroyed immediately, the expression template ends up holding a dangling pointer.

To fix this, I swapped the `auto` declarations for explicit `Eigen::Vector3d` types in `hil.cpp` and `mount_control.cpp` to force immediate evaluation. I also zero-initialized a vector in `landing_target.cpp` to clear out an uninitialized memory warning.

Tested locally and everything compiles cleanly now:
**Summary: 415 tests, 0 errors, 0 failures, 100 skipped**

### DIFF
--- a/mavros_extras/src/plugins/hil.cpp
+++ b/mavros_extras/src/plugins/hil.cpp
@@ -172,10 +172,10 @@ private:
     state_quat.ind_airspeed = req->ind_airspeed * 1E2;
     state_quat.true_airspeed = req->true_airspeed * 1E2;
     // WRT world frame
-    auto ang_vel = ftf::transform_frame_enu_ned(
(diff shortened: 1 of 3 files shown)


vooon (approved):
LGTM, Thanks!

PR #2226, opened 2026-06-27
global_position: fix ENU velocity conversion

### LINKED ISSUE #2054: Topic `global_position/local` not consistent linear z velocity in ENU frame
When filling mavros topic `global_position/local` and converting the linear velocity frame from `NED` to `ENU` the `z` coordinate is not consistent with the transformation, in fact, it remains positive toward down.

https://github.com/mavlink/mavros/blob/ros2/mavros/src/plugins/global_position.cpp#L343
```
    // Linear velocity
    tf2::toMsg(
      Eigen::Vector3d(gpos.vy, gpos.vx, gpos.vz) / 1E2,
      odom.twist.twist.linear);
```
to be consistent it should be:
```
    // Linear velocity
    tf2::toMsg(
      Eigen::Vector3d(gpos.vy, gpos.vx, -gpos.vz) / 1E2,
      odom.twist.twist.linear);
```

You can see it also from this plot:

Image

beniaminopozzan:
`GLOBAL_POSITION_INT` MAVlink [message](https://mavlink.io/en/messages/common.html#GLOBAL_POSITION_INT) expresses velocity in NED frame therefore a NED->ENU conversion is needed to comply with ROS 2 standard.

FIX: #2054: instead of converting into `ENU`, the code was converting into `END` frame.

This PR applies the same logic that is used in the `local_position` plugin and leverages `ftf::transform_frame_ned_enu`.

### DIFF
--- a/mavros/src/plugins/global_position.cpp
+++ b/mavros/src/plugins/global_position.cpp
@@ -342,7 +342,8 @@ private:
 
     // Linear velocity
     tf2::toMsg(
-      Eigen::Vector3d(gpos.vy, gpos.vx, gpos.vz) / 1E2,
+      ftf::transform_frame_ned_enu(
+        Eigen::Vector3d(gpos.vx, gpos.vy, gpos.vz)) / 1E2,
       odom.twist.twist.linear);
 
     // Velocity covariance unknown


vooon (approved):
Well, that's API change, but ok. LGTM.

Issue #1472, opened 2020-08-26
[Bug][Test video] Yaw data is not decreasing when I turn left from 0, but suddenly changes to 3.14


cnpcshangbo:
----

### Issue details

Hello, I am following [this setup](https://discuss.ardupilot.org/t/integration-of-ardupilot-and-vio-tracking-camera-part-2-complete-installation-and-indoor-non-gps-flights/43405). However, I noticed that the yaw data is not decreasing when I turn left from 0, but suddenly changes to 3.14. Here's my [video](https://youtu.be/fGTIb4XI0IU). Both the Terminal and Rivz are showing rostopic of /mavros/vision_pose/pose. The Mission Planner is showing the output of Mavros. I have tried Mavros 1.0, 1.1, 1.2, and the latest on GitHub. They have the same result. Did I miss something? Thanks.

### MAVROS version and platform

Mavros: 1.0, 1.1, 1.2
ROS: M
Ubuntu: 18.04

### Autopilot type and version

[ x ] ArduPilot
[  ] PX4

Version: 4.1.0-dev

### Node logs

```

root@nvidia-desktop:/home/nvidia# roslaunch mavros apm.launch
... logging to /root/.ros/log/67d5b5b2-e7bd-11ea-8e30-00044bf20a08/roslaunch-nvidia-desktop-8520.log
Checking log directory for disk usage. This may take a while.
Press Ctrl-C to interrupt
Done checking log file disk usage. Usage is <1GB.

started roslaunch server http://nvidia-desktop:40421/

SUMMARY
========

CLEAR PARAMETERS
 * /mavros/

PARAMETERS
 * /mavros/cmd/use_comp_id_system_control: False
 * /mavros/conn/heartbeat_mav_type: ONBOARD_CONTROLLER
 * /mavros/conn/heartbeat_rate: 1.0
 * /mavros/conn/system_time_rate: 1.0
 * /mavros/conn/timeout: 10.0
 * /mavros/conn/timesync_rate: 10.0
 * /mavros/distance_sensor/rangefinder_pub/field_of_view: 0.0
 * /mavros/distance_sensor/rangefinder_pub/frame_id: lidar
 * /mavros/distance_sensor/rangefinder_pub/id: 0
 * /mavros/distance_sensor/rangefinder_pub/send_tf: False
 * /mavros/distance_sensor/rangefinder_pub/sensor_position/x: 0.0
 * /mavros/distance_sensor/rangefinder_pub/sensor_position/y: 0.0
[...]

Issue #358, opened 2015-08-04
setpoint_position plugin sets yaw incorrectly


madsciencetist:
When I use the setpoint_position plugin to command a pose, the orientation does not behave as expected. To test, I am using the "2D Nav Goal" button in rviz with /move_base_simple/goal mapped to /mavros/setpoint_position/local. When I command a North heading, a quaternion of (0,0,0,1) is sent, which mavros converts to a yaw of zero, as expected. As I increase the commanded heading clockwise to South, the quaternion approaches (-1,0,0,0), which converts to a yaw of Pi, as expected. Any further though, and the angle goes to *zero*. Observe:

[/mavros]: q = (-0.999456, 0.000000, -0.000000, -0.032994) -> yaw = 3.075592
[/mavros]: q = (-0.999550, 0.000000, -0.000000, 0.029986) -> yaw = 0.059980

Thus, yaw setpoints can only be commanded in the [0,pi] range, as the other half of the circle will also map to [0,pi]. The problem must lie in UAS::quaternion_to_rpy() in uas_frame_conversions.cpp, but I'm not sure what the right answer is.

vooon:
I wrote tests and may confirm that q->rpy results is strange.
@TSC21 may you check math please?

```
vovan@robert:~/ros/src/mavros/mavros$ catkin run_tests --this -j1
...
[==========] Running 10 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 10 tests from UAS
[ RUN      ] UAS.transform_frame__vector3d_123
[       OK ] UAS.transform_frame__vector3d_123 (0 ms)
[ RUN      ] UAS.transform_frame__quaterniond_123
[       OK ] UAS.transform_frame__quaterniond_123 (0 ms)
[ RUN      ] UAS.transform_frame__covariance3x3
[       OK ] UAS.transform_frame__covariance3x3 (0 ms)
[ RUN      ] UAS.quaternion_from_rpy__check_compatibility
[       OK ] UAS.quaternion_from_rpy__check_compatibility (0 ms)
[ RUN      ] UAS.quaternion_from_rpy__paranoic_check
[       OK ] UAS.quaternion_from_rpy__paranoic_check (0 ms)
[ RUN      ] UAS.quaternion_to_rpy__123
[       OK ] UAS.quaternion_to_rpy__123 (0 ms)
[ RUN      ] UAS.quaternion_to_rpy__123_negative
[...]

PR #2232, opened 2026-07-06
frame_tf: fix quaternion_to_rpy yaw folding past 180° (eulerAngles clamping)

odraudE31:
`Eigen::eulerAngles(2,1,0)` clamps the primary extraction axis to `[0, π]`.
  In ZYX the primary axis is Z (yaw), so any heading past 180° is folded back
  rather than continuing to negative values. A full 360° rotation produces
  yaw 0 → π → 0 → π instead of the correct 0 → π → −π → 0.

  This flows through `vision_pose_estimate` into `VISION_POSITION_ESTIMATE`,
  which carries only Euler floats. The EKF receives the folded yaw as truth and
  sees the heading suddenly reverse direction mid-flight.

  Reported in #358 (2015), #444 (2015), and #1472 (2020). No fix was ever
  merged; the ros2 port in 2021 carried the bug forward.

  ## Changes

  Replace with the standard ZYX rotation-matrix decomposition:

  ```cpp
  double sin_pitch = std::clamp(-m(2, 0), -1.0, 1.0);
  double cos_pitch = std::sqrt(m(2, 1) * m(2, 1) + m(2, 2) * m(2, 2));

  if (cos_pitch > 1e-9) {
    return Eigen::Vector3d(
      std::atan2(m(2, 1), m(2, 2)),    // roll  ∈ (-π, π]
      std::atan2(sin_pitch, cos_pitch), // pitch ∈ (-π/2, π/2)
      std::atan2(m(1, 0), m(0, 0)));   // yaw   ∈ (-π, π]
  }
  // Gimbal lock (pitch = ±π/2): only one degree of freedom is observable.
  // Set roll = 0 and place yaw−roll (or yaw+roll) entirely in yaw.
  if (sin_pitch > 0) {
    return Eigen::Vector3d(0.0, M_PI / 2.0,  std::atan2( m(1, 2), m(1, 1)));
  }
  return Eigen::Vector3d(0.0, -M_PI / 2.0, std::atan2(-m(1, 2), m(1, 1)));
``` 

  atan2 returns (−π, π] so yaw is now correct for every heading.
  The gimbal-lock branch handles pitch = ±π/2 consistently with the
  existing quaternion_to_rpy__pm_pi round-trip test.

Also updates quaternion_to_rpy__123 to use pitch=0.5 rad (within the
  ZYX unique domain) and adds quaternion_to_rpy__yaw_full_circle which
  sweeps yaw through all integer degrees in [−179°, 180°] and asserts exact
  round-trip recovery — this test fails on the old code for any yaw in (−π, 0).

  Testing

  - All 8 libmavros-quaternion-utils-test cases pass.
  - Tested on hardware: ZED 2i → ArduCopter 4.5 EKF3 via MAVROS, ROS Humble / Ubuntu 22.04.

### DIFF
--- a/mavros/src/lib/ftf_quaternion_utils.cpp
+++ b/mavros/src/lib/ftf_quaternion_utils.cpp
@@ -39,15 +39,31 @@ Eigen::Quaterniond quaternion_from_rpy(const Eigen::Vector3d & rpy)
 
 Eigen::Vector3d quaternion_to_rpy(const Eigen::Quaterniond & q)
 {
-  // ZYX — atan2-based decomposition avoids Eigen::eulerAngles(2,1,0) clamping
-  // yaw to [0, pi]. All three angles now cover their full atan2/asin range:
-  //   roll  [-pi, pi], pitch [-pi/2, pi/2], yaw [-pi, pi]
+  // ZYX decomposition via rotation matrix. Avoids Eigen::eulerAngles(2,1,0)
+  // clamping yaw to [0, pi].
+  //
+  // Normal case (|pitch| < pi/2): roll, pitch, yaw all recovered via atan2/asin.
+  // Gimbal lock (pitch = ±pi/2): roll and yaw are not independently observable;
+  // we set roll = 0 and place the combined degree of freedom in yaw.
   auto m = q.toRotationMatrix();
-  return Eigen::Vector3d(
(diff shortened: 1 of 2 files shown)


vooon (approved):
LGTM, Thanks!

PR #2261, opened 2026-08-10
Fix fake GPS velocity frame and COG units

### LINKED ISSUE #2091: Wrong velocity and cog in the fake_gps plugin
https://github.com/mavlink/mavros/blob/c655e6343ec81687d51e7185bcba7a651e361fcd/mavros_extras/src/plugins/fake_gps.cpp#L286-L353

The calculation method for hil_gps vn,ve,vd in the fake_gps module is incorrect. Here, vel is the velocity obtained by the difference between old_ecef and current_ecef, meaning that vel is located in the ECEF coordinate system. However, the actual vn,ve,vd should be values ​​in the NED coordinate system. Therefore, the calculation of vel here is wrong.

Another issue is that the unit of hil_gps.cog should be degree * 1e2, but in this module, cog is calculated in radians.

These two problems lead to the wrong HIL GPS mavlink streams and fail to provide the drone a accurate fake gps location. I have tested on real drones and find these two issues.

The same problem lies in ROS1 branch too.

Daniiiil1:
## Summary

Fixes the fake GPS velocity and course fields reported in #2091.

## Root cause

`send_fake_gps()` differentiated ECEF positions, reversed the displacement sign, and then wrote those ECEF components directly into MAVLink fields documented as North/East/Down. The HIL_GPS course calculation also produced radians but stored the result as centidegrees. The first sample additionally used uninitialized previous-position state.

## Changes

- derive the ECEF velocity with the correct displacement sign
- rotate ECEF velocity into the local ENU frame and then into NED
- emit course over ground clockwise from north in centidegrees
- treat the first sample and non-increasing timestamps as zero velocity
- use the corrected NED velocity for both HIL_GPS and GPS_INPUT
- add focused tests for ECEF-to-NED axes, timestamp guards, and cardinal/intercardinal courses

## Validation

- `git diff --check`
- added deterministic gtests covering north/east/up transforms and 0/45/90/180/270 degree courses

The ROS 2 toolchain is not available on the local macOS host, so the package test target is left for the repository CI matrix.

Closes #2091


vooon:
@Daniiiil1 please fix cmake includes for this test.

mavros_extras/src/plugins/fake_gps_utils.hpp
```diff
@@ -0,0 +1,74 @@
...
+    return 0;
+  }
+
+  double course_deg = std::atan2(east, north) * 180.0 / M_PI;
+  if (course_deg < 0.0) {
+    course_deg += 360.0;
+  }
+
+  return static_cast<uint16_t>(std::lround(course_deg * 100.0)) % 36000;
+}
```
vooon, line 70: Maybe worth to move to generic frame_tf.hpp?

shupx:
@Daniiiil1 Great! Thank you for resolving this issue that has been troubling me for a long time. Prior to this, I could only work around it by writing a separate module to bypass MAVROS's fake GPS.

vooon (approved):
I've moved utils to frame_tf.

PR #2291, opened 2026-09-09
Add mode parser for Flix autopilot

okalachev:
Added basic [Flix quadcopter](https://github.com/okalachev/flix) support.

Warning: did only a quick testing - I was able go get the state and the 'connected' flag, but wasn't able to get the telemetry itself, for some reason. Attitude, battery and other topics didn't work, I need to investigate that. Therefore WIP.


README.md
```diff
@@ -4,7 +4,7 @@
 
 **A MAVLink-to-ROS 2 bridge for drones.**
 
-MAVROS turns a MAVLink autopilot (PX4, ArduPilot, …) into a set of familiar
+MAVROS turns a MAVLink autopilot (PX4, ArduPilot, Flix, …) into a set of familiar
```
okalachev, line 7: @vooon, should I add it here, or PX4 and Ardupilot are too much bigger to list Flix in here?

vooon: Why not? Let it be!

vooon (approved):
Looks good. Lyrical build is green - it uses latest released mavlink (2026.9.9).

okalachev:
I tested the connection, and it works! Though there is not a lot of working topics.

But I had to make another change.

The custom mode parser was entirely disabled if `CUSTOM_MODE_ENABLED` flag is not set. The problem is the flag's description says:

> When using this flag to enable a custom mode **all other flags should be ignored**.

But I don't want all the other flags to be ignored! On the contrary, I want them to be considered. QGroundControl doesn't support my custom modes yet, but when `CUSTOM_MODE_ENABLED` is not set, it _gracefully degrades_ and actually shows all the modes (stab, acro, auto) in a sane way, using base mode flags.

I don't want to remove this behavior, so I cannot set the `CUSTOM_MODE_ENABLED` flag.

So there is two options to solve this:

1. Remove `CUSTOM_MODE_ENABLED` check from the begin of the `str_base_mode` method (this is done). It's totally safe since both PX4 and ArduPilot actually never unset this bit: [1](https://github.com/PX4/PX4-Autopilot/blob/e880f1cd79e2453ef3dfaf68ee1c067b98ea9a53/src/modules/mavlink/streams/HEARTBEAT.hpp#L80), [2](https://github.com/ArduPilot/ardupilot/blob/de411b3818bd9ccabab90151c22f4bb3bb367353/ArduCopter/GCS_MAVLink_Copter.cpp#L57), [3](https://github.com/ArduPilot/ardupilot/blob/fbf750c2cad78df7e1fce9f273d906874502b9f0/ArduPlane/GCS_MAVLink_Plane.cpp#L19), [4](https://github.com/ArduPilot/ardupilot/blob/de411b3818bd9ccabab90151c22f4bb3bb367353/Rover/GCS_MAVLink_Rover.cpp#L50), [5](https://github.com/ArduPilot/ardupilot/blob/de411b3818bd9ccabab90151c22f4bb3bb367353/ArduSub/GCS_MAVLink_Sub.cpp#L48).
2. Keep the `CUSTOM_MODE_ENABLED` check but rearrange the main `if` statement so it wouldn't affect the Flix autopilot only. This seems a more complex and less straightforward solution.

@vooon, what do you think?

mavros/src/lib/uas_stringify.cpp
```diff
@@ -205,10 +205,6 @@ static inline bool is_apm_copter(uas::MAV_TYPE type)
 
 std::string UAS::str_mode_v10(uint8_t base_mode, uint32_t custom_mode)
 {
-  if (!(base_mode & enum_value(MAV_MODE_FLAG::CUSTOM_MODE_ENABLED))) {
```
vooon, line 208: @okalachev maybe better if you just set that flag on the firmware side?

